PR #195, opened 2021-10-15
Update rule metadata for version 1.3

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
Metadata changes LGTM, but there appears to be a problem with building.

PR #199, opened 2021-11-08
[CF] SONARIAC-199 Rule S4423: Weak SSL/TLS protocols should not be used (for API Gateway domain names)

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #200, opened 2021-11-08
[CF] Rule S5332: Using clear-text protocols is security-sensitive (For MSK internal communications)

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #202, opened 2021-11-08
[TF] SONARIAC-20 Rule S4423: Weak SSL/TLS protocols should not be used (for Elasticsearch domain endpoints)

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #205, opened 2021-11-08
SONARIAC-260 Rule S6304: No issue should be raised for KMS key policies

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #207, opened 2021-11-09
[TF] SONARIAC-221 Rule S5332: Using clear-text protocols is security-sensitive (for AWS Load Balancer Listeners)

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM, but as discussed, we have FNs when we have multiple `default_action` blocks:
```
resource "aws_lb_listener" "sensitive_redirect" {
  protocol = "HTTP" # Noncompliant
  default_action {
    type = "foo"
    redirect {
      protocol = "HTTPS"
    }
  }

  default_action {
    type = "redirect"
    redirect {
      protocol = "HTTP"
    }
  }
}
```
Up to you if you want to fix this in this PR (I can re-review if you do this) or create another ticket for this (we'll need to adapt `PropertyUtils` for this). If you want to create another ticket, let's add a FN example in the tests for now.

PR #209, opened 2021-11-09
SONARIAC-264 PropertyUtils should return all properties with expected name

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #210, opened 2021-11-09
[TF] SONARIAC-214 Rule S5332: Using clear-text protocols is security-sensitive (for Elasticache Replication Group)

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #217, opened 2021-11-10
[CF] SONARIAC-223 Rule S6258: Disabling logging is security sensitive (X-Ray tracing on AWS API Gateway)

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
LGTM

PR #223, opened 2021-11-10
[CF] SONARIAC-229 Rule S6258: Disabling logging is security sensitive (AWS MSK)

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #229, opened 2021-11-11
[CF] SONARIAC-239 Rule S6258: Disabling logging is security-sensitive (for MQ)

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #231, opened 2021-11-11
[TF] SONARIAC-238 Rule S6258: Disabling logging is security-sensitive (for MQ)

nils-werner-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM

PR #240, opened 2021-11-11
[TF] SONARIAC-253 Rule S6258: Disabling logging is security sensitive (Elastic Load Balancing v1)

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/AbstractResourceCheck.java
```diff
@@ -72,4 +73,17 @@ public static void reportOnFalse(CheckContext ctx, Tree tree, String message) {
   public static void reportResource(CheckContext ctx, BlockTree resource, String message) {
     ctx.reportIssue(resource.labels().get(0), message);
   }
+
+  public static void reportOnDisabled(CheckContext ctx, BlockTree block, boolean enabledByDefault, String message) {
```
karim-ouerghemmi-sonarsource, line 77: Are there other already existing rules that can make use of this? As of now, this is only used in `DisabledLoggingCheck`, so we can just have it in there if there are no other existing rules.

karim-ouerghemmi-sonarsource (approved):
LGTM. One comment.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/AbstractResourceCheck.java
nils-werner-sonarsource: I will update the other checks in a later PR

PR #244, opened 2021-11-12
[TF] SONARIAC-267 S6321 Terraform should consider all ingress blocks

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #248, opened 2021-11-15
Update release workflow

nils-werner-sonarsource:
Following https://jira.sonarsource.com/browse/BUILD-1126 and https://github.com/SonarSource/sonar-dummy-oss/commit/1150c457818ce52a582b4af629cb216380f5c8ad

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
           RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
           RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
           SLACK_API_TOKEN: ${{secrets.SLACK_API_TOKEN }}
-        uses: SonarSource/gh-action_release/main@v3backup
+        uses: SonarSource/gh-action_release/main@v3
         with:
           publish_to_binaries: true
           attach_artifacts_to_github_release: false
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -22,10 +22,12 @@ jobs:
           RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
           RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
           SLACK_API_TOKEN: ${{secrets.SLACK_API_TOKEN }}
-        uses: SonarSource/gh-action_release/main@v3backup
+        uses: SonarSource/gh-action_release/main@master
         with:
           publish_to_binaries: true
           attach_artifacts_to_github_release: false
+          distribute: false
+          distribute_target: false
```
karim-ouerghemmi-sonarsource, line 30: I'm not sure what this means, to be honest.

.github/workflows/release.yml
```diff
@@ -22,10 +22,12 @@ jobs:
           RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
           RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
           SLACK_API_TOKEN: ${{secrets.SLACK_API_TOKEN }}
-        uses: SonarSource/gh-action_release/main@v3backup
+        uses: SonarSource/gh-action_release/main@master
```
karim-ouerghemmi-sonarsource, line 25: Should we always use the latest master? It seems to me that all other projects we have do use `main@v3`.

karim-ouerghemmi-sonarsource (commented):
See comments. Honestly, I'm not too savvy with this, and I think we should let somebody from RE team review.

PR #257, opened 2021-12-14
SONARIAC-283 Declare CloudFormation and Terraform sensor as a sensor processing files independently

nils-werner-sonarsource:
This pull request is inspired by https://github.com/SonarSource/sonar-swift/pull/429 which was given as an example during the weekly.

### DIFF
--- a/iac-common/src/test/java/org/sonar/iac/common/extension/IacSensorTest.java
+++ b/iac-common/src/test/java/org/sonar/iac/common/extension/IacSensorTest.java
@@ -24,6 +24,9 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import org.junit.jupiter.api.Test;
+import org.sonar.api.SonarEdition;
+import org.sonar.api.SonarQubeSide;
+import org.sonar.api.SonarRuntime;
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.api.batch.fs.TextPointer;
 import org.sonar.api.batch.fs.TextRange;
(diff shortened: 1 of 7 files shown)


iac-common/src/test/java/org/sonar/iac/common/extension/IacSensorTest.java
```diff
@@ -71,6 +77,29 @@ void test_descriptor() {
...
+    final boolean[] called = {false};
+    DefaultSensorDescriptor descriptor = new DefaultSensorDescriptor() {
+      public SensorDescriptor processesFilesIndependently() {
+        called[0] = true;
+        return this;
+      }
+    };
+    sensor(SonarRuntimeImpl.forSonarQube(Version.create(9, 3), SonarQubeSide.SCANNER, SonarEdition.DEVELOPER), null)
+      .describe(descriptor);
+    assertThat(called[0]).isTrue();
```
pynicolas, line 92: This test method is weird to me. Could we use something like mockito to make its intent clearer?

iac-common/src/test/java/org/sonar/iac/common/extension/IacSensorTest.java
```diff
@@ -71,6 +77,29 @@ void test_descriptor() {
...
+  }
+
+  @Test
+  void test_descriptor_sonarqube_9_3_reflection_failure() {
+    DefaultSensorDescriptor descriptor = new DefaultSensorDescriptor();
+    sensor(SonarRuntimeImpl.forSonarQube(Version.create(9, 3), SonarQubeSide.SCANNER, SonarEdition.DEVELOPER), null)
+      .describe(descriptor);
+    assertThat(logTester.logs()).contains("Could not call SensorDescriptor.processesFilesIndependently() method");
```
pynicolas, line 100: I don't really get the trick. Why don't we depend on latest sonar-plugin.jar?

pynicolas: I don't quite see what we prove with those tests: just that, if there's a method with a given name, we will call it?
What about adding integration tests to ensure that we have a different behavior between SQ 9.2 and SQ dev?

agigleux: > Why don't we depend on latest sonar-plugin.jar?
I think it's because the sonar-plugin-9.3.jar is not yet released and we don't want to wait it to be released to update all our analyzers and finally get this feature in SQ 9.4

nils-werner-sonarsource: Yeah. We have to do the reflection trick in the sensor as well as create our own implementation of DefaultSensorDescriptor in the tests because it's not available yet in the API.

PR #259, opened 2021-12-15
SONARIAC-275 Rule S6385: Azure custom roles should not grant subscription Owner capabilities

nils-werner-sonarsource:


iac-extensions/terraform/src/test/resources/checks/Azure/SubscriptionOwnerCapabilitiesCheck/noncompliant.tf
```diff
@@ -0,0 +1,113 @@
...
+
+  assignable_scopes = [
+    data.azurerm_management_group.root_example.id
+  ]
+}
+
+# Alternative way to define a custom role with a Management Group assignable scope
+resource "azurerm_role_definition" "management-group-owner-role-alternative-scope" { # Noncompliant
+  name        = "management-group-owner-role"
+  scope       = data.azurerm_management_group.example_parent.id
```
stanislav-jordanov-sonarsource, line 103: Where's the data definition of this data reference, or have I understood data incorrectly?

stanislav-jordanov-sonarsource (approved):

iac-extensions/terraform/src/test/resources/checks/Azure/SubscriptionOwnerCapabilitiesCheck/noncompliant.tf
nils-werner-sonarsource: You've properly understood this correctly. However, for this check, we do not need to define the data block because we don't consider if the block truly exists. When you want to see real executable code, you can have a look at the "[expected issues](https://github.com/SonarSource/security-expected-issues/pull/626/files)" for each rule. The AppSec bubble tries to test the code in a real environment. Nevertheless, for our parser as well as our checks we do not need to write runnable code.

PR #263, opened 2021-12-17
SONARIAC-279 Rule S6387: Azure role assignments that grant access to all resources of a subscription are security-sensitive

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/helper/RoleScopeHelper.java
```diff
@@ -0,0 +1,91 @@
...
+  private RoleScopeHelper() {
+    // helper class
+  }
+
+  public static Predicate<String> matchPredicate(String pattern) {
+    return s -> Pattern.matches(pattern, s);
+  }
+
+  public static boolean isSensitiveScope(ExpressionTree scope, Predicate<String> referenceScopePredicate, Predicate<String> plainScopePredicate) {
+    if (scope instanceof AttributeAccessTree) {
```
andrea-guarino-sonarsource, line 51: _Minor_: what about using `scope.is(ATTRIBUTE_ACCESS)` instead of `instanceof` ?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/helper/RoleScopeHelper.java
```diff
@@ -0,0 +1,91 @@
...
+      .filter(TemplateInterpolationTree.class::isInstance)
+      .map(interpolation -> ((TemplateInterpolationTree) interpolation).expression())
+      .filter(AttributeAccessTree.class::isInstance)
+      .map(AttributeAccessTree.class::cast)
+      .anyMatch(interpolation -> containsSensitiveScope(interpolation, referenceScopePredicate));
+  }
+
+  public static String referenceToString(AttributeAccessTree reference) throws IllegalArgumentException {
+    StringBuilder sb = new StringBuilder();
+    if (reference.object() instanceof AttributeAccessTree) {
```
andrea-guarino-sonarsource, line 75: same here, we could probably replace `instanceof` with `is`

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/helper/RoleScopeHelper.java
```diff
@@ -0,0 +1,91 @@
...
+    } else if (reference.object() instanceof VariableExprTree) {
+      sb.append(((VariableExprTree) reference.object()).value());
+      sb.append('.');
+    }
+    sb.append(reference.attribute().value());
+    return sb.toString();
+  }
+
+  private static boolean isLimitedToResourceGroup(TemplateExpressionTree scope) {
+    return (scope).parts().stream()
```
andrea-guarino-sonarsource, line 87: unnecessary parens around `scope`?

iac-extensions/terraform/src/test/resources/checks/Azure/SubscriptionOwnerCapabilitiesCheck/compliant.tf
```diff
@@ -33,6 +33,6 @@ resource "azurerm_role_definition" "role-with-limited-assignable-scopes-alternat
   }
 
   assignable_scopes = [
-    "/subscriptions/${var.subscription_id}/resourceGroups/${var.environment_name}" # Limited to a resource group
+    "/subscriptions/${data.azurerm_subscription.primary.id}/resourceGroups/${var.environment_name}" # Limited to a resource group
```
andrea-guarino-sonarsource, line 36: I don't fully understand why did you change this existing test

andrea-guarino-sonarsource (approved):
LGTM, I left a few minor comments

iac-extensions/terraform/src/test/resources/checks/Azure/SubscriptionOwnerCapabilitiesCheck/compliant.tf
nils-werner-sonarsource: Sorry for the lack of explanation. The content of this check is to check if a relation is created to a resource and contains the keyword `resourceGroups` in it's path. The focus here is on the keyword `resourceGroups`. I had to change the test, because we don't resolve variables and therefore I entered the resolved relation of the resource to be checked. If the keyword would be missing, then we would have a false negative, or the keyword check would not work, because the relation is not provided as a basic requirement.

andrea-guarino-sonarsource: Ok I see, thanks for clarifying!

PR #265, opened 2021-12-30
SONARIAC-285 [TF] AbstractMultipleResourceCheck should be merged with AbstractResourceCheck

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/DisabledMfaBucketDeletionCheck.java
```diff
@@ -36,11 +36,11 @@
   private static final String MESSAGE_SECONDARY = "Related bucket";
 
   @Override
-  protected void checkResource(CheckContext ctx, BlockTree tree) {
-    if (!isS3Bucket(tree)) {
-      return;
-    }
+  protected void registerChecks() {
+    register(DisabledMfaBucketDeletionCheck::checkBucket, "aws_s3_bucket");
```
pynicolas, line 40: Don't we want to have `aws_s3_bucket` as a constant?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/DisabledS3EncryptionCheck.java
```diff
@@ -29,8 +29,12 @@
...
 
   @Override
-  protected void checkResource(CheckContext ctx, BlockTree resource) {
-    if (isS3Bucket(resource) && PropertyUtils.isMissing(resource, "server_side_encryption_configuration")) {
+  protected void registerChecks() {
+    register(DisabledS3EncryptionCheck::checkBucket, "aws_s3_bucket");
+  }
+
+  private static void checkBucket(CheckContext ctx, BlockTree resource) {
+    if (PropertyUtils.isMissing(resource, "server_side_encryption_configuration")) {
```
pynicolas, line 37: Just an idea: when the check is so short, it could be easier to use a lambda:
```
    register((ctx, resource) -> {
      if (PropertyUtils.isMissing(resource, "server_side_encryption_configuration")) {
        reportResource(ctx, resource, MESSAGE);
      }
    }, "aws_s3_bucket");
```
However, it could be annoying to have the name of the resource as the last argument.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/ShortBackupRetentionCheck.java
```diff
@@ -37,6 +37,11 @@
     defaultValue = "" + DEFAULT)
   int backupRetentionDuration = DEFAULT;
 
+  @Override
+  protected void registerChecks() {
+    // do not register any check for a specific resource type
```
pynicolas, line 42: If this is so common, why not having a default empty implementation for `registerChecks` in `AbstractResourceCheck`?

PR #274, opened 2022-01-11
Rule S6321: Administration services access should be restricted to specific IP addresses for `azurerm_network_security_rule`

nils-werner-sonarsource:


pynicolas (approved):
I'm not sure we need so extensive tests for "azurerm_network_security_rule" given what we have for "azurerm_network_security_group".

PR #278, opened 2022-01-12
SONARIAC-288 Rule S6329: Enabling public network access to Azure resources is security-sensitive

nils-werner-sonarsource:


guillaume-dequenne-sonarsource (approved):
The ticket referenced is `SONARIAC-287` in the PR name, `SONARIAC-281` in the commit and `SONARIAC-288` in the branch name.
From what I see on JIRA, this PR implements ticket `SONARIAC-288` so I believe this is the only ticket that should be referenced. Though I don't know whether you have different conventions for IaC knowing a rule can get divided in many tickets...
LGTM otherwise.

PR #279, opened 2022-01-13
Rule S6329: Enabling public network access to Azure resources is security-sensitive for public IP-related parameters

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/AzurePublicNetworkAccessCheckPart.java
```diff
@@ -52,21 +49,44 @@ protected void registerResourceChecks() {
...
   }
 
-  private static void checkPublicNetworkVariant(CheckContext ctx, BlockTree resource) {
-    PropertyUtils.get(resource, PUBLIC_NETWORK_ENABLED, AttributeTree.class)
-      .ifPresentOrElse(
-        enabled -> reportOnTrue(ctx, enabled, NETWORK_ACCESS_MESSAGE),
-        () -> reportResource(ctx, resource, String.format(OMITTED_MESSAGE, PUBLIC_NETWORK_ENABLED)));
+  private Consumer<Resource> reportPublicIpConfiguration(String propertyName) {
+    return resource -> resource.blocks(propertyName).forEach(
+      gateway -> gateway.attribute("public_ip_address_id")
```
pynicolas, line 88: I think I would rename `gateway` to something like `property` or `block` because this method is quite generic.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/utils/TerraformUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class TerraformUtils {
+
+  private TerraformUtils() {
+    // utils class
+  }
+
+  public static Trilean attributeAccessMatches(ExpressionTree expression, Predicate<String> predicate) {
+    if (expression.is(ATTRIBUTE_ACCESS)) {
+      return predicate.test(expression.toString()) ? Trilean.TRUE : Trilean.FALSE;
```
pynicolas, line 36: I don't really like this call to `toString` but I don't have much to suggest.
I would prefer a predicate on `ExpressionTree`.
`toString` is usually not considered as a strong contract, more intended to be used in messages.
If working on strings is the easiest way, I would be more comfortable if the methods which produce strings and the methods which test them were located in the same place.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/AzurePublicNetworkAccessCheckPart.java
```diff
@@ -52,21 +49,44 @@ protected void registerResourceChecks() {
...
+        .reportAbsence(OMITTED_MESSAGE));
+
+    resourceConsumer("azurerm_dev_test_virtual_network",
+      resource -> resource.block("subnet").ifPresent(
+        subnet -> subnet.attribute("use_public_ip_address")
+          .reportUnexpectedValue("Deny", NETWORK_ACCESS_MESSAGE)
+          .reportAbsence(OMITTED_MESSAGE)));
+
+    resourceConsumer("azurerm_batch_pool",
+      resource -> resource.block("network_configuration").ifPresent(
```
pynicolas, line 70: The [ticket](https://jira.sonarsource.com/browse/SONARIAC-289) mentions:

> If this parameter assignment is missing, this resource is non-compliant by default, so an issue should be raised.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/AzurePublicNetworkAccessCheckPart.java
```diff
@@ -52,21 +49,44 @@ protected void registerResourceChecks() {
...
+    resourceConsumer(List.of("azurerm_dev_test_linux_virtual_machine", "azurerm_dev_test_windows_virtual_machine"),
+      resource -> resource.attribute("disallow_public_ip_address")
+        .reportOnFalse(NETWORK_ACCESS_MESSAGE)
+        .reportAbsence(OMITTED_MESSAGE));
+
+    resourceConsumer("azurerm_dev_test_virtual_network",
+      resource -> resource.block("subnet").ifPresent(
+        subnet -> subnet.attribute("use_public_ip_address")
+          .reportUnexpectedValue("Deny", NETWORK_ACCESS_MESSAGE)
+          .reportAbsence(OMITTED_MESSAGE)));
```
pynicolas, line 67: I think that's not what the ticket specifies but it's not very explicit:
According to the [reference](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/dev_test_virtual_network), `subnet` is optional but `use_public_ip_address` is not.
So I suppose the ticket's intention is to raise an issue when `subnet` is missing, not when `use_public_ip_address` is missing.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/AzurePublicNetworkAccessCheckPart.java
```diff
@@ -52,21 +49,44 @@ protected void registerResourceChecks() {
...
+
+    resourceConsumer("azurerm_batch_pool",
+      resource -> resource.block("network_configuration").ifPresent(
+        configuration -> configuration.attribute("public_address_provisioning_type")
+          .reportUnexpectedValue("NoPublicIPAddresses", NETWORK_ACCESS_MESSAGE)));
+
+    resourceConsumer("azurerm_kubernetes_cluster_node_pool",
+      resource -> resource.block("default_node_pool").ifPresent(
```
pynicolas, line 75: The ticket only mentions `default_node_pool` in the example, not in the specification of the behavior and I don't understand the [reference](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool)...

pynicolas (changes requested):

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/ResourceProvider.java
```diff
@@ -0,0 +1,224 @@
...
+    public Attribute(CheckContext ctx, Block block, String name, @Nullable AttributeTree attributeTree) {
+      this.ctx = ctx;
+      this.block = block;
+      this.name = name;
+      this.attributeTree = attributeTree;
+    }
+
+    public Attribute reportOnTrue(String message, SecondaryLocation... secondaries) {
+      // designed to be extended but noop in standard case
+      return this;
```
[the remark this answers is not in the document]
nils-werner-sonarsource: We don't know if it is a Present or Absent attribute when we call the methods. It would make sense if we were still based on ifPresentOrElse. WDYT?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/AzurePublicNetworkAccessCheckPart.java
nils-werner-sonarsource: I think, there is a mistake in the detection logic. Have to clarify with AppSec.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/AzurePublicNetworkAccessCheckPart.java
nils-werner-sonarsource: I agree and check with AppSec

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/ResourceVisitor.java
```diff
@@ -0,0 +1,223 @@
...
+    protected final AttributeTree attributeTree;
+
+    public Attribute(CheckContext ctx, Block block, String name, @Nullable AttributeTree attributeTree) {
+      this.ctx = ctx;
+      this.block = block;
+      this.name = name;
+      this.attributeTree = attributeTree;
+    }
+
+    public Attribute reportOnTrue(String message, SecondaryLocation... secondaries) {
```
pynicolas, line 132: What about `reportIfTrue`/`reportIfFalse`/`reportIfAbsent` for consistency with the other methods which were just renamed? I feel it might also clarify what the method actually do.

pynicolas (approved):

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/ResourceVisitor.java
nils-werner-sonarsource: Ok for me

PR #289, opened 2022-01-18
SONARIAC-274 Rule S4423: Weak SSL/TLS protocols should not be used (for Azure)

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/aws/AwsWeakSSLProtocolCheckPart.java
```diff
@@ -0,0 +1,73 @@
+/*
```
stanislav-jordanov-sonarsource, line 1: If I understand correctly, you've effectively split the code of the old version of WeakSSLProtocolCheck
into this class (old-style checks) and then you've added class AzureWeakSSLProtocolCheckPart - new style checks?

stanislav-jordanov-sonarsource (approved):

stanislav-jordanov-sonarsource (approved):

PR #294, opened 2022-01-21
SONARIAC-297 Rule S5332: Using clear-text protocols in storage account is security-sensitive (for Azure)

nils-werner-sonarsource:


stanislav-jordanov-sonarsource (approved):
There is one codesmell:
_Make "checkStorageAccountDataSource" a "static" method._

Also (as I said in slack):
_I don't understand where your first commit comes from;
The second commit is exactly cutting out the source_api_id tracing and it is "LGTM"_

PR #304, opened 2022-02-10
use correct messages when reporting S6400

stanislav-jordanov-sonarsource:
The correct messages were specified in a comment to SONARIAC-361


nils-werner-sonarsource (approved):
LGTM!

PR #313, opened 2022-02-16
SONARIAC-357 Rule 6400: Granting highly privileged GCP resource rights is security-sensitive (ACLs)

stanislav-jordanov-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/HighPrivilegedRolesOnWorkloadResourcesCheck.java
```diff
@@ -19,49 +19,46 @@
...
 
-  private static final Predicate<ExpressionTree> IS_OWNER_OR_MANAGER = treePredicate(exactMatchStringPredicate("MANAGER|OWNER", Pattern.CASE_INSENSITIVE));
-  private static final Predicate<ExpressionTree> CONTAINS_PRIVILEGED_ROLE = treePredicate(containsMatchStringPredicate("ADMIN|MANAGER|OWNER|SUPERUSER", Pattern.CASE_INSENSITIVE));
-
   @Override
   protected void registerResourceConsumer() {
     register(List.of(IAM_BINDING_RESOURCE_NAMES),
       resource -> resource.attribute("role")
-        .reportIf(CONTAINS_PRIVILEGED_ROLE, MESSAGE_FOR_BINDING));
+                          .reportIf(matchesPattern(".*(?:ADMIN|MANAGER|OWNER|SUPERUSER).*"), MESSAGE_FOR_BINDING));
```
nils-werner-sonarsource, line 37: I'm not happy with the additional spaces here. The alignment of the "attribute()" and "reportIf()" methods means to me that they both apply to the resource symbol. However, we should find a common format/agreement. Otherwise, we will have a mixed style.

iac-extensions/terraform/src/test/resources/checks/GCP/HighPrivilegedRolesOnWorkloadResourcesCheck/roles_in_lists.tf
```diff
@@ -0,0 +1,63 @@
+resource "google_storage_bucket_acl" "rspecS6400_example1" {
+  bucket = google_storage_bucket.rspecS6400_example.name1
```
nils-werner-sonarsource, line 2: Could be removed because it is not relevant for the check. The same applies to the following repetitions.

nils-werner-sonarsource (approved):
LGTM! Please have a look at my comments. Open to discuss the method alignment.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/HighPrivilegedRolesOnWorkloadResourcesCheck.java
stanislav-jordanov-sonarsource: I agree with your argument (and I also remember that you have used that argument once, and I agreed to it back then)

PR #315, opened 2022-02-17
SONARIAC-360 Rule 6400: Granting highly privileged GCP resource rights is security-sensitive (IAM policy)

stanislav-jordanov-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/AbstractNewResourceCheck.java
```diff
@@ -104,4 +103,26 @@ private Pattern pattern(String regex, int flags) {
...
+  }
+
+  public static boolean isResourceOfType(BlockTree blockTree, String resourceType) {
+    return "resource".equals(blockTree.key().value()) && resourceType.equals(resourceType(blockTree));
+  }
+
+  public static boolean isData(BlockTree blockTree) {
+    return "data".equals(blockTree.key().value());
+  }
+
```
nils-werner-sonarsource, line 118: Both methods are only used in test cases and even not fully covered. Do we need them right now? I think we can add them later, when we need them.

nils-werner-sonarsource (approved):
LGTM! Please get 100% coverage and maybe address my comments.

PR #317, opened 2022-02-18
SONARIAC-320 Rule S6406: Excessive Granting Of GCP IAM Permissions Is Security-Sensitive

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
```diff
@@ -0,0 +1,90 @@
...
+import org.sonar.iac.terraform.symbols.ListSymbol;
+
+@Rule(key = "S6406")
+public class ExcessivePermissionsCheck extends AbstractNewResourceCheck {
+
+  private static final String MESSAGE = "This role grants more than %d sensitive permissions. Make sure they are all required.";
+  private static final String SECONDARY_MESSAGE = "Sensitive permission";
+  public static final int DEFAULT = 5;
+
+  private static final Set<String> SENSITIVE_FEATURE_SUFFIXES = Set.of("abort", "access", "add", "allocate", "analyze", "apply", "approve", "associate", "attach", "begin",
```
stanislav-jordanov-sonarsource, line 41: Couple of quotes from the ticket:
_We will refer to them as PREFIX, FEATURE, and SUFFIX, to avoid ambiguities._
_The SUFFIX starts with one of these words, OR is equal to one of these words ..._

Thus I suggest renaming this to **SENSITIVE_SUFFIX_PREFIXES** or **SENSITIVE_FEATURE_SUFFIX_PREFIXES**

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
```diff
@@ -0,0 +1,90 @@
...
+      "bind", "call", "cancel", "clear", "close", "compute", "connect", "create", "delete", "deploy", "destroy", "detach",
+      "disable", "drop", "enable", "evict", "exec", "import", "install", "invoke", "listVulnerabilities", "manage",
+      "migrate", "move", "mutate", "patch", "pause", "proxy", "publish", "purchase", "purge", "put", "reject", "remove",
+      "reopen", "replace", "rerun", "reset", "resize", "restart", "restore", "resume", "rollback", "rotate", "run",
+      "sample", "scan", "send", "set", "sign", "sourceCodeGet", "sourceCodeSet", "start", "stop", "suspend", "undelete",
+      "undeploy", "update", "upload", "use", "validate", "write");
+
+  private static final Set<String> SENSITIVE_FEATURE_SUFFIX_ELEMENTS = Set.of("login", "create", "delete", "set");
```
stanislav-jordanov-sonarsource, line 49: For analogous reasons, if you choose to go for SENSITIVE_SUFFIX_PREFIXES in the above case, here this should be renamed to simply **SENSITIVE_SUFFIX_ELEMENTS**
(If you go for SENSITIVE_FEATURE_SUFFIX_PREFIXES in the above case, then here it should remain as it is)

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
```diff
@@ -0,0 +1,90 @@
...
+      .isPresent();
+  }
+
+  private static String getPermissionSuffix(String permission) {
+    return (permission.indexOf('.') > -1 && permission.length() > 1) ? permission.substring(permission.lastIndexOf(".") + 1) : "";
+  }
+
+  private static boolean isSensitiveSuffix(String suffix) {
+    return SENSITIVE_FEATURE_SUFFIXES.stream().anyMatch(suffix::startsWith)
+      || SENSITIVE_FEATURE_SUFFIX_ELEMENTS.stream().anyMatch(suffix::contains);
```
stanislav-jordanov-sonarsource, line 88: As you use both these constant Set(s) to only stream() them
they should perhaps be replaced with constant List(s) ? (cheaper to construct, predictive order of streaming)

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
```diff
@@ -0,0 +1,90 @@
...
+    return expression -> TextUtils.getValue(expression)
+      .map(String::toLowerCase)
+      .map(ExcessivePermissionsCheck::getPermissionSuffix)
+      .filter(suffix -> !(suffix.contains("readonly")))
+      .filter(ExcessivePermissionsCheck::isSensitiveSuffix)
+      .isPresent();
+  }
+
+  private static String getPermissionSuffix(String permission) {
+    return (permission.indexOf('.') > -1 && permission.length() > 1) ? permission.substring(permission.lastIndexOf(".") + 1) : "";
```
stanislav-jordanov-sonarsource, line 83: Why not replace this expression with the simpler (and almost equivalent):
`permission.substring(permission.lastIndexOf(".") + 1)`
?
The only difference is that when invoked on '.'-less string, (e.g. "BMW") the original version will return "" while the suggested one will return "BMW" which is not that bad ...

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
```diff
@@ -0,0 +1,90 @@
...
+  public int max = DEFAULT;
+
+  @Override
+  protected void registerResourceConsumer() {
+    register(Set.of("google_organization_iam_custom_role", "google_project_iam_custom_role"),
+      resource -> {
+        ListSymbol permissions = resource.list("permissions");
+        List<ExpressionTree> sensitivePermissions = resource.list("permissions")
+          .getItemIf(isSensitivePermission())
+          .collect(Collectors.toList());
```
stanislav-jordanov-sonarsource, line 63: Correct me if I am wrong: we are aiming for a future where we won't have to juggle between `Resource`s and `Tree`s (like in this case) but do everything on `Resource`s?

stanislav-jordanov-sonarsource (approved):
Please check my comments&suggestions bellow.
LGTM apart from that.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
nils-werner-sonarsource: WDYT about SENSITIVE_ACTION?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
nils-werner-sonarsource: SENSITIVE_ACTIONS_ELEMENTS?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
nils-werner-sonarsource: I don't get the questions. Can you provide more context regarding these lines?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
stanislav-jordanov-sonarsource: (I've addressed this in the above comment)

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/ExcessivePermissionsCheck.java
stanislav-jordanov-sonarsource: Now I see I did not spot something:
On line 61: replace `resource.list("permissions")` with `permissions` ? (as you've already got it on line 60?)

Answer to your question:
On line 60 you're getting the `Resource` and on line 61 - `List<ExpressionTree>`. Are we aiming for future where we will be getting something like `List<ItemSumbols>` on line 61, so that we don't have to deal both with the `Resource`(s) and `Tree`(s) class hierarchy in a Check class ? (Hint 'No' is a legitimate answer)

nils-werner-sonarsource: Thank you for the clarification. Yes, I think that later we could also have symbols for list items. And we can get away from Tree/AST in the medium term. For now, this is an in-between solution and a baby step to switch from trees to symbols.

PR #320, opened 2022-02-21
SONARIAC-312 Rule S6401: Creating keys without a rotation period is security-sensitive

pynicolas:


iac-extensions/terraform/src/test/resources/checks/CryptoKeyRotationPeriod/gcp.tf
```diff
@@ -0,0 +1,15 @@
+resource "google_kms_crypto_key" "k1" {
+  name            = "k1"
+  key_ring        = google_kms_key_ring.keyring.id
+  rotation_period = "100000s"
+}
+
+resource "google_kms_crypto_key" "k2" { # Noncompliant
```
nils-werner-sonarsource, line 7: Maybe add also a message verification. This way we can prevent the message from being changed by mistake.

nils-werner-sonarsource (approved):
LGTM! See comment

PR #321, opened 2022-02-21
SONARIAC-325 Rule S6329: Allowing public network access to cloud resources is security-sensitive (GCP)

nils-werner-sonarsource:
This PR combines all implementations of the sub-tasks of SONARIAC-325.


pynicolas:
I think the first commit message is wrong:
* The JIRA ticket id is wrong
* The last part should be "Cloudbuild Worker Pool" instead of "Google IAM Policy"

pynicolas (approved):
Please adjust the first commit message if you want to keep the different commits.

PR #323, opened 2022-02-22
SONARIAC-319 Rule S6403: Creating GCP SQL instances without requiring TLS is security-sensitive

nils-werner-sonarsource:


stanislav-jordanov-sonarsource (approved):
LGTM!

PR #327, opened 2022-02-24
SONARIAC-363 Rule S6404: Granting public access to GCP resources is security-sensitive (Google IAM Policy)

nils-werner-sonarsource:


pynicolas (approved):

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/symbols/ReferenceSymbol.java
```diff
@@ -32,8 +32,13 @@
 
 public class ReferenceSymbol extends Symbol<AttributeTree> {
 
-  protected ReferenceSymbol(CheckContext ctx, AttributeTree tree, String name, BlockSymbol parent) {
+  private AttributeAccessTree reference;
+
+  private ReferenceSymbol(CheckContext ctx, AttributeTree tree, String name, BlockSymbol parent) {
     super(ctx, tree, name, parent);
+    if (tree != null) {
+      this.reference = (AttributeAccessTree) tree.value();
```
pynicolas, line 40: What I meant was to cast in the same method where we check `is(TerraformTree.Kind.ATTRIBUTE_ACCESS)`.
Sorry for the misunderstanding.

PR #336, opened 2022-03-04
SONARIAC-369 Rule S6414: Excluding users or groups activities from audit logs is security sensitive

nils-werner-sonarsource:


iac-extensions/terraform/src/test/resources/checks/AuditLogMemberExclusionCheck/gcp_audit_config.tf
```diff
@@ -0,0 +1,25 @@
...
+  }
+}
+
+resource "non_google_project_iam_audit_config" "coverage" {
+  audit_log_config {
+    exempted_members = [
+      "user:eric.therond@sonarsource.com",
+    ]
+  }
+}
```
pynicolas, line 25: I would add a test where `exempted_members` is not a list literal, e.g.;
```
exempted_members = var.something
```
And maybe a case where `exempted_members` is omitted.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/AuditLogMemberExclusionCheck.java
```diff
@@ -0,0 +1,42 @@
...
+  @Override
+  protected void registerResourceConsumer() {
+    register("google_project_iam_audit_config",
+      resource -> resource.blocks("audit_log_config")
+        .forEach(block -> block.list("exempted_members")
+          .reportIf(isNotEmpty(), "Make sure excluding members activity from audit logs is safe here.")));
+  }
+
+  private static Predicate<AttributeTree> isNotEmpty() {
+    return attribute -> attribute != null && !((TupleTree) attribute.value()).elements().trees().isEmpty();
```
pynicolas, line 40: The cast to `TupleTree` seems risky here. 
As a general rule, I prefer to cast only right after having tested that the cast will be safe.
It seems that `ListSymbol` already has a check about `Kind.TUPLE` so I would re-use that one: for example, `ListSymbol` may expose a method named `elements()` or `isEmpty()`.
I'm available to talk about that.

pynicolas (approved):

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/AuditLogMemberExclusionCheck.java
nils-werner-sonarsource: Having a `elements()` method is fine for me.

PR #340, opened 2022-03-07
SONARIAC-355 Rule S6258: Disabling logging is security-sensitive (GCP)

nils-werner-sonarsource:
I moved the AWS-related detection logic to a subclass as well as prefixed all AWS test cases with `aws_` instead of `test_`.


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/GcpDisabledLoggingCheckPart.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static org.sonar.iac.terraform.checks.DisabledLoggingCheck.MESSAGE;
+import static org.sonar.iac.terraform.checks.DisabledLoggingCheck.MESSAGE_OMITTING;
+import static org.sonar.iac.terraform.checks.utils.ExpressionPredicate.equalTo;
+import static org.sonar.iac.terraform.checks.utils.ExpressionPredicate.isFalse;
+import static org.sonar.iac.terraform.checks.utils.ExpressionPredicate.notEqualTo;
+
+public class GcpDisabledLoggingCheckPart extends AbstractNewResourceCheck {
+
+  private static final List<String> DATABASE_FLAGS = List.of("log_connections", "log_disconnections", "log_checkpoints", "log_lock_waits");
```
pynicolas, line 36: Shouldn't be a `Set` rather than a `List`?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/GcpDisabledLoggingCheckPart.java
```diff
@@ -0,0 +1,77 @@
...
+    register("google_container_cluster",
+      resource -> resource.attribute("logging_service")
+        .reportIf(equalTo("none"), MESSAGE));
+
+    register("google_sql_database_instance",
+      resource -> {
+        BlockSymbol settings = resource.block("settings")
+          .reportIfAbsent(MESSAGE_OMITTING);
+
+        List<String> requiredFlags = new ArrayList<>(DATABASE_FLAGS);
```
pynicolas, line 63: Shouldn't be a `Set` rather than a `List`?

pynicolas (approved):

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/gcp/GcpDisabledLoggingCheckPart.java
nils-werner-sonarsource: Due to the fact that we only want to report the first missing block and should be reproducible, we have to use a list.

PR #345, opened 2022-03-14
SONARIAC-385 S5332: Update issue message

nils-werner-sonarsource:


pynicolas (changes requested):
Shouldn't we also update `GcpClearTextProtocolsCheckPart`?

pynicolas (approved):

PR #364, opened 2022-03-17
SONARIAC-380 S6317: Should only raise an issue when all the permissions for a given privilege escalation vector are granted

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/checks/PrivilegeEscalationCheckUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+    static class SimplePermission extends Permission {
+
+      protected SimplePermission(String permissionName) {
+        super(permissionName);
+      }
+
+      public boolean isCoveredBy(Permission other) {
+        if (other instanceof WildCardPermission) {
+          return (permissionName.substring(0, permissionName.indexOf(':') + 1) + "*").equals(other.permissionName);
```
pynicolas, line 96: Minor: we could store the "prefix" of each permission in the constructor.
If we do, we should not crash on an unexpected permission name which wouldn't contain a `:`.

iac-common/src/main/java/org/sonar/iac/common/checks/PrivilegeEscalationCheckUtils.java
```diff
@@ -0,0 +1,109 @@
...
+    new PrivilegeEscalationVector("iam:PassRole", "lambda:CreateFunction", "lambda:AddPermission"),
+    new PrivilegeEscalationVector("iam:PassRole", "lambda:CreateFunction", "lambda:CreateEventSourceMapping"),
+    new PrivilegeEscalationVector("iam:PassRole", "cloudformation:CreateStack"),
+    new PrivilegeEscalationVector("iam:PassRole", "datapipeline:CreatePipeline", "datapipeline:PutPipelineDefinition"),
+    new PrivilegeEscalationVector("iam:PassRole", "glue:CreateDevEndpoint"),
+    new PrivilegeEscalationVector("glue:UpdateDevEndpoint"),
+    new PrivilegeEscalationVector("lambda:UpdateFunctionCode")
+  );
+
+  public static boolean escalationVectorApply(Stream<String> actionPermissions) {
```
pynicolas, line 59: I think the name of this method could better show that it returns a boolean.
Maybe something like `isSupersetOfAnEscalationVector`?

pynicolas (approved):

iac-common/src/main/java/org/sonar/iac/common/checks/PrivilegeEscalationCheckUtils.java
nils-werner-sonarsource: I've added a test case to show that we will not crash on unexpected permission formats.

PR #366, opened 2022-03-17
SONARIAC-333 Rule S6364: Defining a short backup retention duration is security-sensitive (Azure)

nils-werner-sonarsource:


iac-extensions/terraform/src/test/resources/checks/ShortBackupRetentionCheck/azurerm_cosmosdb_account.tf
```diff
@@ -0,0 +1,34 @@
...
+  backup {
+    retention_in_hours = 48  # Noncompliant {{Make sure that defining a short backup retention duration is safe here.}}
+  }
+}
+
+resource "azurerm_cosmosdb_account" "s6364-cdba-c" {
+  backup {
+    retention_in_hours = 168  # 7 * 24
+  }
+}
```
pynicolas, line 22: Why not choosing values which are just at the threshold? e.g. 168 (compliant) and 167 (noncompliant).

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/ShortBackupRetentionCheck.java
```diff
@@ -62,6 +65,22 @@ protected void registerResourceConsumer() {
...
+
+    register("azurerm_app_service",
+      resource -> {
+        BlockSymbol backup = resource.block("backup").reportIfAbsent(OMITTING_MESSAGE);
+        AttributeSymbol enabled = backup.attribute("enabled");
+        if (enabled.isPresent()) {
+          enabled.reportIf(isFalse(), "Make sure disabling backup is safe here.");
+          if (enabled.is(isTrue())) {
+            backup.block("schedule")
+              .reportIfAbsent(String.format(OMITTING_MESSAGE, "schedule.retention_period_in_days"))
```
pynicolas, line 77: I don't see that specified in [SONARIAC-336](https://jira.sonarsource.com/browse/SONARIAC-336).
Same thing for the next call to `reportIfAbsent`, 2 lines below.

pynicolas (approved):
I believe we should add description on the backup_retention_duration properties (for both CloudFormation and Terraform), at least to make it explicit that it's a duration in days (not hours). 
Maybe the property key could have been better chosen in the past, I believe we cannot change it anymore.

PR #369, opened 2022-03-22
SONARIAC-347 Rule S6413: Defining a short log retention duration is security-sensitive (Azure)

nils-werner-sonarsource:


iac-extensions/terraform/src/test/resources/checks/ShortLogRetentionCheck/azure_app_service.tf
```diff
@@ -0,0 +1,71 @@
...
+        retention_in_days = 0
+      }
+    }
+    application_logs {
+      azure_blob_storage {
+        retention_in_days = 0
+      }
+    }
+  }
+}
```
pynicolas, line 53: I don't think this test is useful. I believe we should replace it with 2 tests:
* 1 test to check the behavior of 0 with `azurerm_app_service`
* 1 test to check the behavior of a value such as 13 on something which is not `azurerm_app_service`

iac-extensions/terraform/src/test/resources/checks/ShortLogRetentionCheck/azure_monitor_log_profile.tf
```diff
@@ -0,0 +1,72 @@
+resource "azurerm_monitor_log_profile" "s6413-mlp-nc1" {
+  retention_policy {
+    enabled = true
+    days    = 13 # Noncompliant {{Make sure that defining a short log retention duration is safe here.}}
+  }
+}
+
+resource "azurerm_monitor_log_profile" "s6413-mlp-nc2" {
+  retention_policy {
+    enabled = false # Noncompliant {{Make sure that defining a short log retention duration is safe here.}}
```
pynicolas, line 10: The issue message looks weird to me in this case.

PR #370, opened 2022-03-23
SONARIAC-337 Rule S6258: Disabling logging is security-sensitive (Azure Part II)

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/aws/AzureDisabledLoggingCheckPart.java
```diff
@@ -39,42 +44,73 @@ protected void registerResourceConsumer() {
...
+
+    if (disabledLoggings == 3) {
+      logging.report("Make sure that disabling logging is safe here.");
+      return;
+    } else if(disabledLoggings == 2) {
+      logging.report("Make sure that partially enabling logging is safe here.");
+      return;
+    }
+
+    loggingSettings.forEach(setting -> setting.reportIf(isFalse(), "Make sure that partially enabling logging is safe here."));
```
pynicolas, line 114: There's a kind of duplication of what makes a setting noncompliant (isFalse).
Why not something like the following?
```
    List<AttributeSymbol> disabled = Stream.of("delete", "read", "write")
      .map(logging::attribute)
      .filter(setting -> setting.is(isFalse()))
      .collect(Collectors.toList());
```

pynicolas (approved):

PR #376, opened 2022-06-22
SONARIAC-420 Add Terraform Providers Extension

nils-werner-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/plugin/TerraformProviders.java
```diff
@@ -0,0 +1,98 @@
...
+      }
+    }
+
+    final Version providerVersion;
+
+    public Provider(@Nullable Version providerVersion) {
+      this.providerVersion = providerVersion;
+    }
+
+    public boolean isLower(String version) {
```
pynicolas, line 82: Suggested name: `hasVersionLowerThan`.
Using `has` rather than `is` can help to indicate that the method returns `false` when the provider doesn't have a version.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/plugin/TerraformProviders.java
```diff
@@ -0,0 +1,98 @@
...
+      if (providerVersion ==  null) {
+        return false;
+      }
+      return parseVersion(version).filter(v -> providerVersion.compareTo(v) < 0).isPresent();
+    }
+    
+    private static Optional<Version> parseVersion(String version) {
```
pynicolas, line 89: `parseVersion` and `parseProviderVersion` look similar.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/plugin/TerraformSettings.java
```diff
@@ -61,6 +62,17 @@ public static List<PropertyDefinition> getProperties() {
         .category(TERRAFORM_CATEGORY)
         .multiValues(true)
         .subCategory(GENERAL_SUBCATEGORY)
+        .build(),
+
+      PropertyDefinition.builder(TerraformProviders.Provider.Identifier.AWS.key)
+        .index(101)
+        .name("AWS Provider Version")
+        .description("Version of the AWS provider of lifecycle management of AWS resources," +
+          "including EC2, Lambda, EKS, ECS, VPC, S3, RDS, DynamoDB, and more. Use semantic versioning format.")
```
pynicolas, line 71: Maybe we could add an example of a valid "semantic" version.

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/plugin/TerraformProviders.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public Provider(@Nullable Version providerVersion) {
+      this.providerVersion = providerVersion;
+    }
+
+    public boolean isLower(String version) {
+      if (providerVersion ==  null) {
+        return false;
+      }
+      return parseVersion(version).filter(v -> providerVersion.compareTo(v) < 0).isPresent();
```
pynicolas, line 86: If the `version` parameter was of type `Version` rather than `String`, it could allow rule implementations to parse a version as a string only once per project rather than once per analyzed resource.

iac-extensions/terraform/src/test/resources/checks/UnversionedS3BucketCheck/test_v4.tf
```diff
@@ -0,0 +1,38 @@
+resource "aws_s3_bucket" "mynoncompliantbucket" {
```
pynicolas, line 1: I'd like to have minimal changes between `test_v3.tf` and `test_v4.tf`.
Here the name of the resource is the same for the 2 files even if the resource is compliant.
I think we can find better names for resources, here for an example: `bucket_without_any_versioning`.

iac-extensions/terraform/src/test/resources/checks/UnversionedS3BucketCheck/test_v4.tf
```diff
@@ -0,0 +1,38 @@
...
+  versioning { } # Noncompliant {{Make sure using unversioned S3 bucket is safe here.}}
+# ^^^^^^^^^^
+}
+
+resource "aws_s3_bucket" "mynoncompliantfalsebuckets6252" {
+  #      ^^^^^^^^^^^^^^^> {{Related bucket}}
+  bucket = "mynoncompliantfalsebuckets6252name"
+
+  versioning {
+    enabled = false # Noncompliant {{Make sure using suspended versioned S3 bucket is safe here.}}
```
pynicolas, line 19: Shouldn't we have at least one case with `enabled = true`?

its/plugin/src/test/java/org/sonarsource/iac/PropertiesTest.java
```diff
@@ -50,4 +57,10 @@ private void checkCustomFileIdentifierForLanguage(String projectKey, String lang
       .setProperty("sonar." + language + ".file.identifier", identifier));
     assertThat(getMeasureAsInt(projectKey, "ncloc")).isEqualTo(expectedNcloc);
   }
+
+  private void checkTerraformAwsProviderVersion(String projectKey, String version, int expectedHotspots) {
+    ORCHESTRATOR.executeBuild(getSonarScanner(projectKey, BASE_DIRECTORY + "provider/", "terraform", "aws-provider")
+      .setProperty("sonar.terraform.provider.version.aws", version));
+    assertThat(getHotspotsForProject(projectKey)).hasSize(expectedHotspots);
```
pynicolas, line 64: Rather than the number of issues, I think it would be better to assert the lines on which we expect issues. It would:
* Make the assertions stronger (not green when issues are raised on unexpected lines)
* Make it easier for readers to understand where we expect to raise issues.

PR #381, opened 2022-06-24
SONARIAC-417 [TF] S6245 Use Provider Version Sensitive Analysis in DisabledS3EncryptionCheck

nils-werner-sonarsource:


iac-common/src/test/java/org/sonar/iac/common/testing/VerifierTest.java
```diff
@@ -110,6 +110,20 @@ void issues_list_verifier_failure_not_expected_but_found() {
...
+  @Test
+  void issues_list_verifier_verifyNoIssue() {
+    IacCheck emptyCheck = init -> init.register(Tree.class, (ctx, tree) -> {});
+    TreeParser<Tree> parser = mock(TreeParser.class);
+    when(parser.parse(any(), any())).thenReturn(new AbstractTestTree() {
+      @Override
+      public TextRange textRange() {
+        return TextRanges.range(1, 5, 1, 9);
+      }
+    });
```
pynicolas, line 122: That seems a bit complex. Can't we just parse an empty file?

nils-werner-sonarsource: This test is in the common package, which does not have its own parser. This package is independent of Terraform or Cloudformation. Of course, the test is somewhat complex. However, creating a custom parser just for the test is more complex. What do you think?

pynicolas: Ok, I understand better. The code is a bit complex but it's still ok.

PR #382, opened 2022-06-24
SONARIAC-418 [TF] S6258 Use Provider Version Sensitive Analysis in DisabledLoggingCheck

nils-werner-sonarsource:


iac-extensions/terraform/src/test/resources/checks/DisabledLoggingCheck/aws_apigatewayv2_stage.tf
```diff
@@ -1,8 +1,8 @@
 # Noncompliant@+1 {{Omitting access_log_settings makes logs incomplete. Make sure it is safe here.}}
-resource "aws_api_gatewayv2_stage" "missing_access_log_settings" {
+resource "aws_apigatewayv2_stage" "missing_access_log_settings" {
```
pynicolas, line 2: Doesn't that change deserve a separate ticket rather than being hidden here?

iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/symbols/ListSymbol.java
```diff
@@ -58,12 +60,33 @@ public ListSymbol reportItemIf(Predicate<ExpressionTree> predicate, String messa
...
   }
 
   public boolean isEmpty() {
-    return tree != null && items.isEmpty();
+    return tree != null && items.isEmpty() && !isByReference();
+  }
+
+  @Override
+  public boolean isPresent() {
+    return tree != null && !isByReference();
```
pynicolas, line 80: I don't understand the test on `isByReference`.

nils-werner-sonarsource: A list is the value of an argument. Argument values can also be defined via reference. So that there are no false positives, should one decide on the absence of a list, it must also be checked whether the list is not included by a reference.

pynicolas: My concern is that the API is not obvious.
For example, it seems that `isPresent` and `reportIfAbsent` are not consistent for a `ListSymbol`.
It's not a big deal for now, but I think it's a sign that something should be refined.

PR #393, opened 2022-07-06
SONARIAC-409 Rule S6428: Enabling privileged mode on containers is security-sensitive

nils-werner-sonarsource:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/symbols/TupleBlockSymbol.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private final HasTextRange toHighlight;
+
+  protected TupleBlockSymbol(CheckContext ctx, @Nullable HasTextRange toHighlight, MappingTree block, String key, BlockSymbol<?,?> parent) {
+    super(ctx, block, key, parent);
+    this.toHighlight = toHighlight;
+  }
+
+  public static TupleBlockSymbol fromPresent(CheckContext ctx, YamlTree tree, String key, BlockSymbol<?,?> parent) {
+    if (tree instanceof TupleTree && ((TupleTree) tree).value() instanceof MappingTree) {
```
pynicolas, line 40: We have this pattern in multiple places: either the tree is what we expect it to be or we consider it to be absent.
I wonder if we miss a distinction between:
* it's really absent
* it's there, but not as we expect (not the type we expect)

I wonder if that could lead to FPs in the end.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/symbols/TupleSymbol.java
```diff
@@ -0,0 +1,61 @@
...
+package org.sonar.iac.kubernetes.symbols;
+
+import java.util.function.Predicate;
+import javax.annotation.Nullable;
+import org.sonar.iac.common.api.checks.CheckContext;
+import org.sonar.iac.common.api.tree.HasTextRange;
+import org.sonar.iac.common.yaml.tree.TupleTree;
+import org.sonar.iac.common.yaml.tree.YamlTree;
+
+public class TupleSymbol extends KubernetesSymbol<TupleSymbol, TupleTree> {
```
pynicolas, line 29: I don't understand what's the relation between `TupleSymbol` and `TupleBlockSymbol`.

pynicolas (changes requested):

pynicolas:
Why not have similar abstractions to what we have in Terraform, for example, the notion of Attribute?
Even if the syntax is slightly different, my understanding is that we should be able to represent it in a similar way.

iac-common/src/main/java/org/sonar/iac/common/yaml/YamlConverter.java
```diff
@@ -70,8 +67,9 @@ public FileTree convertFile(List<Node> nodes) {
     if (nodes.isEmpty()) {
       throw new ParseException("Unexpected empty nodes list while converting file", null);
     }
-
-    return new FileTreeImpl(convert(nodes.get(0)), range(nodes.get(0)));
+    TextRange fileRange = TextRanges.merge(List.of(range(nodes.get(0)), range(ListUtils.getLast(nodes))));
+    YamlTreeMetadata metadata = new YamlTreeMetadata("FILE", fileRange, Collections.emptyList());
```
pynicolas, line 71: Can we use `YamlTreeMetadata.fromNodes` here?

iac-common/src/main/java/org/sonar/iac/common/yaml/tree/YamlTreeMetadata.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private static TextRange range(Optional<Mark> startMark, Optional<Mark> endMark) {
+    if (startMark.isEmpty()) {
+      throw new ParseException("Nodes are expected to have a start mark during conversion", null);
+    }
+
+    // endMark is not present. This happens for example when we have a file with only a comment.
+    // in that case, the root node will be an empty MappingNode with only a startMark to which the comment is attached
+    return endMark.map(mark -> TextRanges.range(startMark.get().getLine() + 1, startMark.get().getColumn(), mark.getLine() + 1, mark.getColumn()))
+      .orElseGet(() -> TextRanges.range(startMark.get().getLine() + 1, startMark.get().getColumn(), startMark.get().getLine() + 1, startMark.get().getColumn()));
```
pynicolas, line 108: Maybe creating variables could help readability:
```
int startLine = startMark.get().getLine() + 1;
int startColumn = startMark.get().getColumn();
```

iac-extensions/terraform/src/test/resources/checks/Azure/CertificateBasedAuthenticationCheck/api_management.tf
```diff
@@ -16,6 +16,14 @@ resource "azurerm_api_management" "example" {
...
 resource "azurerm_api_management" "example" {
   sku_name = "Consumption_" # Compliant
+
+  foo {
+    foo = bar
+  }
+
+  foo {
+    foo = bar
+  }
```
pynicolas, line 26: Is that an intentional change?

pynicolas (approved):

iac-extensions/terraform/src/test/resources/checks/Azure/CertificateBasedAuthenticationCheck/api_management.tf
nils-werner-sonarsource: Good catch. It's a leftover from our conversation.

PR #395, opened 2022-07-08
SONARIAC-323: Unsimulated intrinsic functions lead to false positives and negatives

marco-bearzi-sonarsource:


iac-extensions/cloudformation/src/test/resources/checks/BucketsInsecureHttpCheck/test.yaml
```diff
@@ -237,16 +237,16 @@ Resources:
...
     Properties:
-      BucketName: "mynoncompliantbucket6249"
-  NoncomplianPolicyDueToUnresolvedRelation:
+      BucketName: "mycompliantbucket6249"
+  CompliantPolicyDueToResolvedRelation:
     Type: 'AWS::S3::BucketPolicy'
     Properties:
       # Bucket relation is not resolved. However, a function call, which is not ref or sub, would not make sense here.
-      Bucket: !Join ['', ['mynoncompliantbucket6249']]
+      Bucket: !Join [ '', [ 'mycompliantbucket6249' ] ]
```
nils-werner-sonarsource, line 249: Maybe also add an example with the full function name like:
```
Bucket:
        Fn::Join: [ '', [ 'mycompliantbucket6249' ] ]
```

nils-werner-sonarsource (approved):
LGTM! Please address the comment about the empty string as a fallback during resource relationship resolution.

PR #397, opened 2022-07-11
SONARIAC-411 [K8] Rule S6430: Allowing process privilege escalation is security-sensitive

marco-bearzi-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #398, opened 2022-07-12
SONARIAC-412 [K8] Rule S5849: Setting capabilities is security-sensitive

marco-bearzi-sonarsource:


nils-werner-sonarsource (commented):
To be more consistence with the implementation of the Terraform extension I would suggest to build a ListObject similar the ListSymbol. Your current implementation is hardly connected to have strings only as list/sequence elements which could bring some limitations.

iac-common/src/test/java/org/sonar/iac/common/yaml/TreePredicatesTest.java
```diff
@@ -44,11 +48,28 @@ void isEqualTo() {
...
     return new FileTreeImpl(null, null);
   }
+
+  private YamlTree yaml(List<String> values) {
+    if (values.size() <= 1) {
+      return new ScalarTreeImpl(values.stream().findFirst().orElse(""), null, null);
+    }
+    List<YamlTree> elements = values.stream().map(v -> new ScalarTreeImpl(v, null, null)).collect(Collectors.toList());
+    return new SequenceTreeImpl(elements, null);
+  }
```
nils-werner-sonarsource, line 74: Same comment as above. Since both test classes are in the same module, you could also move this method to a utils class.

iac-common/src/test/java/org/sonar/iac/common/yaml/object/BlockObjectTest.java
```diff
@@ -93,6 +94,19 @@ void attribute() {
     assertThat(absentAttr.status).isEqualTo(YamlObject.Status.ABSENT);
   }
 
+  @Test
+  void list() {
+    BlockObject block = BlockObject.fromPresent(ctx, parseMap("foo: [bar, car]"), "a");
+    ListObject listPresent = block.list("foo");
+    assertThat(listPresent.status).isEqualTo(YamlObject.Status.PRESENT);
+    assertThat(listPresent.getItemIf(e -> true)
```
nils-werner-sonarsource, line 102: `getItemIf` is only used for testing right now. Maybe you can make the field visible for testing by making it package protected so you can access the items without the predicate.

iac-common/src/test/java/org/sonar/iac/common/yaml/object/ListObjectTest.java
```diff
@@ -0,0 +1,98 @@
...
+  CheckContext ctx = mock(CheckContext.class);
+  private final TupleTree tree = ((MappingTree) PARSER.parse("my_list : [\"my_item\", a]", null).root()).elements().get(0);
+
+
+  @Test
+  void report_fromPresent() {
+    ListObject list = ListObject.fromPresent(ctx, tree, "my_list", null);
+    assertThat(list.key).isEqualTo("my_list");
+    assertThat(list.isEmpty()).isFalse();
+    assertThat(list.getItemIf(e -> true)).hasSize(2);
```
nils-werner-sonarsource, line 50: Same comment as above.

iac-common/src/main/java/org/sonar/iac/common/checks/TextUtils.java
```diff
@@ -68,4 +73,22 @@ public static boolean isValueTrue(@Nullable Tree tree) {
...
+      return Collections.emptyList();
+    }
+  }
+
+  private static List<String> getValuesOfSequenceTree(SequenceTree tree) {
+    return tree.elements().stream()
+      .map(TextUtils::getListValueElements)
+      .flatMap(List::stream)
+      .collect(Collectors.toList());
+  }
```
nils-werner-sonarsource, line 92: I'm not happy with the location. The methods handling Yaml trees. This utils class is also used by Terraform and was intended to handle all trees which implement the TextTree interface, which does not make use of these trees. I would prefer having a new utils class in the YAML package like `YamlTreeUtils`

nils-werner-sonarsource: I would say that we don't need this method for the check you've implemented. The `isSet` predicate always gets a ScalarTree in the context of the `reportIfAnyItem` you can directly handle as a string. Can you confirm?

iac-common/src/main/java/org/sonar/iac/common/yaml/TreePredicates.java
```diff
@@ -35,4 +39,10 @@ public static Predicate<YamlTree> isTrue() {
   public static Predicate<YamlTree> isEqualTo(String parameter) {
     return t -> TextUtils.isValue(t, parameter).isTrue();
   }
+
+  public static Predicate<YamlTree> isSet() {
+    return t -> TextUtils
+      .getListValueElements(t).stream()
```
nils-werner-sonarsource, line 45: See the comment at the `getListValueElements` declaration. The predicate
`TextUtils.matchesValue(t, value -> !STRINGS_CONSIDERED_AS_EMPTY.contains(value)).isTrue()` does the same job.

nils-werner-sonarsource (approved):
LGTM! Please see my comments. IMO there is some code you can get rid of.

PR #403, opened 2022-07-13
SONARIAC-91 Add complete string parsing validation in TerraformTreeModelTest

nils-werner-sonarsource:


iac-extensions/terraform/src/test/java/org/sonar/iac/terraform/tree/impl/BlockTreeImplTest.java
```diff
@@ -42,7 +42,7 @@ void empty_block() {
 
   @Test
   void simple_block() {
-    BlockTree tree = parse("a{\n b = true \nc = null}", HclLexicalGrammar.BLOCK);
```
nils-werner-sonarsource, line 45: Spaces at the line end are not be parsed so also not able to be rebuild

PR #416, opened 2022-08-08
HOTFIX 1.9.2: SONARIAC-435 Do not raise parser errors on Helm syntax template directives

nils-werner-sonarsource:
DO NOT MERGE ON MASTER
Please take into account, that the first 2 commits are already released. The release will be done from this branch and only a cherry-pick of the fix will be on master.

### DIFF
--- a/iac-common/pom.xml
+++ b/iac-common/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.sonarsource.iac</groupId>
     <artifactId>iac</artifactId>
-    <version>1.9.1-SNAPSHOT</version>
+    <version>1.9.2-SNAPSHOT</version>
   </parent>
 
   <artifactId>sonar-iac-common</artifactId>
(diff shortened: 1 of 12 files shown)


.gitmodules
```diff
@@ -19,3 +19,6 @@
 [submodule "its/sources/kubernetes/examples"]
 	path = its/sources/kubernetes/examples
 	url = https://github.com/kubernetes/examples
+[submodule "its/sources/kubernetes/helm"]
+	path = its/sources/kubernetes/helm
+	url = https://github.com/helm/examples.git
```
andreaguarino, line 24: Are we ok to depend directly on an external git repository? I think for other analyzers we usually have a fork under SonarSource organization. 
However I see it's already the case for other submodules, so I guess this is intentional.

andreaguarino (commented):
the implementation LGTM!
I have a concern regarding ruling test: I see that you added helm examples submodule, to be analyzed during ruling. 
 
My understanding is that we don't want those helm files to be parsed, so I'm not sure about including them in ruling without any further assertion. I guess that even if we parse those files with or without a parse error, the ruling test would still be green, right?

If my previous assumption / understanding is correct, I would suggest to either add an assertion on kubernetes ruling (or a new integration test), testing that there are no parse errors on helm files , or to completely drop this submodule. Wdyt?

PR #420, opened 2022-11-10
SONARIAC-441 Create Docker extension infrastructure

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (changes requested):
You have done more than is described in the ticket. All classes that have to do with parsing are not yet relevant for the infrastructure. They also have no coverage. If we now merge the PR, we would not be able to determine in future PRs whether we are using all the classes you have created. You can stay in the PR, but it's not the right baby-step approach.

Also, you blindly copied the provider, even though it has no application in the Docker ecosystem. Please remove it.

Lastly, I would like to ask for more coverage on the sensor. Here it is already possible to query all relevant methods to validate return values.

All in all, the PR is okay, but please stick more to the specification next time. Parser and provider are not part of it.

nils-werner-sonarsource (approved):
LGTM!

iac-extensions/docker/src/test/java/org/sonar/iac/docker/plugin/DockerRulesDefinitionTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.sonar.iac.docker.checks.DockerCheckList;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class DockerRulesDefinitionTest {
+
+  @Test
+  void testActivationSonarLint() {
+    RulesDefinition.Repository repository = dockerRuleRepository(9, 3);
+    assertThat(repository).isNotNull();
```
mstachniuk, line 38: Hint, minor. There is no need to call `isNotNull()` since in the next line is `isEqualTo` check which is more precise (and also check is not null)

mstachniuk (approved):
LGTM

PR #421, opened 2022-11-11
SONARIAC-474 Docker: parse empty file

mstachniuk:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -0,0 +1,61 @@
...
+  public DockerGrammar(GrammarBuilder<SyntaxToken> b, TreeFactory f) {
+    this.b = b;
+    this.f = f;
+  }
+
+  public FileTree FILE() {
+    return b.<FileTree>nonterminal(DockerLexicalGrammar.FILE).is(
+      f.file(
+        b.zeroOrMore(INSTRUCTION()),
+        b.token(DockerLexicalGrammar.SPACING),
```
rudy-regazzoni-sonarsource, line 43: I believe the final spacing should be Optional.
```suggestion
        b.optional(b.token(DockerLexicalGrammar.SPACING)),
```

iac-common/src/main/java/org/sonar/iac/common/parser/grammar/LexicalConstant.java
```diff
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-package org.sonar.iac.terraform.parser.grammar;
+package org.sonar.iac.common.parser.grammar;
```
rudy-regazzoni-sonarsource, line 20: I fear there may be some Lexical differences between HCL/Terraform and DockerFile.
I found one about comments :
- HCL support #, // and /* */
- Docker support only #

So not sure if we can completely move this Lexical class in common.
Maybe we can do so for mutual elements, and have two child class (HclLexicalConstant and DockerLexicalConstant) inheriting from it and completing with their particularity.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammar.java
```diff
@@ -0,0 +1,84 @@
...
+  private static void punctuators(LexerlessGrammarBuilder b) {
+    for (Punctuator p : Punctuator.values()) {
+      b.rule(p).is(SPACING, p.getValue()).skip();
+    }
+  }
+  private static void lexical(LexerlessGrammarBuilder b) {
+    b.rule(SPACING).is(
+      b.skippedTrivia(b.regexp("[" + LexicalConstant.LINE_TERMINATOR + LexicalConstant.WHITESPACE + "]*+")),
+      b.zeroOrMore(
+        b.commentTrivia(b.regexp(LexicalConstant.COMMENT)),
```
rudy-regazzoni-sonarsource, line 68: As mentioned in my previous comment, by doing so here we are parsing comment has either #, // or /* */, while in docker only # is supported.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/DockerTreeImpl.java
```diff
@@ -0,0 +1,42 @@
...
+      if (getKind() == kindIter) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public TextRange textRange() {
+    return TextRanges.range(0, 0, "");
```
rudy-regazzoni-sonarsource, line 40: I would add a TODO comment here to mention that this implementation is TBD.

rudy-regazzoni-sonarsource (commented):
Nice work ! There was a lot of work to be done but now we have a great base to start adding things :D
See my few comments, I liked the fact that you are reusing/mutualizing some classes regarding grammar, just maybe there is some elements that would need to be adapted (comment).

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
mstachniuk: If you look into `TreeFactory.file()` you can see it's ignored. @nils-werner-sonarsource Is there any reason to keep it here?

iac-common/src/main/java/org/sonar/iac/common/parser/grammar/LexicalConstant.java
mstachniuk: Nils suggested moving this class there. But let's create `DockerLexicalConstant` for Docker specific things.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammar.java
mstachniuk: Good catch, changed to `DockerLexicalConstant.COMMENT`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
nils-werner-sonarsource: We don't need to know about the spacing, but there can be spacing between the last instruction and the EOF. One option could be to add spacing before the EOF element while defining the rule. But then the EOL token could be more than a single position. It's not important that it's optional, because the regex will always trigger and create an optional token because it's accepting empty chars as well.

iac-common/src/main/java/org/sonar/iac/common/parser/grammar/LexicalConstant.java
nils-werner-sonarsource: Yeah, we can somehow create our own rule for comments in docker. It can be done directly in the DockerLexicalGrammar

PR #422, opened 2022-11-14
SONARIAC-460 Docker: parse MAINTAINER instruction (deprecated) - basic implementation with StringLiteral

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -27,5 +27,9 @@ private DockerLexicalConstant() {
   }
 
   public static final String COMMENT = "(?:" + LexicalConstant.SINGLE_LINE_COMMENT_HASH + ")";
+  public static final String STRING_LITERAL_WITH_QUOTES = "\"(?:[^\"\\\\]*+(?:\\\\[\\s\\S])?+)*+\"";
+  public static final String STRING_LITERAL_WITHOUT_QUOTES = ".+";
```
mstachniuk, line 31: This will match too much, e.g. new lines what is not expected here

rudy-regazzoni-sonarsource: Mmhh you mean for the `.+` ? Normally it shouldn't, from my understanding `.` match anything except newline.
Or do you refer to the `[\\s\\S]` in `STRING_LITERAL_WITH_QUOTES` ? I must admit I stole this one somewhere (Sonar-Python ?) ^^' It seems quite legit to me at first to match Strings with `"`, but as I don't have a full understanding of docker syntax I may miss something ._.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
```diff
@@ -38,4 +40,8 @@ public FileTree file(Optional<List<InstructionTree>> instructions, Optional<Synt
   public FromTree from(SyntaxToken token) {
     return new FromTreeImpl();
   }
+
+  public MaintainerTree maintainer(SyntaxToken maintainerToken, SyntaxToken authorToken) {
+    return new MaintainerTreeImpl(maintainerToken, authorToken);
```
nils-werner-sonarsource, line 45: Based on the example from the ticket the instruction allows more tokens `MAINTAINER Bob BobFullName bob@bob.com` Or is only the first token a valid token and the rest is dropped? If yes, we have to handle them as well. If not, we should also store them.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImpl.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public MaintainerTreeImpl(SyntaxToken maintainerToken, SyntaxToken authorToken) {
+    this.maintainerToken = maintainerToken;
+    this.authorToken = authorToken;
+  }
+
+  @Override
+  public TextRange textRange() {
+    return TextRanges.merge(List.of(maintainerToken.textRange(), authorToken.textRange()));
+  }
```
nils-werner-sonarsource, line 42: Why do we need to override this method? Isn't `org.sonar.iac.docker.tree.impl.DockerTreeImpl#textRange` doing the job?

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Test;
+import org.sonar.iac.docker.parser.grammar.DockerLexicalGrammar;
+import org.sonar.iac.docker.tree.api.DockerTree;
+import org.sonar.iac.docker.tree.api.MaintainerTree;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.sonar.iac.docker.tree.impl.DockerTestUtils.parse;
+
+class MaintainerTreeImplTest {
```
nils-werner-sonarsource, line 32: Maybe you can also build tests like `org.sonar.iac.terraform.parser.FileTest`. So you can test multiline and antipatterns much easier.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
mstachniuk: The build IntelliJ "Check RegExp" cheat me - there is DOTALL mode enabled where `.` matches a new line. Sorry for that.

Anyway, we have to think if we would store input `foo bar` as a single StringLiteral or two StringLiterals...
And we have to find another name because StringLiteral refers to the text in double quotes (See: `LexicalConstant.STRING_LITERAL`. In our cases, we usually have / should have a single word without quotes (but the Docker parser also accepts multiple words and multiple words surrounded with double quotes).

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImpl.java
rudy-regazzoni-sonarsource: Good point ! I started overriding it before overriding `children()` and I didn't connect the dots ^^'

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
rudy-regazzoni-sonarsource: Indeed it is taking all tokens -> [https://www.java4coding.com/contents/docker/docker-maintainer-command](https://www.java4coding.com/contents/docker/docker-maintainer-command)
Actually in the second `SyntaxToken`, you will have all the arguments including the spaces : `Bob BobFullName bob@bob.com`
This is coming from the way I defined it in the grammar : I'm matching everything else until end of line.

I choose to do it this way but I'm not really sure if that's a good idea. Should I go the other way as we will do for every other method, meaning we catch every token separately and store them like this ?
I'm also not sure how are handled multiple spaces, couldn't find any example. I guess it's also a good idea to keep them.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.sonar.iac.common.api.tree.Tree;
+import org.sonar.iac.docker.tree.api.MaintainerTree;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+
+public class MaintainerTreeImpl extends DockerTreeImpl implements MaintainerTree {
+  private final SyntaxToken maintainerToken;
+  private final List<SyntaxToken> authorsToken;
```
nils-werner-sonarsource, line 31: IMO `maintainer` and `author` are enough. The type specifies that's a token.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public MaintainerTreeImpl(SyntaxToken maintainerToken, List<SyntaxToken> authorsToken) {
+    this.maintainerToken = maintainerToken;
+    this.authorsToken = authorsToken;
+  }
+
+  @Override
+  public List<Tree> children() {
+    return Stream.concat(Stream.of(maintainerToken), authorsToken.stream()).collect(Collectors.toList());
```
nils-werner-sonarsource, line 40: Using two streams to combine elements into a list is not performed.
```
List<Tree> children = new ArrayList<>(maintainerToken);
children.addAll(authorsToken);
return children;

File: iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImpl.java [hunk 7]
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Override
+  public Kind getKind() {
+    return Kind.MAINTAINER;
+  }
+
+  @Override
+  public SyntaxToken maintainerToken() {
```
nils-werner-sonarsource, line 49: Here the same. I think we don't need the addition `Token`

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+  @Test
+  void matchingSimple() {
+    Assertions.assertThat(DockerLexicalGrammar.MAINTAINER)
+      .matches("MAINTAINER bob")
+      .notMatches("maintainer bob") // should match
```
nils-werner-sonarsource, line 36: Please create a ticket that the keywords are case insensitive.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -27,5 +27,8 @@ private DockerLexicalConstant() {
   }
 
   public static final String COMMENT = "(?:" + LexicalConstant.SINGLE_LINE_COMMENT_HASH + ")";
+  public static final String STRING_LITERAL_WITH_QUOTES = "\"(?:[^\"\\\\]*+(?:\\\\[\\s\\S])?+)*+\"";
+  public static final String STRING_LITERAL_WITHOUT_QUOTES = "\\S+";
+  public static final String STRING_LITERAL = "(?:"+STRING_LITERAL_WITH_QUOTES+")|(?:"+STRING_LITERAL_WITHOUT_QUOTES+")";
```
mstachniuk, line 32: I will add spaces around `+` for better readability

iac-extensions/docker/src/test/java/org/sonar/iac/docker/parser/utils/Assertions.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public static ParserAssert assertThat(LexerlessGrammarBuilder b, GrammarRuleKey rule) {
+    return new ParserAssert(new ActionParser<>(
+      StandardCharsets.UTF_8,
+      b,
+      DockerGrammar.class,
+      new TreeFactory(),
+      new DockerNodeBuilder(),
+      rule));
+  }
```
mstachniuk, line 59: This can be simplified to:
```
  public static ParserAssert assertThat(GrammarRuleKey rule) {
    return new ParserAssert(new DockerParser(rule));
  }
```

mstachniuk (approved):
LGTM, I can do small suggested fixes with my PR (when it will works)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
rudy-regazzoni-sonarsource: As mentioned in slack, I changed this in the new commit to have each token separately, as we will probably do for other instructions.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+      .matches("MAINTAINER \"bob\"")
+      .matches("MAINTAINER \"bob")
+      .matches("MAINTAINER bob boberman bob@bob.com")
```
nils-werner-sonarsource, line 39: We should also have a ticket to address new lines which are allowed to have here like
```
MAINTAINER /
bob
```

nils-werner-sonarsource (approved):
LGTM! Please see my comments.

PR #432, opened 2022-11-17
SONARIAC-464 Docker: parse ADD instruction

rudy-regazzoni-sonarsource:
Note : this version of ADD instruction parsing does not support the JSON argument format, waiting for its implementation in the CMD ticket.
Specific ticket was created for this : https://sonarsource.atlassian.net/browse/SONARIAC-485


nils-werner-sonarsource (commented):
Please see my comments. And you should rebase. Let me have a second look afterwards.

rudy-regazzoni-sonarsource:
I removed the OptionTree classes to use only KeyValuePairTree ones, while extending them to support the possibility of having prefix (-- and possibly later more).

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/AddTreeImplTest.java
```diff
@@ -0,0 +1,163 @@
...
+    assertThat(tree.srcs()).hasSize(1);
+    assertThat(tree.srcs().get(0).value()).isEqualTo("\"src\"");
+    assertThat(tree.dest().value()).isEqualTo("\"dest\"");
+  }
+
+  @Test
+  void addInstructionNoSrc() {
+    AddTree tree = parse("ADD dest", DockerLexicalGrammar.ADD);
+    assertThat(tree.getKind()).isEqualTo(DockerTree.Kind.ADD);
+    assertThat(tree.keyword().value()).isEqualTo("ADD");
```
nils-werner-sonarsource, line 82: For the next time. For me, it would be okay, to only test what has changed, compared to the previous test. If we don't expect, that the kind and keyword changed, it's okay to skip this assertion.

nils-werner-sonarsource (approved):
LGTM! Maybe you want to use the `ParamTree` instead of overloading the KeyValuePair.
However, I realized, that may implementation of the prefix is wrong for `ParamTree`. We should use "--" instead of two `Punctuator.MINUS` which allows whitespace between them. I leave it up to you. We can discuss which one to use.

rudy-regazzoni-sonarsource:
> LGTM! Maybe you want to use the `ParamTree` instead of overloading the KeyValuePair. However, I realized, that may implementation of the prefix is wrong for `ParamTree`. We should use "--" instead of two `Punctuator.MINUS` which allows whitespace between them. I leave it up to you. We can discuss which one to use.

I agree on this, I will remove the KeyValuePair recent change and use ParamTree instead now that it is available :D I will take this opportunity to fix this Punctuator thing.

rudy-regazzoni-sonarsource:
@nils-werner-sonarsource I would like a quick second review from you, as I modified `ParamTree` to accept null equals/values to support toggle option (`--link` for example), just want to make sure that it's ok for you 👍

nils-werner-sonarsource (approved):
LGTM!

PR #434, opened 2022-11-17
SONARIAC-470 Docker: parse USER instruction

rudy-regazzoni-sonarsource:
I haven chosen the following approach, open to be challenged :
- I'm basing on the reference, which mention this syntax `USER <user>[:<group>]`, although we can put anything in argument (multiple user, groups, ...) as in the end the docker build only put those value in the field Config.User (see `UserTreeImplTest` for a view of how it is handled)
- I store only one value (`USER` + `<user>` + `:` + `<group>`) directly in the class UserTree, so I didn't use the KeyValuePair class although it is a possibility
- I had to create another `STRING_LITERAL_NO_COLON` as we have here a new character to stop on, we will really need to refactor this


iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/UserTreeImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+class UserTreeImplTest {
+  @Test
+  void matchingSimple() {
+    Assertions.assertThat(DockerLexicalGrammar.USER)
+      .matches("USER bob")
+      .matches("USER bob:group")
+      .matches("USER bob:group bob2")
+      .matches("user bob")
+      .notMatches("USER")
```
mstachniuk, line 39: Could you add more negative cases? Like `USERR`, `USER\nbob`, `""`, etc.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/UserTreeImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Test
+  void userInstructionSimple() {
+    UserTree tree = parse("USER bob", DockerLexicalGrammar.USER);
+    assertThat(tree.getKind()).isEqualTo(DockerTree.Kind.USER);
+    assertThat(tree.keyword().value()).isEqualTo("USER");
+    assertThat(tree.textRange().start().line()).isEqualTo(1);
+    assertThat(tree.textRange().start().lineOffset()).isZero();
+    assertThat(tree.textRange().end().line()).isEqualTo(1);
+    assertThat(tree.textRange().end().lineOffset()).isEqualTo(8);
```
mstachniuk, line 51: Hint. 

```
    TextRangeAssert.assertTextRange(tree.textRange()).hasRange(1, 0, 1, 8);
```

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/UserTreeImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    assertThat(tree.children()).hasSize(4);
+    assertThat(tree.user().value()).isEqualTo("bob");
+    assertThat(tree.colon().value()).isEqualTo(":");
+    assertThat(tree.group().value()).isEqualTo("group");
+  }
+
+  @Test
+  void userInstructionAttemptMultipleUser() {
+    // if there is something after the group name, we store everything after the comma in the group field.
+    // Docker does not even differentiate user and group at the build step, it all get stored in the user field (no group field at all).
```
mstachniuk, line 72: +1

mstachniuk (approved):
LGTM, please see my comments. Your approach is ok for me, however I would like to know @nils-werner-sonarsource opinion.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -172,6 +175,15 @@ public EnvTree ENV() {
     );
   }
 
+  public UserTree USER() {
+    return b.<UserTree>nonterminal(DockerLexicalGrammar.USER).is(
+      b.firstOf(
+        f.user(b.token(DockerKeyword.USER), b.token(STRING_LITERAL_NO_COLON), b.token(Punctuator.COLON), b.token(STRING_UNTIL_EOL)),
+        f.user(b.token(DockerKeyword.USER), b.token(STRING_UNTIL_EOL))
```
nils-werner-sonarsource, line 182: Maybe you have a look at the [grammar](https://github.com/camdencheek/tree-sitter-dockerfile/blob/main/src/grammar.json#L477) to avoid haveing yet another string literal pattern. And so you could use Optionals.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -29,7 +29,9 @@ private DockerLexicalConstant() {
   public static final String COMMENT = "(?:" + LexicalConstant.SINGLE_LINE_COMMENT_HASH + ")";
   public static final String STRING_LITERAL_WITH_QUOTES = "\"(?:[^\"\\\\]*+(?:\\\\[\\s\\S])?+)*+\"";
   public static final String STRING_LITERAL_WITHOUT_QUOTES = "[^\\s=]+";
```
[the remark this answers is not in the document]
nils-werner-sonarsource: See comment above and take discussion about dedicated patterns into account

nils-werner-sonarsource (commented):
Please check the grammar patterns to reduce the technical depth.

rudy-regazzoni-sonarsource:
I introduced the grammar as in the other file.
This has some change :
- it does not support anymore invalid USER instruction (invalid on execution, not on build)
- the regex I took from the grammar.json does not seems to support quotes. Again I couldn't find any example and since docker build does not perform any control on the USER instruction I couldn't check this myself, but we may be careful about this.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
```diff
@@ -133,6 +135,10 @@ public KeyValuePairTree keyValuePairEquals(SyntaxToken key, SyntaxToken equals,
     return new KeyValuePairTreeImpl(key, equals, value);
   }
 
+  public UserTree user(SyntaxToken keyword, SyntaxToken user, Optional<SyntaxToken> colon, Optional<SyntaxToken> group) {
```
nils-werner-sonarsource, line 138: WDYT about using a tuple here? Colon and group are always a pair. In the constructor, you can separate them again.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -190,6 +190,13 @@ public EnvTree ENV() {
     );
   }
 
+  public UserTree USER() {
+    return b.<UserTree>nonterminal(DockerLexicalGrammar.USER).is(
+      f.user(b.token(DockerKeyword.USER), b.token(DockerLexicalGrammar.USER_NAME),
+        b.optional(b.token(Punctuator.COLON)), b.optional(b.token(DockerLexicalGrammar.USER_GROUP)))
```
nils-werner-sonarsource, line 196: `Punctuator.COLON` will allow whitespace before the character. Is this intended?

nils-werner-sonarsource, line 196: See the comment above. Maybe a tuple makes more sense. Right now, a single colon without group would also be allowed.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -29,7 +29,9 @@ private DockerLexicalConstant() {
...
+  public static final String STRING_LITERAL_WITHOUT_QUOTES_NO_COLON = "[^\\s:]+";
```
nils-werner-sonarsource, line 32: Do you actually use this constant? Same for `STRING_LITERAL_NO_COLON`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/UserTreeImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+import org.sonar.iac.common.api.tree.Tree;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+import org.sonar.iac.docker.tree.api.UserTree;
+
+public class UserTreeImpl extends InstructionTreeImpl implements UserTree {
+  private final SyntaxToken user;
```
nils-werner-sonarsource, line 30:
```suggestion

  private final SyntaxToken user;
```

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/UserTreeImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+      .matches("USER bob")
+      .matches("USER bob:group")
+      .matches("user bob")
+      .matches("USER $var")
+      .matches("USER ${var}")
+      .matches("USER ${var with space}")
+      .matches("USER $var:group")
+      .matches("USER $var:$group")
+      .matches("USER ${var}:${group}")
+      .matches("USER\nbob")
```
nils-werner-sonarsource, line 45: Shouldn't match!

nils-werner-sonarsource (changes requested):
Please see my comments. I think you have to adjust your logic because a line break should not match.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
rudy-regazzoni-sonarsource: Excellent suggestion ! That's what I was looking for, I was not satisfied with this.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
rudy-regazzoni-sonarsource: I didn't saw that one coming ^^' I replacing this and added a test case 👍

rudy-regazzoni-sonarsource: I put the tuple in place and added a test case for it.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
rudy-regazzoni-sonarsource: No more needed indeed, thanks.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammar.java
```diff
@@ -130,6 +137,12 @@ private static void lexical(LexerlessGrammarBuilder b) {
     b.rule(PARAM_PREFIX).is(Punctuator.MINUS, Punctuator.MINUS);
     b.rule(PARAM_NAME).is(b.regexp("[a-z][-a-z]*+"));
     b.rule(PARAM_VALUE).is(b.regexp("[^\\s]+"));
+
+    b.rule(USER_STRING).is(b.regexp("(?:[a-z][-a-z0-9_]*|[0-9]+)"));
+    b.rule(USER_VARIABLE).is(b.regexp("\\$(?:[a-zA-Z_][a-zA-Z0-9_]*|\\{[^}]+\\})"));
```
nils-werner-sonarsource, line 142: These regexes will not match `USER foo${var}` which is also valid. However, I guess it will only be solved with [SONARIAC-486](https://sonarsource.atlassian.net/browse/SONARIAC-486).

nils-werner-sonarsource (approved):
The new regex will limit the parser. TBH I don't now if it makes sense to take the problem with global replacements in this ticket. Maybe you drop the support so far and we take it later.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammar.java
rudy-regazzoni-sonarsource: Agree on this. I followed the grammar too strictly and it does not support such use case, but anyway we will need to review our way of handling arguments, so I prefer to drop it there.

PR #438, opened 2022-11-18
SONARIAC-457 CMD Basic implementation to support shell form

mstachniuk:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CmdTreeImplTest.java
```diff
@@ -67,7 +86,7 @@ void shouldCheckParseCmdTree() {
   }
```
rudy-regazzoni-sonarsource, line 87: You are missing test case for shell form, which would highlight a `NullPointerException` in `CmdTreeImpl` ;) 
```suggestion


  @Test
  void shouldCheckParseCmdShellFormTree() {
    CmdTree tree = DockerTestUtils.parse("CMD executable param1 param2", DockerLexicalGrammar.CMD);
    assertThat(tree.getKind()).isEqualTo(DockerTree.Kind.CMD);
    assertThat(tree.keyword().value()).isEqualTo("CMD");

    assertThat(tree.cmdArguments().stream().map(TextTree::value)).containsExactly("executable", "param1", "param2");
    assertThat(((SyntaxToken)tree.children().get(0)).value()).isEqualTo("CMD");

    ExecFormTree execForm = (ExecFormTree) tree.children().get(1);
    assertThat(execForm).isSameAs(tree.execForm());
  }
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CmdTreeImpl.java
```diff
@@ -27,15 +27,18 @@
...
 
   private final ExecFormTree execForm;
+  private final ShellFormTree shellForm;
 
-  public CmdTreeImpl(SyntaxToken keyword, @Nullable ExecFormTree execForm) {
+  public CmdTreeImpl(SyntaxToken keyword, @Nullable ExecFormTree execForm, @Nullable ShellFormTree shellForm) {
     super(keyword);
     this.execForm = execForm;
+    this.shellForm = shellForm;
   }
```
rudy-regazzoni-sonarsource, line 42: You forgot to take care of this new attribute `shellForm` in the method `cmdArguments` ;)
Also as mentioned in my other comment in `CmdTreeImplTest.java`, you don't have a test case about shellForm, so you didn't saw this miss + the resulting NullPointerExeption.

rudy-regazzoni-sonarsource (changes requested):
Nice work, but you currently have an issue in your code, please see my comments.
I like the fact that the method `cmdArguments` will give the arguments irrespective of the form (shell or exec), so that we don't need to handle this on the rule side.
You can ping me back when the issue is fixed and I will do a second review promptly ;)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CmdTreeImpl.java
mstachniuk: Excellent catch!
Yes, I forgot to extend `cmdArguments()`. By the way, I also extended `CmdTree` to care `shellForm`.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CmdTreeImplTest.java
mstachniuk: Thank you for the test suggestion

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CmdTreeImpl.java
```diff
@@ -60,17 +61,32 @@ public Kind getKind() {
...
+        for (ExecFormLiteralTree element : literals.elements()) {
+          result.add(element.value());
+        }
+      }
+    }
+    if (shellForm != null) {
+      List<SyntaxToken> literals = shellForm.literals();
+      if (literals != null) {
+        result.addAll(literals);
       }
```
rudy-regazzoni-sonarsource, line 90: I would tend to skip the second check for null `literals != null` as the method always return an arrays and don't has a Nullable/CheckForNull annotation.
Same for execForm, WDYT ?

rudy-regazzoni-sonarsource (approved):
LGTM ! See my comment.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CmdTreeImpl.java
mstachniuk: ok

PR #442, opened 2022-11-22
SONARIAC-484 Support port range for instruction EXPOSE

rudy-regazzoni-sonarsource:
I choose the simple approach of adding another field to handle port + replacing the existing one : `port` -> `portMin` and `portMax`.
In case there is only one port provided -> `portMin == portMax`
Another solution would be to have this in another class `PortRangeTreeImpl`, open to challenge !
I took also the opportunity to fix the related grammar.

### DIFF
--- a/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/PortTreeImpl.java
+++ b/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/PortTreeImpl.java
@@ -21,6 +21,7 @@ package org.sonar.iac.docker.tree.impl;
 
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
 import org.sonar.iac.common.api.tree.Tree;
 import org.sonar.iac.docker.tree.api.PortTree;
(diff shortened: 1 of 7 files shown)


iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/ExposeTreeImplTest.java
```diff
@@ -57,8 +58,8 @@ void exposeInstructionWithSimpleValue() {
```
mstachniuk, line 58: Maybe it will be good to add some `notMatches()` to method `matchingSimple()` (GitHub doesn't allow me to put a comment in line 48).

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/PortTreeImpl.java
```diff
@@ -56,13 +65,13 @@ public Kind getKind() {
...
   }
 
   @Override
-  public SyntaxToken separator() {
-    return separator;
+  public SyntaxToken portMax() {
+    return portMax;
   }
 
   @Override
```
mstachniuk, line 77: Please add `@CheckForNull`

mstachniuk (approved):
LGTM, but please see my little comments

PR #443, opened 2022-11-22
SONARIAC-466 Docker: parse ENTRYPOINT instruction

mstachniuk:


rudy-regazzoni-sonarsource (approved):
LGTM ! No particular comment, everything seems fine to me.
More globally, since `CMD` and `ENTRYPOINT` have the exact same syntax, their respective classes have the same code, and of course we get the same/similar test code. I don't think it's a big issue as their code are quite straightforward.
BTW, I noticed in `CmdTreeImpl` an useless override of the keyword method ;)

mstachniuk:
Ok, I will fix `keyword` method in my next PR. Thanks for review

PR #444, opened 2022-11-22
SONARIAC-465 Docker: parse COPY instruction

rudy-regazzoni-sonarsource:
`COPY` has the exact same syntax as for `ADD`, so instead of copying everything, I created a shared abstract class `CommandSrcsDestOptionInstructionTreeImpl` which contain most of the code from which both `AddTree` and `CopyTree` extends from.

### DIFF
--- a/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddTree.java
+++ b/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddTree.java
@@ -19,10 +19,5 @@
  */
 package org.sonar.iac.docker.tree.api;
 
-import java.util.List;
-
-public interface AddTree extends InstructionTree {
-  List<ParamTree> options();
-  List<SyntaxToken> srcs();
-  SyntaxToken dest();
+public interface AddTree extends CommandSrcsDestOptionInstructionTree {
 }
--- a/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddTree.java
+++ b/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddTree.java
@@ -19,5 +19,5 @@
  */
 package org.sonar.iac.docker.tree.api;
 
-public interface AddTree extends CommandSrcsDestOptionInstructionTree {
+public interface AddTree extends TransferInstructionTree {
 }
--- a/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/CommandSrcsDestOptionInstructionTree.java
+++ b/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/CommandSrcsDestOptionInstructionTree.java
@@ -17,22 +17,12 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-package org.sonar.iac.docker.tree.impl;
+package org.sonar.iac.docker.tree.api;
 
 import java.util.List;
-import org.sonar.iac.docker.tree.api.AddTree;
-import org.sonar.iac.docker.tree.api.LiteralListTree;
-import org.sonar.iac.docker.tree.api.ParamTree;
-import org.sonar.iac.docker.tree.api.SyntaxToken;
 
-public class AddTreeImpl extends CommandSrcsDestOptionInstructionTreeImpl implements AddTree {
-
-  public AddTreeImpl(SyntaxToken add, List<ParamTree> options, LiteralListTree srcsAndDest) {
-    super(add, options, srcsAndDest);
-  }
-
-  @Override
-  public Kind getKind() {
-    return Kind.ADD;
-  }
+public interface CommandSrcsDestOptionInstructionTree extends InstructionTree {
+  List<ParamTree> options();
+  List<SyntaxToken> srcs();
+  SyntaxToken dest();
 }
--- a/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CommandSrcsDestOptionInstructionTreeImpl.java
+++ b/iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CommandSrcsDestOptionInstructionTreeImpl.java
@@ -1,0 +1,71 @@
+/*
+ * SonarQube IaC Plugin
+ * Copyright (C) 2021-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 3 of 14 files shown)


iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/CommandSrcsDestOptionInstructionTree.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.tree.api;
+
+import java.util.List;
+
+public interface CommandSrcsDestOptionInstructionTree extends InstructionTree {
```
mstachniuk, line 24: Maybe `SourceDestinationOptionTree`? I know that finding a good name is hard

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CommandSrcsDestOptionInstructionTreeImpl.java
```diff
@@ -0,0 +1,71 @@
...
+import org.sonar.iac.docker.tree.api.CommandSrcsDestOptionInstructionTree;
+import org.sonar.iac.docker.tree.api.LiteralListTree;
+import org.sonar.iac.docker.tree.api.ParamTree;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+
+/**
+ * To be used when we want to implement a command that expect one+ src with one dest (supporting both SHELL and EXEC format) with Params.
+ * Examples :
+ * MY_CMD --param=value src1 src2 dest
+ * MY_CMD --param=value ["src1", "src2", "dest"]
```
mstachniuk, line 34:
```suggestion
<p>
 * {@code MY_CMD --param=value src1 src2 dest}
 * {@code MY_CMD --param=value ["src1", "src2", "dest"]}
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddTree.java
```diff
@@ -19,10 +19,5 @@
  */
 package org.sonar.iac.docker.tree.api;
 
-import java.util.List;
-
-public interface AddTree extends InstructionTree {
-  List<ParamTree> options();
-  List<SyntaxToken> srcs();
-  SyntaxToken dest();
+public interface AddTree extends CommandSrcsDestOptionInstructionTree {
```
nils-werner-sonarsource, line 22: :exploding_head: I had to read this class name 10 times :smile: 
I would choose something like `TransferInstructionTree` because both instruction transfer elements from the local to the Docker file system.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CopyTreeImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    Assertions.assertThat(DockerLexicalGrammar.COPY)
+      .matches("COPY src dest")
+      .matches("copy src dest")
+      .matches("COPY dest")
+      .matches("COPY src1 src2 src3 dest")
+      .matches("COPY --option src dest")
+      .matches("COPY --option=value src dest")
+      .matches("COPY --chown=55:mygroup files* /somedir/")
+      .matches("COPY --link /foo /bar")
+      .matches("COPY \"src\" \"dest\"")
```
mstachniuk, line 45: nitpick `  COPY src dest`, `copy src dest`, `COPY . /opt/foo`, `COPY . .`

mstachniuk (approved):
LGTM, see my small comments

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CopyTreeImplTest.java
nils-werner-sonarsource, line 45: What about `COPY ["src1", "src2", "dest"]`?

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CopyTreeImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.sonar.iac.common.testing.TextRangeAssert.assertTextRange;
+import static org.sonar.iac.docker.tree.impl.DockerTestUtils.parse;
+
+class CopyTreeImplTest {
+  @Test
+  void matchingSimple() {
+    Assertions.assertThat(DockerLexicalGrammar.COPY)
+      .matches("COPY src dest")
+      .matches("copy src dest")
+      .matches("COPY dest")
```
nils-werner-sonarsource, line 39: What does this mean for the Docker image? src == dest?

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CopyTreeImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+    assertThat(tree.srcs()).hasSize(2);
+    assertThat(tree.srcs().get(0).value()).isEqualTo("src1");
+    assertThat(tree.srcs().get(1).value()).isEqualTo("src2");
+    assertThat(tree.dest().value()).isEqualTo("dest");
+  }
+
+  @Test
+  void copyInstructionExecForm() {
+    CopyTree tree = parse("COPY [\"src1\", \"src2\", \"dest\"]", DockerLexicalGrammar.COPY);
+    assertThat(tree.getKind()).isEqualTo(DockerTree.Kind.COPY);
```
nils-werner-sonarsource, line 70: If the kind is not expected to be different, you don't have to repeat the assertion.

nils-werner-sonarsource (approved):
LGTM! Please see my comments.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddTree.java
rudy-regazzoni-sonarsource: I'm especially bad at finding names ^^' After spending several minutes I surrender and fall back on some poor choice I find.
I will take the `TransferInstructionTree` 👍

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CopyTreeImplTest.java
rudy-regazzoni-sonarsource: For Docker image this will not "compile" and will throw an error.
I discussed with Pierre-Yves about how to parse this, the way that SSLR is implemented prevent from handling such use case : `COPY (word)+ word`
It won't work because at first it will match everything as **src** `(word)+` but then there is nothing left for **dest** `word`, SSLR won't backtrack/try to match **src** differently.

I choose to accept such use case, but I guess we could also throw a parse error or something similar, open to challenge.

nils-werner-sonarsource: Okay fine.

PR #449, opened 2022-11-22
SONARIAC-447 Docker: parse HEALTHCHECK instruction

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/HealthCheckTreeImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    HealthCheckTree tree = DockerTestUtils.parse("HEALTHCHECK NONE", DockerLexicalGrammar.HEALTHCHECK);
+    assertThat(tree.getKind()).isEqualTo(DockerTree.Kind.HEALTHCHECK);
+    assertThat(tree.keyword().value()).isEqualTo("HEALTHCHECK");
+
+    assertThat(tree.isNone()).isTrue();
+    assertThat(tree.cmd()).isNull();
+    assertThat(tree.options()).isNull();
+
+    assertThat(((SyntaxToken)tree.children().get(0)).value()).isEqualTo("HEALTHCHECK");
+    assertThat(((SyntaxToken)tree.children().get(1)).value()).isEqualTo("NONE");
```
nils-werner-sonarsource, line 66: Please add assertTextRange to some or every test to show that all children are represented in the textRange

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -345,6 +347,22 @@ public EntrypointTree ENTRYPOINT() {
...
+    return b.<HealthCheckTree>nonterminal(DockerLexicalGrammar.HEALTHCHECK).is(
+      b.firstOf(
+        f.healthcheck(
+          b.token(DockerKeyword.HEALTHCHECK),
+          b.token(DockerKeyword.NONE)
+        ),
+        f.healthcheck(
+          b.token(DockerKeyword.HEALTHCHECK),
+          b.zeroOrMore(PARAM()),
+          CMD()
```
nils-werner-sonarsource, line 360: You could also consider NONE as an instruction, like CMD. Then the second parameter would simply be an instruction tree. You would make the grammar narrower and do without the second constructor. Also, you don't have the problem of redundant fields `none` and `cmd`. The only disadvantage would be, you must create an instruction `NONE`, which may appear only in HEALTHCHECK. IMO this solution would be less hacky.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/HealthCheckTreeImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import org.sonar.iac.docker.tree.api.CmdTree;
+import org.sonar.iac.docker.tree.api.HealthCheckTree;
+import org.sonar.iac.docker.tree.api.ParamTree;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+
+public class HealthCheckTreeImpl extends InstructionTreeImpl implements HealthCheckTree {
+
+  private final SyntaxToken none;
+  private final List<ParamTree> options;
+  private final CmdTree cmd;
```
nils-werner-sonarsource, line 35: IMO having two contrary fields is not neatly solved and should be avoided.

nils-werner-sonarsource (approved):
Please see my comment about the grammar. IMO there is something you could improve to avoid contrary fields in the tree class implementation.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
rudy-regazzoni-sonarsource: Very interesting approach ! I will give it a try, sounds good to me 👍

rudy-regazzoni-sonarsource: Using this approach, it is possible to parse such instruction : `HEALTHCHECK --interval=30s NONE`
I was thinking it was invalid for docker, but apparently it is not ! Two birds with one stone 😶‍

PR #453, opened 2022-11-23
SONARIAC-493 Rule S6476: Instructions should be upper case

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Nice catch for the FROM/AS, I checked other keyword and didn't find any similar case.

PR #454, opened 2022-11-24
SONARIAC-494 Allow comments in multi line instructions

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/testing/IacTestUtils.java
```diff
@@ -0,0 +1,33 @@
...
+import org.sonar.api.internal.apachecommons.lang.StringUtils;
+
+public class IacTestUtils {
+
+  private IacTestUtils() {
+    // utils class
+  }
+
+  public static String code(String... lines) {
+    return StringUtils.join(lines, '\n');
```
rudy-regazzoni-sonarsource, line 31: I would recommend using `System.getProperty("line.separator");`, in order to exploit fully the windows build in our CI and ensure that our parser work with windows.

I'm mentioning this because it seems it does not :D Since recently, probably after I pulled the merged master with the support multiline branch, the test `InstructionFormatCheckTest` is failing on my computer (parse error line 10). I can make it work if I replace `CRLF` by `LF` in `InstructionFormatCheck/Dockerfile`, but OFC it's a patch and we need to make it work for windows.

rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comment about supporting windows EOL, not directly related to this ticket so we may create a ticket and fix this in another PR.

PR #455, opened 2022-11-24
SONARIAC-495 Analyze also Dockerfiles which has a suffix

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Didn't know that Dockerfile with extension was a thing, I didn't find much after some quick search. I see we have some in the imported projects, I was thinking there were like backup of some previous version.

PR #458, opened 2022-11-24
SONARIAC-455 RUN: Add support for options

mstachniuk:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -134,7 +137,7 @@ public ParamTree PARAM() {
         b.token(DockerLexicalGrammar.PARAM_PREFIX),
         b.token(DockerLexicalGrammar.PARAM_NAME),
         b.token(DockerLexicalGrammar.EQUALS_OPERATOR),
-        b.token(DockerLexicalGrammar.PARAM_VALUE)
+        b.optional(b.token(DockerLexicalGrammar.PARAM_VALUE))
```
mstachniuk, line 140: It's not a valid syntax, e.x. `COPY --option= src dest`, but in my opinion, it's better to have `--option=` as an `option` and not as `src` in `Tree`. Remember that files/directories may contain dash and equals character in name

iac-extensions/docker/src/test/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammarTest.java
```diff
@@ -83,8 +83,37 @@ void testKeyword() {
...
+      .notMatches("CMD& src dest")
+      .notMatches("CMD£ src dest")
+      .notMatches("CMD§ src dest")
+      .notMatches("CMD` src dest")
+      .notMatches("CMD~ src dest")
+      .notMatches("CMD* src dest")
+      .notMatches("CMD( src dest")
+      .notMatches("CMD) src dest")
+      .notMatches("CMD= src dest")
+      .notMatches("CMD{ src dest")
```
rudy-regazzoni-sonarsource, line 107: That's a lot of use cases XD I understand that you may want to test/ensure/prevent regression in the future, still that seems a lot, and from my perspective even with that big list of chars, there will still probably some other character.

I don't have a strong opinion on it, If I really wanted to have such restriction I probably would have written test differently, like a list of forbidden character and some loop / parameterized test.

rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comments.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammarTest.java
mstachniuk: The reason for that test was that I realized that the old implementation of  `DockerLexicalGrammar.keywords()` had a bug and `CMD` + special character, was parsed in the wrong way. 
I will improve this test in a separate PR to unblock other work on the project.

PR #460, opened 2022-11-24
SONARIAC-444 Create metric and highlight visitor

nils-werner-sonarsource:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/visitors/DockerMetricsVisitorTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  protected MetricsVisitor metricsVisitor(FileLinesContextFactory fileLinesContextFactory) {
+    return new DockerMetricsVisitor(fileLinesContextFactory, noSonarFilter);
+  }
+
+  @Test
+  void emptySource() {
+    scan("");
+    assertThat(visitor.linesOfCode()).isEmpty();
```
mstachniuk, line 51: Nitpick. I don't like personally not self-explanatory tests. Here the `scan()` is called and then are some assertions on `visitor` and it's not clear from where it comes. I know it's implemented in `AbstractMetricsTest`, but the flow is unclear if I'm looking at this test.
I would suggest that `scan()` may return `visitor` (or anything that is needed to assert). Then it will be clear from where it comes.

mstachniuk (approved):
LGTM, please see my little suggestions

iac-common/src/main/java/org/sonar/iac/common/extension/visitors/MetricsVisitor.java
```diff
@@ -46,7 +48,17 @@ protected MetricsVisitor(FileLinesContextFactory fileLinesContextFactory, NoSona
     languageSpecificMetrics();
   }
 
-  protected abstract void languageSpecificMetrics();
+  protected void languageSpecificMetrics() {
+    register(IacToken.class, (ctx, token) -> {
+      if (!(token.value().trim().isEmpty())) {
```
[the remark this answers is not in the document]
nils-werner-sonarsource: I agree, StringUtils would look smarter. However, we try to avoid using apache commons-lang. So far our parsers should not create and store in the AST a token with whitespace or even empty. However, this common class should catch the cases if one of the used parsers is creating such a token. I don't get the clue with the [Whitespace](https://en.wikipedia.org/wiki/Whitespace_(programming_language)).

iac-extensions/docker/src/test/java/org/sonar/iac/docker/visitors/DockerMetricsVisitorTest.java
nils-werner-sonarsource: I will adapt this. :+1:

PR #463, opened 2022-11-30
SONARIAC-440 [CF, TF] FP on rule S6304: "Granting access to all resources" should not be raised on actions without resource-level permissions

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/checks/ResourceAccessPolicyVector.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final String VECTOR_FILE = "ResourceAccessPolicyVector.json";
+
+  private ResourceAccessPolicyVector() {
+  }
+
+  private static final List<String> RESOURCE_ACCESS_POLICIES = loadResourceAccessPolicies();
+
+  private static List<String> loadResourceAccessPolicies() {
+    try {
+      String resourceAccessPolicies = loadJsonFile(VECTOR_FILE);
```
mstachniuk, line 47: Why not just call 
```
IOUtils.toString(VECTOR_FILE, StandardCharsets.UTF_8);
```
?

iac-common/src/test/java/org/sonar/iac/common/checks/CommonTestUtils.java
```diff
@@ -0,0 +1,118 @@
...
+  public static class TestTree extends AbstractTestTree {
+
+    public static Tree tree() {
+      return new TestTree();
+    }
+  }
+
+  public static class TestTextTree extends AbstractTestTree implements TextTree {
+
+    public static TextTree text(String value) {
```
mstachniuk, line 48: I would suggest renaming the method to `textTree` or similar.

iac-common/src/test/java/org/sonar/iac/common/checks/policy/PolicyValidatorTest.java
```diff
@@ -0,0 +1,107 @@
...
+      properties(attribute("Action", "sns:Publish"), attribute("Resource", "*"), attribute("Effect", "Allow")),
+      properties(attribute("Action", "sns:Publish"), attribute("NotResource", "*"), attribute("Effect", "Deny")),
+      properties(attribute("Action", "foo:bar"), attribute("Resource", "*"), attribute("Effect", "Allow")),
+      properties(attribute("Action", "sns:Publish"), attribute("Effect", "Allow")),
+      properties(attribute("Action", "sns:Publish"), attribute("Resource", "*")),
+      properties(attribute("Action", "sns:Publish"), attribute("NotResource", "*"))
+    );
+
+    Policy policy = new Policy(null, tree -> statements);
+    assertThat(PolicyValidator.findInsecureStatements(policy)).hasSize(2);
```
mstachniuk, line 87: I would suggest checking exactly what is in the result, not only the size.

iac-common/src/test/java/org/sonar/iac/common/checks/policy/PolicyValidatorTest.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  static class TestIterable extends AbstractTestTree implements Iterable<Tree> {
```
mstachniuk, line 90: Maybe it makes sense to move it to `CommonTestUtils`?

mstachniuk (approved):
LGTM, see my comments

iac-common/src/main/java/org/sonar/iac/common/checks/ResourceAccessPolicyVector.java
nils-werner-sonarsource: We don't want to relay on apache commons-io

PR #467, opened 2022-12-13
SONARIAC-497 Rule S5332: Using clear-text protocols is security-sensi…

andreaguarino:
…tive


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/UnencryptedProtocolCheck.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final Pattern UNENCRYPTED_PROTOCOLS = Pattern.compile("(http|ftp)://(?<rest>.+)", Pattern.CASE_INSENSITIVE);
+
+  private static final String LOOPBACK_IPV4 = "^127(?:\\.\\d+){0,2}\\.\\d+";
+  private static final String LOOPBACK_IPV6 = "^(?:0*:){0,7}?:?0*1";
+  private static final Pattern LOOPBACK = Pattern.compile("^localhost|" + LOOPBACK_IPV4 + "|" + LOOPBACK_IPV6, Pattern.CASE_INSENSITIVE);
+  private static final String MESSAGE = "Make sure that using clear-text protocols is safe here.";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(CommandInstructionTree.class, (ctx, commandInstruction) -> {
```
rudy-regazzoni-sonarsource, line 46: Interesting approach, but that also mean that this rule will apply on COPY instruction.
It's probably not a big deal as COPY is not supposed to support/have URL as argument.
I would still be stricter and replace it with directly concerned classes, to avoid uncontrolled/untested behavior and also to prevent futur bug/FP in case next version of Docker will have new instruction on which we would extend this class.

rudy-regazzoni-sonarsource (changes requested):
Great work overall ! Please see my comments, few little adjustments may be needed.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/UnencryptedProtocolCheck.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.sonar.iac.docker.tree.api.DockerTree.Kind.ENTRYPOINT;
+import static org.sonar.iac.docker.tree.api.DockerTree.Kind.RUN;
+
+@Rule(key = "S5332")
+public class UnencryptedProtocolCheck implements IacCheck {
+
+  private static final Pattern UNENCRYPTED_PROTOCOLS = Pattern.compile("(http|ftp)://(?<rest>.+)", Pattern.CASE_INSENSITIVE);
+
+  private static final String LOOPBACK_IPV4 = "^127(?:\\.\\d+){2}\\.\\d+";
+  private static final String LOOPBACK_IPV6 = "^(?:0*:){7}:?0*1|^::1|^1:1:1:1:1:1:1:1";
```
rudy-regazzoni-sonarsource, line 45:
```suggestion
  private static final String LOOPBACK_IPV6 = "^(?:0*:){7}:?0*1|^::1";
```
I checked, ::1 is the equivalent of 0:0:0:0:0:0:0:1 : the ping command resolve both of them to ::1
image

rudy-regazzoni-sonarsource (approved):
LGTM ! Please just fix my last comment ;)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/UnencryptedProtocolCheck.java
andreaguarino: Thanks a lot!

PR #468, opened 2022-12-13
SONARIAC-482 Rule S6470: Copying context directories recursively is security-sensitive

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/DirectoryCopySourceCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    SAFE, ROOT_OR_CURRENT, TOP_LEVEL_GLOBBING
+  }
+
+  /**
+   * Indicate if a path is sensitive. A path is sensitive if any of the current condition is true :
+   * - if the provided path resolve to a root level or to the current directory
+   * - if there is any top level entry (level <= 1) ending with a wildcard *
+   * Examples of root level : '/' ; '/test/..' ; 'c:/' ; 'c:/test/..'
+   * Examples of current level : '.' ; './test/..'
+   * Examples of top level entry ending with wildcard : 'a*' ; './a*' ; '/a*' ; './test/../a*'
```
andreaguarino, line 84: To make it prettier for JavaDoc we need to use tags:
```suggestion
  /**
   * Indicate if a path is sensitive. A path is sensitive if any of the current condition is true :
   * <ul>
   * <li>if the provided path resolve to a root level or to the current directory</li>
   * <li>if there is any top level entry (level <= 1) ending with a wildcard</li>
   * </ul>
   * Examples of root level : '/' ; '/test/..' ; 'c:/' ; 'c:/test/..'
   * Examples of current level : '.' ; './test/..'
   * Examples of top level entry ending with wildcard : 'a*' ; './a*' ; '/a*' ; './test/../a*'
   */
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/DirectoryCopySourceCheck.java
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+          // in case of relative path, if we go above the current folder, we suppose we lose the track and can't report any issue
+          return PathSensitivity.SAFE;
+        }
+      }
+    }
+
+    if (level == 0) return PathSensitivity.ROOT_OR_CURRENT;
+    if (level == 1 && hasTopLevelWithWildcard) return PathSensitivity.TOP_LEVEL_GLOBBING;
+    return PathSensitivity.SAFE;
```
andreaguarino, line 117: would it be possible to implement it using `Path`? I think some operations might be simplified, e.g. we might use `path.normalize()` to resolve the `..`

rudy-regazzoni-sonarsource: I tried, Path is system dependent.

Although it give really great result especially from Unix system (still it may give strange result with driver letter format such as C:/), on windows we get blocked quite quickly as * is not an authorized character. We get also different result, so it implies that we support any potential result.

It is apparently not possible to force the Unix Path format, the class used in that case come directly from installed FileSystemProviders, which are defined in the JDK installed on the machine running the code.

I found some shady solutions online about building some memory file system, but it seems a bit too much for our usage :D

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/DirectoryCopySourceCheck.java
```diff
@@ -0,0 +1,121 @@
...
```
andreaguarino, line 119: As discussed:
```suggestion
  private static PathSensitivity isSensitivePath(String path) {
    String[] levels = normalize(path);
    if (levels.length == 0) return PathSensitivity.ROOT_OR_CURRENT;
    if (levels.length == 1 && (isRootOrCurrent(levels[0]))) return PathSensitivity.ROOT_OR_CURRENT;
    if (Arrays.stream(levels).limit(2).anyMatch(l -> l.endsWith("*"))) return PathSensitivity.TOP_LEVEL_GLOBBING;
    return PathSensitivity.SAFE;
  }

  private static boolean isRootOrCurrent(String level) {
    return (level.isEmpty() || ".".equals(level) || WINDOWS_DRIVE_PATTERN.matcher(level).find());
  }


  private static String[] normalize(String path) {
    Deque<String> levels = new ArrayDeque<>();
    for (String current : path.split("/")) {
      if ("..".equals(current) && !levels.isEmpty()) {
        levels.removeLast();
      } else if (current.isEmpty()) {
        if (levels.isEmpty()) {
          levels.add(current);
        }
      } else if (".".equals(current) && levels.isEmpty()) {
        levels.add(current);
      } else if (!".".equals(current)) {
        levels.add(current);
      }
    }
    return levels.toArray(new String[] {});
  }
```

andreaguarino (approved):
LGTM, just a small refactoring on the `isSensitivePath` (and update of tests)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/DirectoryCopySourceCheck.java
rudy-regazzoni-sonarsource: Wonderful piece of code ! And as you saw, it even remove FP/FN of the previous code ^^'
Thanks !

andreaguarino: woohoo!

PR #469, opened 2022-12-14
SONARIAC-478 Rule S6471: Running containers as a privileged user is security-sensitive

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
```diff
@@ -0,0 +1,118 @@
...
+        String user = lastUser.get().user().value();
+        if(UNSAFE_USERS.contains(user)) {
+          ctx.reportIssue(lastUser.get(), String.format(MESSAGE_ROOT_USER, user));
+        }
+      }
+    });
+  }
+
+  private static Optional<UserTree> getLastUser(DockerImageTree dockerImage) {
+    return Optional.ofNullable(TreeUtils.getLastDescendant(dockerImage, tree -> ((DockerTree) tree).is(DockerTree.Kind.USER))).map(UserTree.class::cast);
```
andreaguarino, line 99: maybe we can have `TreeUtils.getLastDescendant` returning an `Optional`?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/TreeUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.annotation.CheckForNull;
+import org.sonar.iac.common.api.tree.Tree;
+import org.sonar.iac.docker.tree.api.DockerTree;
+
+public class TreeUtils {
+
+  private TreeUtils() {}
+
+  @CheckForNull
+  public static DockerTree getLastDescendant(Tree tree, Predicate<Tree> predicate) {
```
andreaguarino, line 32: what is the difference between a Tree and a DockerTree? should this method return a `Tree`?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
```diff
@@ -0,0 +1,118 @@
...
+    key = "safeImages",
+    description = "Comma separated list of safe images (no default root user).",
+    defaultValue = DEFAULT_SAFE_IMAGES)
+  public String safeImages = DEFAULT_SAFE_IMAGES;
+
+  private Set<String> safeImagesSet;
+  private Set<String> safeImages() {
+    if (safeImagesSet == null) {
+      safeImagesSet = Stream.of(safeImages.split(","))
+        .map(String::trim).collect(Collectors.toSet());
```
andreaguarino, line 62: when a user provides safe images, shouldn't we merge this provided list with the `DEFAULT_SAME_IMAGES`? 
I guess for the default ones, we're sure they aren't problematic regardless of the provided parameter.

andreaguarino (approved):
Overall LGTM, I left a few comments / questions

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
rudy-regazzoni-sonarsource: Sounds good to me, I wasn't sure if the provided list should be a default editable list or not, but indeed we are normally sure they are safe so the user should only be able to add their safe images.
Will update this.
I will also modify the code in case the user put an unsafe image in their personalized safe list to not report an error.

PR #476, opened 2023-01-04
SONARIAC-508 Rule S6364: Exclude Aurora for Cloudformation

rudy-regazzoni-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/checks/PropertyUtils.java
```diff
@@ -48,6 +49,13 @@ public static Trilean has(@Nullable Tree tree, String key) {
     return Trilean.FALSE;
   }
 
+  public static boolean hasKeyWithValue(@Nullable Tree tree, String key, Collection<String> values) {
```
nils-werner-sonarsource, line 52: IMO this method is too specific and the name is not self-explaining. Due to we already have a helper value, I could imagine a helper like `valueIs(@Nullable Tree tree, String key, Predicate<Tree>)` WDYT?

iac-extensions/cloudformation/src/main/java/org/sonar/iac/cloudformation/checks/ShortBackupRetentionCheck.java
```diff
@@ -41,7 +43,9 @@ public class ShortBackupRetentionCheck extends AbstractResourceCheck {
 
   @Override
   protected void checkResource(CheckContext ctx, Resource resource) {
-    if ((resource.isType("AWS::RDS::DBInstance") && PropertyUtils.isMissing(resource.properties(), "SourceDBInstanceIdentifier"))
+    if ((resource.isType("AWS::RDS::DBInstance")
+        && PropertyUtils.isMissing(resource.properties(), "SourceDBInstanceIdentifier")
+        && !PropertyUtils.hasKeyWithValue(resource.properties(), "Engine", ENGINES_EXCEPTION))
```
nils-werner-sonarsource, line 48: I would switch the conditions.

nils-werner-sonarsource (approved):
LGTM but see my comments about the helper method

iac-common/src/main/java/org/sonar/iac/common/checks/PropertyUtils.java
rudy-regazzoni-sonarsource: You are right, I will change this method to have a less specific one.

PR #479, opened 2023-01-04
SONARIAC-510 Usernames in the USER instruction can be uppercase

mstachniuk:
Actually, a username can be almost everything except colon, whitespace, and tabulation.
There are only recommendations about usernames. Docker accepts everything except:
```
# USER $user:${group with space}
# USER ${var with space}
```

Source: https://www.unix.com/man-page/linux/8/useradd/ (section CAVEATS).


nils-werner-sonarsource (approved):
LGTM!

PR #480, opened 2023-01-05
SONARIAC-504 Newlines without prior spaces provoke parsing errors

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
This does not fix the SONARIAC-505 issue : this fix is working for element using STRING_LITERAL from DockerLexicalGrammar, which is not the case for example for ENV, which is using KeyValuePair format.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -25,7 +25,7 @@ public class DockerLexicalConstant {
 
   public static final String COMMENT = "(?:" + LexicalConstant.SINGLE_LINE_COMMENT_HASH + ")";
   public static final String STRING_LITERAL_WITH_QUOTES = "\"(?:[^\"\\\\]*+(?:\\\\[\\s\\S])?+)*+\"";
-  public static final String STRING_LITERAL_WITHOUT_QUOTES = "[^\\s]+";
+  public static final String STRING_LITERAL_WITHOUT_QUOTES = "[^\\s]+(?<!\\\\)";
```
rudy-regazzoni-sonarsource, line 28: In case we don't want to use the lookahead style for some reason, we could also setup this : `(?:\\\\*[^\\s\\\\]+)+`

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/MaintainerTreeImplTest.java
```diff
@@ -46,7 +46,9 @@ void matchingSimple() {
       .matches("MAINTAINER bob \\\r\n boberman")
       .matches("MAINTAINER \"bob boberman bob@bob.com\"")
       .matches("MAINTAINER bob /  boberman")
-      .matches("MAINTAINER bob \\ boberman")
+      // TODO : this is a valid syntax that we should support
+      .notMatches("MAINTAINER bob \\ boberman")
+
```
rudy-regazzoni-sonarsource, line 51: This solution has this counterpart : we do not support such instruction anymore, where we have a `\` with spaces before and after, in the middle of a `StringLiteral`.

PR #481, opened 2023-01-05
SONARIAC-501 S6471 raises multiple issues for a multi-stage build

mstachniuk:
Sorry for creating so many Dockerfiles in tests, but it was necessary by this ticket, since only the last `FROM` instruction in Dockerfile may be considered insecure. I tried to do this code inline in `PrivilegedUserCheckTest`, but this needs big changes in test-commons -> to implement Verifiers based on String instead of a file.


iac-extensions/docker/src/test/resources/checks/PrivilegedUserCheck/Dockerfile_multi_stage_build
```diff
@@ -0,0 +1,5 @@
+FROM gcc as build
+
+# Noncompliant@+1 {{Scratch images run as root by default. Make sure it is safe here.}}
+  FROM scratch
+# ^^^^^^^^^^^^
```
rudy-regazzoni-sonarsource, line 5: Could you add another test file where the first `FROM` is sensitive and the second is not ? I don't think I saw this use case.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/checks/PrivilegedUserCheckTest.java
```diff
@@ -19,20 +19,72 @@
...
+    CheckContext ctx = Mockito.mock(CheckContext.class);
+    SyntaxToken imageName = new SyntaxTokenImpl("scratch", TextRanges.range(1,5, "scratch"), List.of());
+    SyntaxToken fromToken = new SyntaxTokenImpl("FROM", TextRanges.range(1,0, "FROM"), List.of());
+    FromTree from = new FromTreeImpl(fromToken, null, new ImageTreeImpl(imageName, null, null), null);
+    DockerImageTree tree = new DockerImageTreeImpl(from, List.of());
+
+    check.handle(ctx, tree);
+
+    Mockito.verify(ctx).reportIssue(from, "Scratch images run as root by default. Make sure it is safe here.");
   }
```
rudy-regazzoni-sonarsource, line 89: Is this a plausible use case ? I find it strange that a `DockerImage` can have no parent.
The fact that you had to use a Mockito to simulate that tend to prove that we normally can't encounter such use case ^^'
Unless we have good reason for this, I would tend to remove this test + remove the condition `if (parent instanceof FileTree) {` in `isLastDockerImageInFile` and cast directly.

rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comments + fix the code smell.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/checks/PrivilegedUserCheckTest.java
mstachniuk: Good point. `DockerImage` should always have a parent and it will be always `FileTree`. Unfortunately, Java type system (or our design) allows any `DockerTree` to be a parent. My intention when I wrote `if (parent instanceof FileTree)` was to not fail with ugly `ClassCastException`. However, the structure of the trees is in our control, so such exception means an issue in the parsing phase. 
So now, I'm ok with remove of the `if` and artificial test just for coverage.

PR #485, opened 2023-01-09
SONARIAC-511 [S6472] Wordlist check only matches on uppercase

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
I kept the logic with regex to test elements.
I replaced List by Set as they are only used with contains method.

nils-werner-sonarsource (approved):
LGTM!

PR #486, opened 2023-01-09
SONARIAC-512 Empty ENV variables with an equal sign cause parsing errors

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #497, opened 2023-01-30
SONARIAC-537 Remove Tree keyword from existing Tree elements and add …

mstachniuk:
…Instruction keyword for relevant element


iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/AddInstruction.java
```diff
@@ -19,5 +19,5 @@
  */
 package org.sonar.iac.docker.tree.api;
 
-public interface HereDocumentTree extends LiteralListTree {
+public interface AddInstruction extends TransferInstruction {
```
rudy-regazzoni-sonarsource, line 22: To double check, I guess it's a diff issue from Github

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/NoneInstruction.java
```diff
@@ -19,5 +19,5 @@
  */
 package org.sonar.iac.docker.tree.api;
 
-public interface ShellFormTree extends LiteralListTree {
+public interface NoneInstruction extends Instruction {
```
rudy-regazzoni-sonarsource, line 22: Now that I see it, not sure if None should be an Instruction.
Anyway, not in the scope of this ticket, I will note it to check this later.

rudy-regazzoni-sonarsource (approved):
LGTM ! I just added a general comment not related to this change directly and not in the scope of this ticket, you can merge whenever you want ;)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/NoneInstruction.java
mstachniuk: I think that git wrongly detects the move of the file.

PR #499, opened 2023-02-01
SONARIAC-538 Modify top syntax elements File / Body

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/AbstractDockerTreeImpl.java
```diff
@@ -42,9 +44,14 @@ public final boolean is(Kind... kind) {
...
     if (textRange == null) {
-      List<TextRange> childRanges = children().stream().map(HasTextRange::textRange).collect(Collectors.toList());
+      List<TextRange> childRanges = children().stream().map(HasTextRange::textRange)
+        .filter(Objects::nonNull).collect(Collectors.toList());
+      if (childRanges.isEmpty()) {
+        return null;
+      }
       textRange = TextRanges.merge(childRanges);
     }
     return textRange;
```
rudy-regazzoni-sonarsource, line 57: For empty files, a Body will have two empty list as Arg instructions and DockerImages, so it mean it will literaly have no children, which mean we need to handle such use case for textRange : possibility to return null, and filter out null range when merging.

rudy-regazzoni-sonarsource:
If we don't plan to have Directive element, having a Body element may make much less sense.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/FileImplTest.java
```diff
@@ -37,73 +36,73 @@ void shouldParseEmptyFile() {
     File file = parseFile("");
     assertThat(file.getKind()).isEqualTo(DockerTree.Kind.FILE);
     assertTextRange(file.textRange()).hasRange(1,0,1,0);
```
rudy-regazzoni-sonarsource, line 38: If we plan to keep this Body element, I think we should move most of those test into a new `BodyImplTest` element and only have test related to the `File` element itself here.

rudy-regazzoni-sonarsource:
I didn't use Optional for Body, I'm considering that an empty dockerfile is invalid as it seems to be the case.
If we want a more conservative approach and not throw a parse error, we can go the Optional way.
WDYT ?

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/FileImplTest.java
```diff
@@ -19,113 +19,56 @@
...
   }
 
   @Test
-  void shouldParseEmptyFileWithByteOrderMark() {
-    File file = parseFile("\uFEFF");
-    assertThat(file.getKind()).isEqualTo(DockerTree.Kind.FILE);
-    assertTextRange(file.textRange()).hasRange(1,0,1,0);
-    assertThat(file.globalArgs()).isEmpty();
-    assertThat(file.dockerImages()).isEmpty();
-  }
```
nils-werner-sonarsource, line 51: Please add another test case with a valid body and a byte order mark. So we can test if the parser works correctly.

nils-werner-sonarsource (approved):
LGTM! Please add a test case which checks the byte order mark

PR #501, opened 2023-02-01
SONARIAC-539 Create Argument syntax element, parse quoted string

mstachniuk:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/QuotedString.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.tree.api;
+
+public interface QuotedString extends DockerTree {
```
nils-werner-sonarsource, line 22: How to access the value of the String?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/StringNoSpacing.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.tree.api;
+
+public interface StringNoSpacing extends DockerTree {
```
nils-werner-sonarsource, line 22: Instead of a class per kind of Strings, you could implement the StringLiteral interface and describe string types/kinds. In the end, we are only interested in the value. In addition, there should not be a difference between StringNoSpacing and StringWithSpace. It is only a grammar specification but not relevant to the AST. Creating so many different types of trees will make it hard to handle them.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/DoubleQuotedString.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.tree.api;
+
+public interface DoubleQuotedString extends DockerTree {
```
mstachniuk, line 22: Do we need some base interface for all `Argument` related classes?
Any common methods come to your mind, ex: `text()` or `value()`?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalGrammar.java
```diff
@@ -151,6 +163,9 @@ private static void lexical(LexerlessGrammarBuilder b) {
 
     b.rule(EOF).is(b.token(GenericTokenType.EOF, b.endOfInput())).skip();
 
+    b.rule(QUOTED_STRING_REGEX).is(b.regexp("'[^']*+'"));
+    b.rule(STRING_NO_SPACING_REGEX).is(b.regexp("(?:\\$|\\\\\"|[^\"$])++"));
```
mstachniuk, line 167: The regex looks scary, but it works. I'm not sure why so many backslashes are needed before a double quote 🤔

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/Argument.java
```diff
@@ -0,0 +1,23 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.tree.api;
+
+public interface Argument extends DockerTree {
+}
```
nils-werner-sonarsource, line 23: What about a method to access the elements of the argument?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/DoubleQuotedStringImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import org.sonar.iac.docker.tree.api.DoubleQuotedString;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+
+public class DoubleQuotedStringImpl extends AbstractDockerTreeImpl implements DoubleQuotedString {
+  private final SyntaxToken leftQuote;
+  @Nullable
+  private final List<DockerTree> words;
+  private final SyntaxToken rightQuote;
+
+  public DoubleQuotedStringImpl(SyntaxToken leftQuote, @Nullable List<DockerTree> words, SyntaxToken rightQuote) {
```
nils-werner-sonarsource, line 36: This shouldn't be nullable. It should be an empty list when there is not a word. In the factory, you can perform `or(Collection.emptyList)` on the Optional provided by `zeroOrMore` builder method.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -276,6 +281,61 @@ public UserInstruction USER() {
     );
   }
 
+  public Argument ARGUMENT() {
+    return b.<Argument>nonterminal(DockerLexicalGrammar.ARGUMENT).is(
+      f.argument2(
```
nils-werner-sonarsource, line 286: What about argument method number 1? Do we still need it? Do we plan to remove it?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/DockerTree.java
```diff
@@ -38,6 +38,11 @@ enum Kind implements GrammarRuleKey {
     IMAGE(Image.class),
     PARAM(Param.class),
     ALIAS(Alias.class),
+    ARGUMENT(Argument.class),
+    QUOTED_STRING(QuotedString.class),
+    DOUBLE_QUOTED_STRING(DoubleQuotedString.class),
+    STRING_WITH_SPACING(StringWithSpacing.class),
+    STRING_NO_SPACING(StringNoSpacing.class),
```
nils-werner-sonarsource, line 45: Should be one class. There is no need to differentiate in the AST.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/QuotedStringImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+import org.sonar.iac.common.api.tree.Tree;
+import org.sonar.iac.docker.tree.api.QuotedString;
+import org.sonar.iac.docker.tree.api.SyntaxToken;
+
+public class QuotedStringImpl extends AbstractDockerTreeImpl implements QuotedString {
+
+  private final SyntaxToken token;
+
+  public QuotedStringImpl(SyntaxToken token) {
```
nils-werner-sonarsource, line 31: Instead of a pure token, you should expect a StringLiteral-like tree. This will make it easier to handle this value in combination with the result of the StringNoSpacing and StringWithSpacing grammar rule.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/DoubleQuotedStringImpl.java
```diff
@@ -0,0 +1,57 @@
...
+  public List<Tree> children() {
+    List<Tree> children = new ArrayList<>();
+    children.add(leftQuote);
+    if (words != null) {
+      children.addAll(words);
+    }
+    children.add(rightQuote);
+    return children;
+  }
+
```
nils-werner-sonarsource, line 52: Without a propper getter method for the words. We have to use children and skip the first and last elements of the list. This is not handy.

nils-werner-sonarsource (changes requested):
Please merge the two String classes into a simple StringLiteral class. Do not expect tokens for more than punctuators, expect more extendable types. 

I know it is the first iteration of the Argument implementation but please provide simple, maybe common, getters in the interfaces and implementations. Do not use children to access necessary properties like the value of quoted strings.

nils-werner-sonarsource (changes requested):
Now you are violating the grammar described in the ticket. How should a double-quoted string become an argument?

PR #502, opened 2023-02-01
SONARIAC-532 Preprocess line escaping before parsing a Dockerfile - Handle custom escaping

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerPreprocessor.java
```diff
@@ -54,10 +58,21 @@ public String process(String source) {
...
       shiftedOffsetMap.put(m.end() - shiftedIndex, shiftedIndex);
     }
+
     sourceOffset = new SourceOffset(source, shiftedOffsetMap);
     return sb.toString();
   }
 
+  private static Matcher matchEscapedLineBreaks(String source) {
+    String escapeCharacter = determineEscapeCharacter(source);
+    String escapedLineBreakPattern = "(?<!escape=)" + escapeCharacter + "[" + WHITESPACE + "]*+" + EOL;
```
rudy-regazzoni-sonarsource, line 68: Will need to adapt this regex for spaces before/after =.
Also, I guess this would prevent removing the EOL for such instruction :
`LABEL escape=\\n"my value"`
Probably not very common, but maybe we can also include the # in the lookbehind to narrow the possibilities.

rudy-regazzoni-sonarsource (approved):
LGTM ! Very nice work, please just see my few comments.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerPreprocessor.java
nils-werner-sonarsource: IMO super edge case. Will not handle it for now. But nice catch.

PR #507, opened 2023-02-08
SONARIAC-542 Adapt ExecForm implementation

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Baby step to move Shell/Exec/Heredoc form to Argument.
Focus on ExecForm, had to introduce a method `tokens()` in Argument that may not be needed anymore once the move is complete.
Started the deprecation of  `literals()`, which will be removed completely at the end of the move.

nils-werner-sonarsource (changes requested):
You can remove some technical dept, by removing ExecFromLiteral. Reduce the complexity of the grammar I commented on.

Maybe the ExecForm and ExecFormUtils can be a bit more generalized. So that we can use a utils class in all checks which accept a command structure. So there is no difference in handling exec and shell. However, this would already be another baby step.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/ExecFormLiteralImpl.java
```diff
@@ -22,13 +22,13 @@
 import java.util.List;
 import org.sonar.iac.common.api.tree.Tree;
 import org.sonar.iac.docker.tree.api.ExecFormLiteral;
-import org.sonar.iac.docker.tree.api.SyntaxToken;
+import org.sonar.iac.docker.tree.api.ExpandableStringLiteral;
 
 public class ExecFormLiteralImpl extends AbstractDockerTreeImpl implements ExecFormLiteral {
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Damn that's a very good point.
I removed this useless layer, this simplify things a bit.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
```diff
@@ -260,22 +258,19 @@ public HereDocument hereDocument(SyntaxToken content) {
...
-        for (Tuple<SyntaxToken, SyntaxToken> comaAndLiteral : comaAndLiterals) {
-          separators.add(comaAndLiteral.first());
-          elements.add(new ExecFormLiteralImpl(comaAndLiteral.second()));
+    SeparatedList<ExpandableStringLiteral> separatedList = new SeparatedListImpl<>(elements, separators);
+    if (firstExpression.isPresent()) {
+      elements.add(firstExpression.get());
+      if (otherExpressions.isPresent()) {
+        for (Tuple<SyntaxToken, ExpandableStringLiteral> expressionWithComma : otherExpressions.get()) {
+          separators.add(expressionWithComma.first());
+          elements.add(expressionWithComma.second());
```
nils-werner-sonarsource, line 273: Much better!

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/ExecFormTestUtils.java
```diff
@@ -0,0 +1,56 @@
...
+import org.sonar.iac.docker.tree.api.Argument;
+import org.sonar.iac.docker.tree.api.DockerTree;
+import org.sonar.iac.docker.tree.api.ExpandableStringCharacters;
+import org.sonar.iac.docker.tree.api.ExpandableStringLiteral;
+import org.sonar.iac.docker.tree.api.Expression;
+import org.sonar.iac.docker.tree.api.Literal;
+
+public class ExecFormTestUtils {
+
+  public static String toString(Argument arg) {
```
nils-werner-sonarsource, line 32: 1. Please do not use the method name `toString` it is reserved for other purposes
2. You can simply use ArgumentUtils instead of duplicating its logic.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/ExecFormImpl.java
```diff
@@ -58,11 +59,21 @@ public SyntaxToken leftBracket() {
...
-    for (ExecFormLiteral element : literalsWithSeparators.elements()) {
-      result.add(element.value());
+    throw new UnsupportedOperationException("Method to be removed from LiteralList interface once SONARIAC-541 and SONARIAC-572 are done.");
+  }
+
+  @Override
+  public List<Argument> arguments() {
+    List<Argument> result = new ArrayList<>();
+    for (ExpandableStringLiteral element : expressionsWithSeparators.elements()) {
+      result.add(new ArgumentImpl(List.of(element)));
```
nils-werner-sonarsource, line 76: The tree class should not take care of this. Maybe we can let the grammar provide the arguments in the separate list directly? WDYT? (However, we can do it in another PR)

nils-werner-sonarsource (approved):
LGTM! Still a bit odd due to all the deprecation and replacements. We should get rid of the LiteralList as well (Next step) This can by simplified to List<Argument>.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/ExecFormImpl.java
rudy-regazzoni-sonarsource: I agree on this, I put this in place in my last commit.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
```diff
@@ -262,22 +260,23 @@ public HereDocument hereDocument(SyntaxToken content) {
     return new HereDocumentImpl(content);
   }
 
-  public ExecForm execForm(SyntaxToken leftBracket,
-    Optional<Tuple<SyntaxToken, Optional<List<Tuple<SyntaxToken, SyntaxToken>>>>> literals,
+  public Argument expressionAsArgument(Expression expression) {
```
nils-werner-sonarsource, line 263:
```suggestion
  public Argument singleExpressionArguement(Expression expression) {
```

nils-werner-sonarsource (approved):
LGTM! Can you explain, why we loose so many issues on ITs?

rudy-regazzoni-sonarsource:
> LGTM! Can you explain, why we loose so many issues on ITs?

This is because we already have a rule using Exec/Shell form, working with the contract that the method `literals()` will provide expected elements (it use it directly and indirectly with `srcs()`/`dest()`).
This is still working for Shell/Heredoc form, but now `ExecForm` throw the `UnsupportedOperationException`, which throw error during rule execution and prevent further scanning / issue detection.

This is why at first I tried to keep the previous `literals()` method of `ExecForm` working temporarily, but at some point it was bringing too much hassle.

So unfortunately we can only expect to detect those issues back once we are done with the transition on ExecForm/ShellForm/HeredocForm + removed old method `literals()` and adapted the rule to `arguments()`.

nils-werner-sonarsource:
> So unfortunately we can only expect to detect those issues back once we are done with the transition on ExecForm/ShellForm/HeredocForm + removed old method `literals()` and adapted the rule to `arguments()`.

Okay. Another approach would be to use `arguments()` inside the `literals()` declaration. So it would prove that this rule is working with the new data from `arguments()`. Instead of throwing an exception, you could map the arguments to the expected output of ` `literals()`

rudy-regazzoni-sonarsource (commented):
LGTM !

PR #509, opened 2023-02-08
SONARIAC-558 Create an Utils to evaluate Argument and replace variable with their values - Simple implementation without variables

nils-werner-sonarsource:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/utils/ArgumentUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+    "f\"o\"o",
+    "f'oo'",
+    "f'o'\"o\""
+  })
+  void resolveArgument(String input) {
+    Argument argument = parseArgument(input);
+    assertThat(ArgumentUtils.resolve(argument)).extracting(ArgumentUtils.ArgumentResolution::value)
+      .isNotNull()
+      .isEqualTo("foo");
+  }
```
rudy-regazzoni-sonarsource, line 46: We are missing the use case to show that variable are not replaced in single quote string : `'$foo'`

rudy-regazzoni-sonarsource (approved):
LGTM ! Really great PR, now we clearly see that there was a flaw in our previous implementation, we were indeed missing a level. Please see my comment.

rudy-regazzoni-sonarsource (approved):
LGTM ! Nice first baby step, please just see my comment.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/utils/ArgumentUtilsTest.java
nils-werner-sonarsource: Unfortunately, I'm not able to escape the single quotes correctly. The CsvSource annotation is cropping them all the time.

PR #510, opened 2023-02-08
SONARIAC-539 Create Argument syntax element

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !
I mixed a bit the PRs, here is my comments for this one :
Really great PR, now we clearly see that there was a flaw in our previous implementation, we were indeed missing a level.

PR #512, opened 2023-02-09
SONARIAC-574 Modify KeyValuePair to handle pairs with equal sign and arguments

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comment.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/NewKeyValuePairImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+      .matches("key=value=")
+      .matches("${key}=value=")
+      .matches("key=")
+
+      .matches("\"key=\"=value") // This is invalid syntax, but it would increase the grammar complexity to cover this case
+
+      .notMatches("key")
+      .notMatches("key= value")
+      .notMatches("\"key\"= \"value\"")
+    ;
```
rudy-regazzoni-sonarsource, line 52: I would add few more tests cases, even if they are not yet supported :
MATCHES
`ke\"y=value`
`key=val\"ue`

PR #513, opened 2023-02-09
SONARIAC-576 Unquoted string literals should allow escaped quotes

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #514, opened 2023-02-10
SONARIAC-547 Modify UserInstruction implementation to use Argument

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
```diff
@@ -234,12 +235,8 @@ public RunInstruction run(SyntaxToken token, Optional<List<Param>> options, Opti
...
   }
 
-  public UserInstruction user(SyntaxToken keyword, SyntaxToken user, Optional<Tuple<SyntaxToken, SyntaxToken>> colonAndGroup) {
-    if (colonAndGroup.isPresent()) {
-      return new UserInstructionImpl(keyword, user, colonAndGroup.get().first(), colonAndGroup.get().second());
-    } else {
-      return new UserInstructionImpl(keyword, user, null, null);
-    }
+  public UserInstruction user(SyntaxToken keyword, List<Tuple<SyntaxToken, Argument>> argumentsWithWhitespace) {
+    return new UserInstructionImpl(keyword, argumentsWithWhitespace.stream().map(Tuple::second).collect(Collectors.toList()));
```
nils-werner-sonarsource, line 239: It's identical to https://github.com/SonarSource/sonar-iac/pull/515/files#diff-b23fa74b6952185eb37724c814d82f308050d356545e4fd2fd09cbec78fbfe7eR155. Maybe you can define a method for this task which name describes the actual task.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -282,13 +282,10 @@ public UserInstruction USER() {
     return b.<UserInstruction>nonterminal(DockerLexicalGrammar.USER).is(
       f.user(
         b.token(DockerKeyword.USER),
-        b.token(DockerLexicalGrammar.USER_NAME),
-        b.optional(
-          f.tuple(
-            b.token(DockerLexicalGrammar.USER_SEPARATOR),
-            b.token(DockerLexicalGrammar.USER_GROUP)
-          )
-        )
```
nils-werner-sonarsource, line 291: Are these grammar rules still used? If not, let's drop them off to keep everything clean.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
```diff
@@ -96,7 +97,7 @@ private void handle(CheckContext ctx, DockerImage dockerImage) {
         ctx.reportIssue(dockerImage.from(), MESSAGE_OTHER_IMAGE);
       }
     } else {
-      String user = lastUser.get().user().value();
+      String user = ArgumentUtils.resolve(lastUser.get().arguments()).value();
```
nils-werner-sonarsource, line 100: Seeing this for the second time: WDYT about an interface like `HasArguments` which describes the method `arguments()`. So we can extend the ArgumentUtils to handle these classes by calling the `arguments()`. This way you would get rid of calling it several places.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/utils/ArgumentUtils.java
```diff
@@ -31,6 +32,17 @@ private ArgumentUtils() {
...
+  /**
+   * Resolve a list of Argument as a single string, used by some instructions like USER.
+   */
+  public static ArgumentResolution resolve(List<Argument> arguments) {
+    StringBuilder sb = new StringBuilder();
+    for (Argument argument : arguments) {
+      sb.append(resolve(argument).value);
+    }
+    return new ArgumentResolution(sb.toString());
+  }
```
nils-werner-sonarsource, line 44: I don't know if we should merge multiple solutions into a single one. It may make sense in this context, but it does not apply to other situations where it is necessary to handle argument by argument. You could rename the method to something like resolveAndMerge(). This represents what is happening here.

nils-werner-sonarsource (approved):
LGTM! Please see my comments. I see some potential to optimize the implementation.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/TreeFactory.java
rudy-regazzoni-sonarsource: Found an alternative : on the grammar I removed the tuple and replaced it with something similar to the `withOptionalWhitespace()`, which accept any first element (to mutualize whitespace and optional whitespace cases).
On the `TreeFactory` part it is simplier as we get directly a `List<Argument>`.

PR #515, opened 2023-02-10
SONARIAC-545 Modify WorkdirInstruction implementation to use Argument

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -223,7 +223,10 @@ public WorkdirInstruction WORKDIR() {
     return b.<WorkdirInstruction>nonterminal(DockerLexicalGrammar.WORKDIR).is(
       f.workdir(
         b.token(DockerKeyword.WORKDIR),
-        ARGUMENTS()
+        b.oneOrMore(f.tuple(
+          b.token(DockerLexicalGrammar.WHITESPACE),
+          ARGUMENT()
+        ))
```
nils-werner-sonarsource, line 229: I can imagen having an `ARGUMENTS()` rule in the end which creates this list of tuples. For now, it's fine like this.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/WorkdirInstructionImpl.java
```diff
@@ -22,23 +22,24 @@
...
 import org.sonar.iac.docker.tree.api.SyntaxToken;
 import org.sonar.iac.docker.tree.api.WorkdirInstruction;
 
 public class WorkdirInstructionImpl extends InstructionImpl implements WorkdirInstruction {
 
-  private final List<SyntaxToken> workdirList;
+  private final List<Argument> arguments;
 
-  public WorkdirInstructionImpl(SyntaxToken keyword, List<SyntaxToken> workdirList) {
+  public WorkdirInstructionImpl(SyntaxToken keyword, List<Argument> arguments) {
```
nils-werner-sonarsource, line 33: You could keep the variable name, to show what the arguments are actually representing in this context. But on the other hand, I could imagine an interface for such Instructions to provide the arguments() method. Could be useful for further processing. However, fine for now IMO.

nils-werner-sonarsource (approved):
LGTM! I've added some thoughts but no adjustments are needed.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
rudy-regazzoni-sonarsource: 👍 For the first/two element I did it like this, but I will definitely create an `ARGUMENTS()` to handle the next ones (also will need to rename to old crap `ARGUMENTS()` :D )

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/WorkdirInstructionImpl.java
rudy-regazzoni-sonarsource: 👍 good point, I was also thinking that later we will probably have common interface/abstract class parent as we will have many instruction with this similar grammar `keyword argument+`

PR #517, opened 2023-02-10
SONARIAC-575 Modify KeyValuePair to handle pairs without equal sign and arguments

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -549,6 +558,26 @@ public NewKeyValuePair NEW_KEY_VALUE_PAIR() {
...
+  public NewKeyValuePair KEY_VALUE_PAIR_WITHOUT_EQUAL() {
+    return b.<NewKeyValuePair>nonterminal().is(
+      f.newKeyValuePair(
+        KEY_ARGUMENT(),
+        b.optional(
+          f.tuple(
+            b.token(DockerLexicalGrammar.WHITESPACE),
+            ARGUMENT()
+          )
+        ),
```
rudy-regazzoni-sonarsource, line 570: Since this first `WHITESPACE` is ignored, maybe we could use the [ignoreFirst](https://github.com/SonarSource/sonar-iac/pull/514/commits/e4798390cf33e4ab3b41c05408c7e74b5c0a77ed#diff-873c87fe71cc1f3e453928e084bf604da85e0ed99c360db0bab8a6639c3c32d2R226-R229) method so that it doesn't even appear in the `TreeFactory` `newKeyValuePair()` method.
On the DockerGrammar side it would also explicitly tell that the first element is processed differently from the others.

rudy-regazzoni-sonarsource (approved):
LGTM ! New KeyValuePair implementation is great ! Nice idea for merging all expressions including whitespace in a single Argument. Please see my comment.

PR #520, opened 2023-02-13
SONARIAC-541 Modify ShellForm implementation

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -437,7 +436,7 @@ public ShellForm SHELL_FORM() {
     return b.<ShellForm>nonterminal(DockerLexicalGrammar.SHELL_FORM).is(
       f.shellForm(
         b.oneOrMore(
-          b.token(STRING_LITERAL)
+          f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), ARGUMENT_GENERIC())
```
rudy-regazzoni-sonarsource, line 439: I ignore every whitespace.
Maybe instead we should only ignore the first one, and then capture other whitespace and make them expression/argument, as it is done for KeyValuePair ?

rudy-regazzoni-sonarsource:
Splitted SHELL_FORM grammar into two versions : SHELL_FORM and SHELL_FORM_GENERIC.
This is duplicating several grammar elements, the normal version allow Docker syntax only for variable, while the generic one is more permissive and allow any modifier.

rudy-regazzoni-sonarsource:
This transition to Argument trigger some parsing error, ticket SONARIAC-580 created to track this.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
nils-werner-sonarsource: Since they will result in spaces in the resulting command, we should deal with them.

nils-werner-sonarsource (approved):
LGTM! At the end of our sprint, we should have a look at our grammar to bring a bit of structure, sections, and more comments.

PR #522, opened 2023-02-13
SONARIAC-572 Adapt HereDocForm to the new grammar

mstachniuk:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/HereDocumentImpl.java
```diff
@@ -45,13 +46,12 @@ public Kind getKind() {
 
   @Override
   public List<SyntaxToken> literals() {
-    return List.of(content);
+    throw new UnsupportedOperationException("TODO SONARIAC-579 Remove LiteralList.literals()");
   }
```
rudy-regazzoni-sonarsource, line 50: By throwing an exception in this method, it is stopping rules currently using this method from running, which prevent us from seeing what is the impact of this PR on the ruling execution.
If possible, I would recommend to have a temporarily replacement of this method content like we did for ShellForm and ExecForm, you can very probably just copy/paste the same logic there.

rudy-regazzoni-sonarsource (changes requested):
Great work ! Please see my comments, few elements to adapt.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/HereDocumentImpl.java
rudy-regazzoni-sonarsource: 👍 Perfectly fine, we didn't cover also as this is temp code that is soon to be removed.

rudy-regazzoni-sonarsource (approved):
LGTM !

iac-extensions/docker/src/main/java/org/sonar/iac/docker/utils/ArgumentUtils.java
```diff
@@ -76,6 +79,15 @@ private static String resolveExpression(Expression expression) {
...
   }
 
+  @Nullable
+  public static SyntaxToken argumentToSyntaxToken(Argument argument) {
+    String value = ArgumentUtils.resolve(argument).value();
+    if (value != null) {
+      return new SyntaxTokenImpl(value, argument.textRange(), Collections.emptyList());
+    }
+    return null;
+  }
```
nils-werner-sonarsource, line 89: Please add a comment. We should remove this helper after we drop all `literals()` methods.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/HereDocumentImpl.java
```diff
@@ -45,13 +46,16 @@ public Kind getKind() {
 
   @Override
   public List<SyntaxToken> literals() {
```
nils-werner-sonarsource, line 48: A comment is missing to drop the method.

nils-werner-sonarsource (commented):
Please add a few comments to track the method which we should remove.

PR #525, opened 2023-02-14
SONARIAC-553 Modify FromTree implementation to use Argument and Flag

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
```diff
@@ -83,8 +84,11 @@ private void handle(CheckContext ctx, DockerImage dockerImage) {
     if(!isLastDockerImageInFile(dockerImage)) {
       return;
     }
-    String imageName = dockerImage.from().image().name().value();
+    String imageName = getImageName(dockerImage.from());
     Optional<UserInstruction> lastUser = getLastUser(dockerImage);
+    if (imageName == null) {
```
mstachniuk, line 89: This `if` block can be before `getLastUser()` call - it's not necessary to call it when `imageName` is `null`.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/TestUtils.java
```diff
@@ -45,4 +46,22 @@ public static void assertKeyValuePair(NewKeyValuePair keyValuePair, String expec
...
     }
   }
+
+  public static void assertFrom(FromInstruction from, String expectedName, @Nullable String expectedFlagName, @Nullable String expectedFlagValue, @Nullable String expectedAlias) {
+    assertThat(argValue(from.image())).isEqualTo(expectedName);
+    if (expectedFlagName != null) {
+      assertThat(from.platform().name()).isEqualTo(expectedFlagName);
+      if (expectedFlagValue != null) {
+        assertThat(argValue(from.platform().value())).isEqualTo(expectedFlagValue);
+      }
```
mstachniuk, line 56: I think there should be something like:

```
} else {
    assertThat(argValue(from.platform().value())).isNull();
}
```

iac-extensions/docker/src/test/java/org/sonar/iac/docker/TestUtils.java
```diff
@@ -45,4 +46,22 @@ public static void assertKeyValuePair(NewKeyValuePair keyValuePair, String expec
       assertThat(argValue(keyValuePair.value())).isEqualTo(expectedValue);
     }
   }
+
+  public static void assertFrom(FromInstruction from, String expectedName, @Nullable String expectedFlagName, @Nullable String expectedFlagValue, @Nullable String expectedAlias) {
```
mstachniuk, line 50: Minor. There is a better way of implementing assertions in a more fluent way, see: `TextRangeAssert` and https://assertj.github.io/doc/#assertj-core-custom-assertions

then the code:

```
assertFrom(from, "foobar", null, null, null);
```

can be replaced by:

```
assertThat(from).hasName("foobar");
```

The current form is not bad, consider refactor it or not - it's up to you :-)

mstachniuk (approved):
LGTM, minor comments

iac-extensions/docker/src/test/java/org/sonar/iac/docker/TestUtils.java
rudy-regazzoni-sonarsource: Interesting approach ! I tried both, while I prefer the second usage over the first one, the implementation is quite bigger (x2), which is a bit of an overkill for an element that is not much used. I discussed with Nils about it, he also would like to keep it simple as this element is not used at many place and is not expose in a shared library.
But I will definitely keep this approach in mind for an other appropriate use case.

PR #529, opened 2023-02-15
SONARIAC-581 Track value of variables of the different scopes and instructions of a Dockerfile

nils-werner-sonarsource:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/visitors/DockerSymbolVisitorTest.java
```diff
@@ -0,0 +1,240 @@
...
+      assertThat(usage.kind()).isEqualTo(Usage.Kind.ASSIGNMENT);
+      assertThat(usage.tree()).isEqualTo(keyValuePair);
+    });
+  }
+
+  @Test
+  void argInstructionShouldCreateTwoSymbols() {
+    DockerImage image = scanImage("ARG foo1=bar foo2=bar");
+    List<NewKeyValuePair> keyValuePairs = firstDescendant(image, ArgInstruction.class).keyValuePairs();
+
```
rudy-regazzoni-sonarsource, line 78: Nitpick : I would add `assertThat(scope.getSymbols()).hasSize(2);`

iac-extensions/docker/src/test/java/org/sonar/iac/docker/visitors/DockerSymbolVisitorTest.java
```diff
@@ -0,0 +1,240 @@
...
+    NewKeyValuePair keyValuePair = firstDescendant(image, NewKeyValuePair.class);
+
+    Scope scope = image.scope();
+
+    assertThat(scope.getSymbols()).hasSize(1);
+    assertThat(image.scope().getSymbol("bar")).isNull();
+
+    Symbol symbol = image.scope().getSymbol("foo");
+    assertThat(symbol).isNotNull();
+    assertThat(symbol.usages()).allSatisfy(usage -> {
```
rudy-regazzoni-sonarsource, line 68: I would add an assert to ensure that there is only one usage, just in case, but maybe it's too much ^^'

iac-extensions/docker/src/main/java/org/sonar/iac/docker/visitors/DockerSymbolVisitor.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public void setGlobalScope(InputFileContext ctx, Body body) {
+    body.setScope(globalScope);
+  }
+
+  public void setImageScope(InputFileContext ctx, DockerImage dockerImage) {
+    currentScope = new Scope(globalScope);
+    dockerImage.setScope(currentScope);
+  }
```
rudy-regazzoni-sonarsource, line 53: We will probably need to store the inheritance information, as by default for an `Image`, inherited global variable are not accessible.
```
ARG global_var=value
FROM scratch
LABEL key=$global_var
# will give : {"key":""}
```
VS
```
ARG global_var=value
FROM scratch
ARG global_var
LABEL key=$global_var
# will give : {"key":"value"}
```

Another use case to consider : global variable are naturally always accessible in FROM instruction :
```
ARG image=scratch
FROM $image
```

Those use cases can probably be processed in separate tickets.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/visitors/DockerSymbolVisitorTest.java
```diff
@@ -0,0 +1,240 @@
...
+    Body body = scanBody( "ARG foo=bar\n" + IMAGE_PREFIX + "ARG foo=bar");
+    Scope bodyScope = body.scope();
+    Scope imageScope = body.dockerImages().get(0).scope();
+
+    assertThat(bodyScope.getSymbol("foo")).isNotNull()
+      .satisfies(symbol -> assertThat(symbol.usages()).hasSize(1));
+
+    assertThat(imageScope.getSymbol("foo")).isNotNull()
+      .satisfies(symbol -> assertThat(symbol.usages()).hasSize(2));
+  }
```
rudy-regazzoni-sonarsource, line 129: I would add a check to the value, and I would add another test case :
```
ARG foo=bar
FROM scratch
ARG foo
```
For now I guess the second `foo` will have no value, but later when we will support it we will expect it to have `bar` for value.

And for later, other test cases :
```
ARG foo=bar
FROM scratch
ARG foo
LABEL my_label=$foo
# assert the value is bar
```
and
```
ARG foo=bar
FROM scratch
LABEL my_label=$foo
# assert the value is empty string + no symbol created for the $foo in the image
```

rudy-regazzoni-sonarsource (approved):
LGTM ! Awesome work ! I would have expected the support of variable with scope to be more complicated but in the end it's quite straightforward !
Please see my comments, I didn't find any reference to the global variable things I mention in the tickets, so I added some details about it.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/visitors/DockerSymbolVisitor.java
nils-werner-sonarsource: Why do you think that they are not accessible? They should be via:
```
currentScope = new Scope(globalScope);
dockerImage.setScope(currentScope);
```

rudy-regazzoni-sonarsource: As discussed, we need to have a mention of the global variable in a ARG instruction inside the image to have access to it.
Even the [dockerfile reference](https://docs.docker.com/engine/reference/builder/#scope) is not very clear about it as they don't give a full example.

Also, I just saw that this PR does not support ENV at all, shouldn't we ? I didn't saw any reference to this in the Jira ticket.
 (better to have [SONARIAC-549](https://sonarsource.atlassian.net/browse/SONARIAC-549) completed first)
I saw also [this](https://docs.docker.com/engine/reference/builder/#using-arg-variables) in the Dockerfile reference :

> You can use an ARG or an ENV instruction to specify variables that are available to the RUN instruction. Environment variables defined using the ENV instruction always override an ARG instruction of the same name. Consider this Dockerfile with an ENV and ARG instruction.

mstachniuk:
Good stuff!

PR #530, opened 2023-02-16
SONARIAC-588 Symbol usage should know about its scope

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/symbols/Usage.java
```diff
@@ -25,18 +25,25 @@ public class Usage {
 
   public enum Kind {
     ASSIGNMENT,
-    ARGUMENT
+    ACCESS
```
nils-werner-sonarsource, line 28: IMO `ACCESS` is more suitable.

rudy-regazzoni-sonarsource (approved):
LGTM !

PR #531, opened 2023-02-16
SONARIAC-585 Remove/Replace ListLiteral class

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
I didn't manage to remove `LiteralList` fully, I sort of replaced it with `ArgumentsForm` (a better name to be found).
For me this layer is required, especially since the `ExecForm` has some elements regarding the format itself, which need to be provided by the `children()` method : `[`, `,`, `]`
Overall I tried to simplify as much as possible, removing the `instruction.arguments().arguments()` things we had and also not returning `null` value (to be challenged, it may make sense in some case, like an alone `CMD`)

nils-werner-sonarsource (approved):
LGTM! I also thought about how to get rid of the additional layer of the argument wrapper, but we will lose some information. IMO ArgumentsFrom sounds a bit odd, maybe ArguementList suits more.

PR #534, opened 2023-02-16
SONARIAC-587 Update parent pom to fix copyright year in license header

mstachniuk:


nils-werner-sonarsource (approved):
Typically we name these commits "Happy new year 2023!"

PR #535, opened 2023-02-17
SONARIAC-591 S6470 cause NullPointerException when ADD src is environment variable

mstachniuk:


nils-werner-sonarsource (approved):
LGTM!

PR #537, opened 2023-02-17
SONARIAC-533 Handle trivias inside multiline instructions

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/RunInstructionImplTest.java
```diff
@@ -315,4 +317,101 @@ void shouldCheckParseRunMultiLineMultipleRedirect() {
...
+    assertThat(tree.keyword().value()).isEqualTo("RUN");
+    assertThat(tree.arguments()).hasSize(2);
+    assertArgumentsValue(tree.arguments(), "executableparameters", "parameter2");
+
+    SyntaxTokenImpl syntaxToken = (SyntaxTokenImpl) TreeUtils.getLastDescendant(tree, SyntaxTokenImpl.class::isInstance).get();
+    assertThat(syntaxToken.value()).isEqualTo("parameter2");
+    assertThat(syntaxToken.comments())
+      .hasSize(1)
+      .extracting("value").containsExactly("# my comment");
+  }
```
rudy-regazzoni-sonarsource, line 400: Couldn't fix this : the inline comment get attached to the `SyntaxToken` `"parameter2"` instead of the merged `SyntaxToken` `"executableparameters"`.
This is because to prevent attaching the comment to the previous `SyntaxToken`, I added a check to detect when overlapping parent of the comment end at offset `0` (which is the current behavior of our elements at ends of lines).
But unfortunately this check has the side effect that we also exclude overlapping parent which are continued and last the whole next line(s) (probably a very rare case).

PR #540, opened 2023-02-20
SONARIAC-598 Kubernetes & CloudFormation slow check if the file shoul…

mstachniuk:
…d be scanned for big files


nils-werner-sonarsource (approved):
LGTM! But wouldn't it make sence to move the logic to the common YamlSensor class?
Both Sensors are Yaml sensors.

mstachniuk:
The check for CloudFormation is simpler. The common part is already in YamlSensor. I will keep it as it is for now.

PR #542, opened 2023-02-20
SONARIAC-601 Stack Overflow when resolving variable which directs to itself

nils-werner-sonarsource:
This PR simply builds a statefull wrapper around the argument resolving. In the end we could get rid of the utils class in call directly ArgumentResolver. However, I don't want to blow up the PR for now.


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #543, opened 2023-02-20
SONARIAC-590 Implement a fail fast mechanism

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/extension/IacSensor.java
```diff
@@ -194,10 +197,18 @@ private void analyseFile(InputFileContext inputFileContext) {
...
           inputFileContext.reportAnalysisError(e.getMessage(), null);
           LOG.error("Cannot analyse '" + inputFile +"': " + e.getMessage(), e);
+
+          interruptOnFailFast(inputFileContext.sensorContext, inputFile, e);
         }
       }
     }
 
+    private void interruptOnFailFast(SensorContext context, InputFile inputFile, Exception e) {
+      if (context.config().getBoolean(FAIL_FAST_PROPERTY_NAME).orElse(false)) {
```
mstachniuk, line 207: Minor. The value can be read once and cached

mstachniuk (approved):
LGTM, small comment, but it's not a blocker for merge

iac-common/src/main/java/org/sonar/iac/common/extension/IacSensor.java
nils-werner-sonarsource: Fair point. I thought about the same. But how often we should have exceptions :see_no_evil: ?

PR #544, opened 2023-02-21
SONARIAC-598 Kubernetes & CloudFormation slow check if the file should be scanned for big files

mstachniuk:
Fix new line separation on Windows


nils-werner-sonarsource (approved):
LGTM!

PR #551, opened 2023-02-22
SONARIAC-614 Should parse `$` not followed by variable identifier as a regular string

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -70,7 +70,7 @@ public class DockerLexicalConstant {
   public static final String ESCAPED_UNQUOTED_STRING_CHARACTERS = "\\\\$|\\\\'|\\\\\"";
 
   public static final String UNQUOTED_STRING_LITERAL = "(?:"
-    + "(?:" + ESCAPED_UNQUOTED_STRING_CHARACTERS + "|[^\\s'\"$])++"
+    + "(?:" + ESCAPED_UNQUOTED_STRING_CHARACTERS + "|[^\\s'\"\\$]|\\$(?![a-zA-Z_0-9{]))++"
```
mstachniuk, line 73: 1. It can be improved by adding `<`: `"|[^\\s'\"\\$]|\\$(?<![a-zA-Z_0-9{]))++"`
2. I don't like using `^` and `!` for negation in one regex. Is it possible to use `^` in the second part? Or meaning of `!` is here different?

mstachniuk (approved):
LGTM, but please see my comment

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
nils-werner-sonarsource: 1. The regex behaves different when adding a `<` it would change a lookahead to a lookbehind.
2. One it's a negative character class and the other is a negative look ahead. Don't know how to avoid one or the other.

mstachniuk: ok fine.

PR #552, opened 2023-02-23
SONARIAC-613 Parsing errors should be mapped to the real code

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Difficult to have more code coverage :
- the throw of the `new RecognitionException()` seems to be a FP of the code coverage tool
- the uncovered else for `m.find()` is a defensive check in case we have a `RecognitionException` of a different format (should not happen), I would still prefer to keep it
- the uncovered cases in `isNewLine()` seems to come from the fact that the code coverage is run on only one specific system (Unix ?) while this code handle both Unix and Windows like newline forms (`\n` and `\r\n`).

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
```diff
@@ -74,4 +81,52 @@ private static void setParents(DockerTree tree) {
...
+        String newErrorMessage = buildNewErrorMessage(correctedLineAndColumn[0], correctedLineAndColumn[1], rest);
+        throw new RecognitionException(correctedLineAndColumn[0], newErrorMessage, originalException.getCause());
+      } else {
+        return originalException;
+      }
+    }
+
+    private static String buildNewErrorMessage(int line, int column, String rest) {
+      return LINE_MESSAGE + line + COLUMN_MESSAGE + column + rest;
+    }
```
nils-werner-sonarsource, line 109: Instead of a concat like this, maybe use a format string instead.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
```diff
@@ -74,4 +81,52 @@ private static void setParents(DockerTree tree) {
...
+
+    public static RecognitionException adjustLineAndColumnNumber(RecognitionException originalException, DockerPreprocessor.PreprocessorResult preprocessorResult) {
+      Matcher m = RECOGNITION_EXCEPTION_LINE_COLUMN_PATTERN.matcher(originalException.getMessage());
+      if (m.find()) {
+        int line = Integer.parseInt(m.group("line"));
+        int column = Integer.parseInt(m.group("column"));
+        String rest = m.group("rest");
+        int index = computeIndexFromLineAndColumn(preprocessorResult.processedSourceCode(), line, column);
+        int[] correctedLineAndColumn = preprocessorResult.sourceOffset().sourceLineAndColumnAt(index);
+        String newErrorMessage = buildNewErrorMessage(correctedLineAndColumn[0], correctedLineAndColumn[1], rest);
```
nils-werner-sonarsource, line 100: `String.format(PARSING_ERROR_MESSAGE, line, column, rest)`
with PARSING_ERROR_MESSAGE = "Parse error at line %d column %d %s"

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
```diff
@@ -74,4 +81,52 @@ private static void setParents(DockerTree tree) {
...
   }
+
+  static class RecognitionExceptionAdjuster {
+    private static final String LINE_MESSAGE = "Parse error at line ";
+    private static final String COLUMN_MESSAGE = " column ";
+    private static final Pattern RECOGNITION_EXCEPTION_LINE_COLUMN_PATTERN = Pattern.compile(LINE_MESSAGE + "(?<line>\\d+)" + COLUMN_MESSAGE + "(?<column>\\d+)(?<rest>.*)");
+
+    private RecognitionExceptionAdjuster() {}
+
+    public static RecognitionException adjustLineAndColumnNumber(RecognitionException originalException, DockerPreprocessor.PreprocessorResult preprocessorResult) {
```
nils-werner-sonarsource, line 92: Maybe provide the SourceOffset directly?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
nils-werner-sonarsource, line 100: We may want to change `RecognitionException` on SSLR side because it's build with a `TextRange` which would could perfectly use here.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
```diff
@@ -74,4 +81,52 @@ private static void setParents(DockerTree tree) {
...
+      Matcher m = RECOGNITION_EXCEPTION_LINE_COLUMN_PATTERN.matcher(originalException.getMessage());
+      if (m.find()) {
+        int line = Integer.parseInt(m.group("line"));
+        int column = Integer.parseInt(m.group("column"));
+        String rest = m.group("rest");
+        int index = computeIndexFromLineAndColumn(preprocessorResult.processedSourceCode(), line, column);
+        int[] correctedLineAndColumn = preprocessorResult.sourceOffset().sourceLineAndColumnAt(index);
+        String newErrorMessage = buildNewErrorMessage(correctedLineAndColumn[0], correctedLineAndColumn[1], rest);
+        throw new RecognitionException(correctedLineAndColumn[0], newErrorMessage, originalException.getCause());
```
nils-werner-sonarsource, line 101:
```suggestion
        return new RecognitionException(correctedLineAndColumn[0], newErrorMessage, originalException.getCause());
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
```diff
@@ -74,4 +81,52 @@ private static void setParents(DockerTree tree) {
...
+      } else {
+        return originalException;
```
nils-werner-sonarsource, line 103: You can cover this line by the following code:
```
@Test
  void shouldNotCreateNewExceptionOnInvalidMessageFormat() {
    RecognitionException exception = new RecognitionException(0, "InvalidMessage");
    assertThat(DockerParser.RecognitionExceptionAdjuster.adjustLineAndColumnNumber(exception, null)).isSameAs(exception);
  }
```

nils-werner-sonarsource (approved):
LGTM! I've added a couple of comments where you can adjust as well as increase the coverage.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerParser.java
rudy-regazzoni-sonarsource: 👍 Indeed that could be a nice improvement.
I see you assigned a task for yourself in the google doc "SSLR error reporting improvement", If you want you can assign this task to me and I will put the explanation directly.

PR #553, opened 2023-02-23
SONARIAC-620 S6472 should not raise NPE on missing `ENV` value

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/EnvSecretCheck.java
```diff
@@ -77,7 +77,8 @@ public void initialize(InitContext init) {
   }
 
   private static void checkEnvVariableAssignment(CheckContext ctx, KeyValuePair envVarAssignment) {
-    if (isSensitiveName(ArgumentUtils.resolve(envVarAssignment.key()).value()) && isSensitiveValue(ArgumentUtils.resolve(envVarAssignment.value()).value())) {
+    if (isSensitiveName(ArgumentUtils.resolve(envVarAssignment.key()).value())
+      && envVarAssignment.value() != null && isSensitiveValue(ArgumentUtils.resolve(envVarAssignment.value()).value())) {
```
mstachniuk, line 81: I would suggest hiding not `null` check inside `resolve()` method - it can be beneficial in other checks

mstachniuk (approved):
LGTM, but please check my comment

PR #559, opened 2023-02-27
SONARIAC-624 Modify RUN instruction to have whitespace between argument as part of Arguments

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
I splitted my ticket, as I needed a modification on the `RUN` instruction : it need to have the whitespaces embedded in its arguments (as we do for `KeyValuePair`).
I introduced a new expression `BlankLiteralImplt`, dedicated to store blank characters as an expression, so that it is easier to recognize blank argument. This was needed for `TransferInstruction` so that `srcs()` and `dest()` methods acts as before.

nils-werner-sonarsource (changes requested):
IMO this is not the way to go. You've changed a lot in the grammar and trees to be able to handle whitespaces. You should only handle whitespaces in the SHELL_FORM_GENERIC to avoid having to adjust the transfer statements. My suggestion would be to create a `TreeFactory.shellFormGeneric` that can handle whitespace instead of extending the `TreeFactory.shellForm`. Next: Why not reuse `LiteralImpl` for whitespace? More trees means more complexity. 

```
  public ShellForm SHELL_FORM_GENERIC() {
    return b.<ShellForm>nonterminal(DockerLexicalGrammar.SHELL_FORM_GENERIC).is(
      f.shellFormGeneric(
        f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), ARGUMENT_GENERIC()),
        b.zeroOrMore(
          b.firstOf(WHITESPACE_ARGUMENT(), ARGUMENT_GENERIC())
        )
      )
    );
  }

  public Argument WHITESPACE_ARGUMENT() {
    return b.<Argument>nonterminal().is(
      f.whitespaceArgument(
        b.token(DockerLexicalGrammar.WHITESPACE)
      )
    );
  }
```
See my suggestion. This way you only have to add two more simple methods to the TreeFactory, add one and adapt an existing grammar rule.

rudy-regazzoni-sonarsource:
@nils-werner-sonarsource Mmhh I didn't think of it that way, indeed seems much simpler, and it left `TransferInstruction` untouched.
I will try this way, thanks !

nils-werner-sonarsource (approved):
LGTM! You improved it a lot. However, after our morning discussions, I wonder if we need these explicit whitespaces. If it is okay for you, I would not merge this PR for now and have it as a fallback in case we need this feature.

PR #564, opened 2023-03-01
Set fail-fast property for ITs

nils-werner-sonarsource:
We can make use of the fail-fast property for our ruling ITs. This way we can already identify issues within checks before merging on the master.


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #565, opened 2023-03-01
SONARIAC-566 Rule S4507: Delivering code in production with debug features activated is security-sensitive

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/DebugModeCheck.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  public void initialize(InitContext init) {
+    init.register(EnvInstruction.class, DebugModeCheck::checkEnvDebug);
+  }
+
+  private static void checkEnvDebug(CheckContext ctx, EnvInstruction envInstruction) {
+    for (KeyValuePair variable : envInstruction.environmentVariables()) {
+      String name = ArgumentUtils.resolve(variable.key()).value();
+      String value = ArgumentUtils.resolve(variable.value()).value();
+      if (name != null && value != null && (isDevEnv(name, value) || isDebugMode(name, value) || isPhpXDebugMode(name, value))) {
```
nils-werner-sonarsource, line 51: Nitpick: You could move the three method calls to a wrapper method to make the condition with AND and OR more readable.

nils-werner-sonarsource (approved):
LGTM! See my nitpick comment and maybe I missed it, but did you test KeyValuePair without `=` as well?

rudy-regazzoni-sonarsource:
> LGTM! See my nitpick comment and maybe I missed it, but did you test KeyValuePair without `=` as well?

Good point ! I had some example that I prepared without '=', but I lost this commit (I apparently not only tend to lose Github comment, but also commit now...)
I will fix that, thanks !

PR #568, opened 2023-03-02
SONARIAC-629 Adapt Heredoc form parsing to be closer to what Exec and Shell form do

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Not ready to merge, need to be reworked a bit and polished, just want to ensure this is an acceptable solution before moving forward.

rudy-regazzoni-sonarsource:
Seems we report 3 new TP, however there is an issue with the highlighting.
If this is confirmed they are TP, I will create a ticket to fix that (either the highlighting or the FP).

nils-werner-sonarsource (approved):
LGTM!

PR #569, opened 2023-03-02
SONARIAC-633 ArgumentUtils::resolve should not allow a nullable argument

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #570, opened 2023-03-03
SONARIAC-632 Global ARG variable should be accessible to any FROM instruction

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (changes requested):
I see your point of adding the possibility to register a leave action when traversing the AST. But I don't see a reason for providing this in the common TreeVisitor. To reduce the impact you can move the changes to the `DockerSymbolVisitor` itself. Just make the visit method protected to allow you to override it for a certain visitor.

PR #573, opened 2023-03-06
SONARIAC-567 Rule S6497: Using a container image based on its digest is security-sensitive

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM!

PR #580, opened 2023-03-07
SONARIAC-635 Invalid text range for heredoc argument

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Really nice improvement, and as you said the changes are constrained in a dedicated classes, which is great.

PR #582, opened 2023-03-07
SONARIAC-607 Move regex of grammar rules to DockerLexicalConstant

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -81,9 +81,12 @@ public class DockerLexicalConstant {
     + "(?:" + ESCAPED_UNQUOTED_STRING_CHARACTERS + "|[^\\s'\"$=])++"
     + ")";
 
-  public static final String HEREDOC_NAME = "(?:"
-    + "<<-?\"?[a-zA-Z0-9_]+\"?"
-    + ")";
+  public static final String HEREDOC_NAME = "<<-?\"?([a-zA-Z0-9_]++)\"?";
+
+  public static final String HEREDOC_EXPRESSION = "(?:" + HEREDOC_NAME + "\\s+)+[\\s\\S]*?([\\n\\r])\\1(?=[\\n\\r]|$)";
```
mstachniuk, line 86: Please comment on what those 2 RegEx are doing, because they are complex.

mstachniuk (approved):

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
rudy-regazzoni-sonarsource: Good point !
Instead of explaining the regex in detail, I described what it is matching and added a link to the reference, WDYT ?

mstachniuk: This is not exactly what I was thinking about. It's clear that it is for parsing heredoc because a field name points to that. Also, our parser is somehow based on Docker reference documentation. 
I would like to see something like this:

```
It starts with the heredoc indicator (2 Less-than signs), followed by an optional minus, followed by an optional double quote then the 1st capturing group, followed by an optional double quote. In 1st capturing group: one or more (possessive) uppercase/lowercase letters, numbers, or underscore.
```

WDYT? Is it valuable?
Sites like https://regex101.com/ or https://regexr.com/ can help you write such descriptions.

rudy-regazzoni-sonarsource: OK I see ! That's a good point.

PR #585, opened 2023-03-07
SONARIAC-622 Rework Log Message of Parsing Error

mstachniuk:


iac-common/src/test/java/org/sonar/iac/common/extension/IacSensorTest.java
```diff
@@ -65,7 +65,14 @@
 class IacSensorTest extends AbstractSensorTest {
 
   TreeParser<Tree> testParser = (source, inputFileContext) -> {
-    throw new RuntimeException();
+    throw new RuntimeException("RuntimeException message");
+  };
```
rudy-regazzoni-sonarsource, line 69: I would rename this variable to `testParserThrowsRuntimeException`, so that we directly/clearly understand on usage that this is not gonna parse anything but will just throw an error.

iac-common/src/test/java/org/sonar/iac/common/extension/IacSensorTest.java
```diff
@@ -355,14 +409,53 @@ protected List<TreeVisitor<InputFileContext>> visitors(SensorContext sensorConte
...
+      }
+
+      @Override
+      protected List<TreeVisitor<InputFileContext>> visitors(SensorContext sensorContext, DurationStatistics statistics) {
+        return Collections.singletonList(new ChecksVisitor(checkFactory.create(repositoryKey()), statistics));
+      }
+
+      @Override
+      protected String getActivationSettingKey() {
+        return "testsensor.active";
```
rudy-regazzoni-sonarsource, line 458: Instead of tripling this method code, I would instead modify the origin sensor method and add a third parameter `TreeParser<Tree> treeParser`
This avoid the code duplication at the code of slightly heavier call to `sensor`.

iac-common/src/test/java/org/sonar/iac/common/testing/ExtensionSensorTest.java
```diff
@@ -74,12 +75,16 @@ void shouldRaiseIssueOnParsingErrorWhenIssueActive() {
...
+    assertThat(logTester.logs(LoggerLevel.ERROR)).hasSize(1);
+    String errorMessage = "Cannot parse '%s:1:1'";
+    assertThat(logTester.logs(LoggerLevel.ERROR).get(0))
+      .startsWith(String.format(errorMessage, inputFile.filename()));
+
+    verifyDebugMessages(logTester.logs(LoggerLevel.DEBUG));
   }
 
+  protected abstract void verifyDebugMessages(List<String> logs);
+
```
rudy-regazzoni-sonarsource, line 87: I would move this method declaration with the other, at the beginning of the class declaration.

PR #586, opened 2023-03-07
SONARIAC-617 Track ENV arguments similar to ARG arguments when resolving variables

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/symbols/Symbol.java
```diff
@@ -44,7 +44,7 @@ public void addUsage(Scope scope, DockerTree tree, Usage.Kind kind) {
     Usage usage = new Usage(scope, tree, kind);
     usages.add(usage);
     if (tree instanceof HasSymbol) {
-      ((HasSymbol) tree).setSymbol(this);
+      ((HasSymbol) tree).setSymbol(new Symbol(this));
     }
```
rudy-regazzoni-sonarsource, line 48: I understand what we want to achieve this, this is to keep each symbol reference with only previous usages, in order to easily find the last assigned value in `findLastAccessibleAssignedValue`.
It seems to be the contrary to what we usually want to achieve, meaning having a common symbol object.
I guess this is also why you are not satisfied with this solution ^^ I'm fine going this way for now to unblock other tickets.
Otherwise, I remember that in Python we had a similar things, with finding last proper usage using line number. Maybe we could have done similar things ? Meaning we keep a single shared symbol object, and in `findLastAccessibleAssignedValue` we also provide a line number in order to not take care of usages after this line number.
It's just an idea, I may miss part of the problem.

rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comment.

PR #587, opened 2023-03-08
SONARIAC-643 Get rid of ArgumentUtils::resolveAndMerge

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
```diff
@@ -84,27 +86,36 @@ private void handle(CheckContext ctx, DockerImage dockerImage) {
...
+  }
+
+  private static void checkLastUserInstruction(CheckContext ctx, UserInstruction userInstruction) {
+    for(Argument user: userInstruction.arguments()) {
+      String userName = ArgumentUtils.resolve(user).value();
+      if (UNSAFE_USERS.contains(userName)) {
+        ctx.reportIssue(userInstruction, String.format(MESSAGE_ROOT_USER, userName));
+      }
+    }
+  }
```
rudy-regazzoni-sonarsource, line 103: It's not really working as before, as it is checking each USER argument individually.
So currently, `USER my user root` would raise an issue, while it was not before.

From what I read, this is apparently possible in Windows system to have spaces in username :D `John Smith` is a valid user name (although this apparently cause a lot of trouble for poor user who have put spaces in their username, especially for dev trying to install some tool/package which use %USERNAME% env variable).
I was able to create one in my system :
image


Still our current implementation or Argument does not really permit this, as we don't store whitespace. Probably this could be fixed in another ticket.
This also mean that we would still need a method like `resolveAndMerge()` to merge an `Argument` list as a single string.

I hope that most dev/admin does not use user with whitespace in it, still this is possible.

rudy-regazzoni-sonarsource (changes requested):
I did some more research, apparently username with spaces is a thing in some systems (Windows :D), so currently this PR is preventing some use case from working.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PrivilegedUserCheck.java
nils-werner-sonarsource: Nice catch and thank you for testing. I will adapt the check impl.

rudy-regazzoni-sonarsource (approved):
LGTM ! Indeed no more need for resolveAndMerge, nice job !

PR #588, opened 2023-03-08
SONARIAC-644 ArgumentResolution should reflect resolution status

nils-werner-sonarsource:
I had to rework the `PrivilegedUserCheckTest` because debugging was hard. I refactored `ArgumentUtils` a lot and introduced a builder for ArgumentResolution. It may be over-engineered, but it makes things a bit cleaner. Missing converge in the SymbolVisitor can be reached with SONARIAC-646.


iac-extensions/docker/src/main/java/org/sonar/iac/docker/utils/ArgumentUtils.java
```diff
@@ -38,68 +38,79 @@
...
-    private ArgumentResolution resolveSymbol(Symbol symbol) {
+    private void resolveVariable(Variable variable) {
+      Symbol symbol = variable.symbol();
+      if (!visitedVariable.add(variable) || symbol == null) {
+        resolution.addPlaceholder(variable);
+        return;
+      }
+
       List<Usage> usages = symbol.usages();
       Collections.reverse(usages);
```
rudy-regazzoni-sonarsource, line 115: I just saw that this will create a bug.
By calling `Collection.reverse()`, it reverse the elements of the provided list.
Since `symbol.usages()` provide the original list of usages, everytime we resolve a variable, we reverse the order of usages for the symbol.
This is easily testable in `ArgumentUtilsTest.shouldResolveLabelValue()`, by doing a second resolution :
```
ArgumentUtils.ArgumentResolution resolution = ArgumentUtils.resolve(label.value());
assertThat(resolution.value()).isEqualTo("bar");

ArgumentUtils.ArgumentResolution resolution2 = ArgumentUtils.resolve(label.value());
assertThat(resolution2.value()).isEqualTo("bar");
```
A possible solution could be to use `Lists.reverse()`, which return a view and does not modify the original List (and also is an operation which take constant time instead of linear time).

rudy-regazzoni-sonarsource (approved):
LGTM !
I'm a bit concerned with the resolution of unresolved variable by a placeholder, as docker would instead replace it with empty string. But still I don't think this will reduce detection/precision apart maybe in some edgy case, so I'm okay with it.
I like how you improved the test for `PriviligedUserCheck`, we could even go further by naming file after what they are checking (ex : `NonCompliant/Dockerfile.14` -> `NonCompliant/Dockerfile.checkSensitiveImageFromArg`). Still it make the PR a bit big and is not directly related to the ticket, could maybe have benefited to be in its own PR ;)
Also I found a bug not related to the PR just by chance, see my comment.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/utils/ArgumentUtils.java
nils-werner-sonarsource: Ah, nice catch. I had this issue this week when playing around with the implementation. I will fix it.

PR #589, opened 2023-03-08
SONARIAC-637 Don't use Plugin's API TextPointer and TextRange in Trees

karim-ouerghemmi-sonarsource:
Looks like many changes but, in most locations, it is just adapting the import.

See comment https://github.com/SonarSource/sonar-iac/pull/589#discussion_r1132316892 .

Adding @nils-werner-sonarsource also as a reviewer.


iac-common/src/main/java/org/sonar/iac/common/extension/visitors/SyntaxHighlightingVisitor.java
```diff
@@ -40,17 +46,25 @@ protected SyntaxHighlightingVisitor() {
...
 
   @Override
   protected void after(InputFileContext ctx, Tree root) {
     newHighlighting.save();
   }
 
-  protected void highlight(HasTextRange range, TypeOfText typeOfText) {
-    newHighlighting.highlight(range.textRange(), typeOfText);
+  protected void highlight(HasTextRange tree, TypeOfText typeOfText) {
+    if (tree.textRange().end().compareTo(tree.textRange().start()) <= 0) {
```
karim-ouerghemmi-sonarsource, line 60: The check was added after seeing noticing [YamlHighlightingVisitorTest#scalar_key](https://github.com/SonarSource/sonar-iac/blob/32ba45207106a08be9fe3b65a05283c701ae2a39/iac-common/src/test/java/org/sonar/iac/common/yaml/visitors/YamlHighlightingVisitorTest.java#L50) failing. Reason is that [InputFile#newRange](https://github.com/SonarSource/sonar-plugin-api/blob/af6f59056ad3258572ca76c47f4802b54063bfa2/plugin-api/src/main/java/org/sonar/api/batch/fs/InputFile.java#L181) actually checks that the range makes sense and raises an exception if not. 
In Yaml we currently parse `key:` to a `Mapping` with a `Tuple` whose key is a `ScalarTree` `"key:"` and it's value is a `ScalarTree` that represents an empty String and hence startColumn == endColumn

iac-common/src/main/java/org/sonar/iac/common/extension/visitors/InputFileContext.java
```diff
@@ -108,6 +109,10 @@ public void reportAnalysisError(String message, @Nullable TextPointer location)
     error.save();
   }
 
+  private org.sonar.api.batch.fs.TextRange toInputFileRange(TextRange textRange) {
```
navendu-barua-sonarsource, line 112: is TextRange import correct here? or should it be used via the new ones?

iac-common/src/main/java/org/sonar/iac/common/api/tree/impl/TextPointer.java
```diff
@@ -0,0 +1,62 @@
...
+  public int compareTo(TextPointer other) {
+    if (this.line == other.line()) {
+      return Integer.compare(this.lineOffset, other.lineOffset());
+    }
+    return Integer.compare(this.line, other.line());
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) return true;
```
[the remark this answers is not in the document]
nils-werner-sonarsource: Usually, in equal methods, we don't have curly braces. However, it's a fair point. I don't have a strong opinion.

iac-common/src/main/java/org/sonar/iac/common/extension/visitors/InputFileContext.java
nils-werner-sonarsource: I would say yes. This method converts the new `TextRange' to the API interface.

iac-extensions/terraform/src/test/java/org/sonar/iac/terraform/plugin/TerraformSensorTest.java
```diff
@@ -64,7 +65,8 @@ void test_one_rule() {
     IssueLocation location = issue.primaryLocation();
     assertThat(location.inputComponent()).isEqualTo(inputFile);
     assertThat(location.message()).isEqualTo("Rename tag key \"anycompany:cost-center\" to match the regular expression \"^([A-Z][A-Za-z]*:)*([A-Z][A-Za-z]*)$\".");
-    TextRangeAssert.assertTextRange(location.textRange()).hasRange(2, 11, 2, 35);
+    TextRangeAssert.assertTextRange(TextRanges.range(location.textRange().start().line(), location.textRange().start().lineOffset(), location.textRange().end().line(),
```
nils-werner-sonarsource, line 68: As this is quite hard to read, maybe you can use a variable for `location.textRange()` and for the results of `TextRanges.range()`

nils-werner-sonarsource (approved):
LGTM! Please see my nitpick comment.

iac-common/src/main/java/org/sonar/iac/common/extension/visitors/InputFileContext.java
```diff
@@ -53,18 +54,18 @@ public void reportIssue(RuleKey ruleKey, @Nullable TextRange textRange, String m
...
-        issueLocation.at(textRange);
+      if (textRange != null && TextRanges.isValidAndNotEmpty(textRange)) {
+        issueLocation.at(toInputFileRange(textRange));
       }
 
       issue.forRule(ruleKey).at(issueLocation);
 
       secondaryLocations.stream()
-        .filter(Objects::nonNull)
+        .filter(l -> l != null && TextRanges.isValidAndNotEmpty(l.textRange))
```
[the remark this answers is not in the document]
karim-ouerghemmi-sonarsource: Good point, as far as I know, we do not have an agreed-upon format for this here. Personally, I sometimes use short names when it is very clear what they refer to (as here the line `secondaryLocations.stream()` is directly above, it is clear that it is secondaryLocations). For a bit more clarity I renamed it to `location`. Cannot hurt indeed. Thanks for pointing this out :slightly_smiling_face:

PR #595, opened 2023-03-09
SONARIAC-577 Rule S6472: Add ARG secret handling detection

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsHandlingCheck.java
```diff
@@ -125,21 +157,31 @@ private static boolean containsSecretEntityWordCombination(List<String> words) {
...
+      }
+      if (CAMELCASE_NAME_PATTERN.matcher(name).matches()) {
+        return toUpperCase(CAMELCASE_SPLIT_PATTERN.split(name));
+      }
+      return Collections.emptyList();
+    }
+
+    private static List<String> toUpperCase(String[] strings) {
+      return Stream.of(strings).map(s -> s.toUpperCase(Locale.ROOT)).collect(Collectors.toList());
+    }
```
rudy-regazzoni-sonarsource, line 184: Possible suggestion : an alternative could be to call `toUpperCase()` on String name before working on it.
The cons is that even if we don't match anything, we will do an operation which costs may not be negligible.
The pros is we simplify a bit the code and don't do streams/collect, which may also be costly, but only happened for matched result.
I don't have a strong opinion on this, I let you decide.

iac-extensions/docker/src/test/resources/checks/SecretsHandlingCheck/SecretsHandlingCheck_arg.dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# Noncompliant@+1
+ARG ACCESS_TOKEN=${ACCESS_TOKEN}
+#   ^^^^^^^^^^^^
+
+ARG ACCESS_TOKEN=${RANDOM}
+
+ARG CCC
+ARG ${CCC}=BBB
+
+ARG ACCESS_TOKEN=""
```
rudy-regazzoni-sonarsource, line 24: From the ticket description, this apparently should be NonCompliant.
Also an extra usecase : `ARG ACCESS_TOKEN`

iac-extensions/docker/src/test/resources/checks/SecretsHandlingCheck/SecretsHandlingCheck_arg.dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM foo
+
+
+# Noncompliant@+1 {{Make sure that using ARG to handle a secret is safe here.}}
+ARG ACCESS_TOKEN=ThisIsSomethingThatShouldProbablyBeSecret
+#   ^^^^^^^^^^^^
```
rudy-regazzoni-sonarsource, line 6: Possible missing usecase in lowercase : `ARG access_token=ThisIsSomethingThatShouldProbablyBeSecret`

iac-extensions/docker/src/test/resources/checks/SecretsHandlingCheck/SecretsHandlingCheck_arg.dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# Noncompliant@+1
+ARG TOKEN=AAAA
+# Noncompliant@+1
+ARG token=AAAA
+# Noncompliant@+1
+ARG ACCESS_TOKEN="hts://www.secrets.com"
```
rudy-regazzoni-sonarsource, line 12: Possible extra usecase :
```
ARG path=http://www.secrets.com
ARG TOKEN=$path
```
Or even
```
ARG secret=http://www.secrets.com
ARG TOKEN=$secret
```

rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comments.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsHandlingCheck.java
```diff
@@ -75,37 +80,64 @@ public class EnvSecretCheck implements IacCheck {
...
-  private static boolean isSensitiveName(@Nullable String name) {
-    if (name == null) return false;
-    List<String> words = splitEnvVarName(name);
-    return isSecretWordOnly(words) || containsSecretEntityWordCombination(words);
+  /**
+   * Check if the left hand of the assignment is a sensitive
+   */
+  private static boolean isSensitiveName(Argument nameArgument) {
+    ArgumentResolution nameResolution = ArgumentResolution.of(nameArgument);
+    if (!nameResolution.is(RESOLVED)) {
```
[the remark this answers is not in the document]
nils-werner-sonarsource: This way Status.EMPTY is also covered

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsHandlingCheck.java
```diff
@@ -55,15 +59,16 @@ public class EnvSecretCheck implements IacCheck {
...
+  // Patterns to split the identifier of a variable into separate words
   private static final Pattern UNDERSCORE_NAME_PATTERN = Pattern.compile("^\\w+$");
   private static final Pattern DASH_NAME_PATTERN = Pattern.compile("^[a-zA-Z0-9-]+$");
-
   private static final Pattern CAMELCASE_NAME_PATTERN = Pattern.compile("^[A-Za-z]+$");
-
   private static final Pattern CAMELCASE_SPLIT_PATTERN = Pattern.compile("(?<!(^|[A-Z]))(?=[A-Z])|(?<!^)(?=[A-Z][a-z])");
 
+  // Pattern to identify a URL
   private static final Pattern URL_PATTERN = Pattern.compile("^(http|ftp)s?://");
```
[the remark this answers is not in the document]
nils-werner-sonarsource: Do you think this is intended by the rule?

iac-extensions/docker/src/test/resources/checks/SecretsHandlingCheck/SecretsHandlingCheck_arg.dockerfile
nils-werner-sonarsource: Will add.

iac-extensions/docker/src/test/resources/checks/SecretsHandlingCheck/SecretsHandlingCheck_arg.dockerfile
nils-werner-sonarsource: Okay. Nice catch.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsHandlingCheck.java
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: To my understanding we want to check if the value is a secret or not, by excluding some known pattern.
One of them is detecting if it looks like an URL.

I like the conservative approach that "we don't know which protocol can be used, but potentially any" and not just http/ftp, even if the probably represent 90%+ use case. Since I'm not proficient with all existing protocols, I prefer to see larger, for example in my previous company some tools were using sftp and not ftps, maybe also an sftp url can exist ? Dunno :D
Also there is still a risk that we exclude a real secret token matching this more general regex.

I don't have a strong opinion on this, we can also go the current way : exclude http/ftp as it should cover most cases, and if eventually later a user report a FP, we can extend or choose another approach.

PR #605, opened 2023-03-14
SONARIAC-652 Migrate Tree methods from TestUtils to TreeUtils

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
Cannot reach more coverage apparently, I specifically added a test class for this, It's like it's looking for the missing case `root == null` + `root.children().isEmpty()`, which of course is impossible.

nils-werner-sonarsource (commented):
LGTM!

PR #606, opened 2023-03-14
SONARIAC-655 Using weak hashing algorithms with OpenSSL is security-sensitive

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static class GroupStringWithTextRange {
+      private final List<String> strings;
+      private final TextRange textRange;
+
+      public GroupStringWithTextRange(List<Argument> arguments, List<String> resolvedArguments) {
+        this.textRange = arguments.isEmpty() ? null : TextRanges.merge(arguments.stream().map(HasTextRange::textRange).collect(Collectors.toList()));
+        this.strings = resolvedArguments;
+      }
+    }
```
rudy-regazzoni-sonarsource, line 155: Baby step to retain the `textRange` of resolved argument + report issue on the whole `Command` block.
Next baby step is to only report the issue on related element : `openssl` keyword + `-md5` option for example.
A solution could be to keep the Argument object reference in ArgumentResolution, so that we can resolve argument, get/check the value and later access the original Argument object to report the issue on it.
WDYT ?

rudy-regazzoni-sonarsource:
Baby step on the rule Rule S4790: Using weak hashing algorithms is security-sensitive
More to do :
- SONARIAC-656 : detect issue on also `shasum`
- SONARIAC-657 : don't highlight the full command but only related element/option

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,156 @@
...
+  private static boolean isOpenSslSensitiveSubCommand(Command command) {
+    return OPENSSL_SENSITIVE_SUBCOMMAND.contains(command.parameters.get(0));
+  }
+
+  private static boolean isOpenSslSensitiveDgstSubCommand(Command command) {
+    return "dgst".equals(command.parameters.get(0))
+      && OPENSSL_SENSITIVE_DGST_OPTION.stream().anyMatch(command.options::containsKey);
+  }
+
+  private static class Command {
```
rudy-regazzoni-sonarsource, line 74: Introducing the `Command` class, which goal is to parse provided list of `Argument` to split it (`&&` and `|`) and transform it into a list of `Command` with executable/options/parameters + `textRange` to process them individually.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
nils-werner-sonarsource: Do you think [SONARIAC-647](https://sonarsource.atlassian.net/browse/SONARIAC-647) is the solution for this? PR is ready for this.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
nils-werner-sonarsource: Yeah, good idea. You also need to include `;` which is a valid delimiter. However, this "simple" split approach can be dangerous when we have nested elements. 
```
CMD foo(bar1 && bar2)
```
Here the split is incorrect.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
rudy-regazzoni-sonarsource: Indeed it would totally solve that ! Let me know if/when you want me to review this PR, I don't see it in the list currently.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
rudy-regazzoni-sonarsource: Mmhh indeed, we need at least for now to document such failing case.
This approach is too naïve to handle all cases properly, I'm thinking about it like a babystep to provide a solution with a good balance between time spent and detection/precision.
I will do more search/test to see what cases it can handle and if we can improve it simply.

rudy-regazzoni-sonarsource:
Simpler approach : searching in all arguments and not splitting them by block.
Also introducing a `CommandDetector` class to easily adapt it to other rules.
Possible optimization : resolve argument only once by keeping `Argument` object reference in `ArgumentResolver`.

iac-common/src/test/java/org/sonar/iac/common/api/tree/impl/TextRangesTest.java
```diff
@@ -56,6 +58,19 @@ void test_merge_no_range() {
       .withMessage("Can't merge 0 ranges");
   }
 
+  @Test
+  void test_mergeRange() {
+    TestHasTextRange testRange1 = new TestHasTextRange(1,2, 3, 4);
+    TestHasTextRange testRange2 = new TestHasTextRange(5,6, 7, 8);
+    assertThat(TextRanges.mergeRanges(Arrays.asList(testRange1, testRange2)))
+      .isEqualTo(TextRangesTest.range(1,2, 7, 8));
```
nils-werner-sonarsource, line 66:
```suggestion
      .isEqualTo(TextRangesTest.range(1, 2, 7, 8));
```

iac-common/src/main/java/org/sonar/iac/common/api/tree/impl/TextRanges.java
```diff
@@ -51,6 +53,10 @@ public static TextRange merge(List<TextRange> ranges) {
     );
   }
 
+  public static TextRange mergeRanges(List<? extends HasTextRange> hasRanges) {
+    return hasRanges.isEmpty() ? null : merge(hasRanges.stream().map(HasTextRange::textRange).collect(Collectors.toList()));
+  }
```
nils-werner-sonarsource, line 58: I think this will duplicate the methods Jonas has already implemented. When returning null you should annotate this utils method.

iac-common/src/test/java/org/sonar/iac/common/api/tree/impl/TextRangesTest.java
```diff
@@ -56,6 +58,19 @@ void test_merge_no_range() {
       .withMessage("Can't merge 0 ranges");
   }
 
+  @Test
+  void test_mergeRange() {
+    TestHasTextRange testRange1 = new TestHasTextRange(1,2, 3, 4);
+    TestHasTextRange testRange2 = new TestHasTextRange(5,6, 7, 8);
```
nils-werner-sonarsource, line 64: Maybe you can also checkout Jonas test case which avoids having a dedicated Test class.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+    CommandDetector sensitiveOpenSslSubcommand = CommandDetector.builder()
+      .with("openssl"::equals)
+      .with(OPENSSL_SENSITIVE_SUBCOMMAND::contains)
+      .build();
+    CommandDetector sensitiveOpenSslDgst = CommandDetector.builder()
+      .with("openssl"::equals)
+      .with("dgst"::equals)
+      .withOptional(s -> !OPENSSL_SENSITIVE_DGST_OPTION.contains(s) && s.startsWith("-"))
+      .with(OPENSSL_SENSITIVE_DGST_OPTION::contains)
+      .build();
```
nils-werner-sonarsource, line 59: Since these detectors are not dependent on the context or instruction, you could make a constant out of it.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public static Builder builder() {
+      return new Builder();
+    }
+
+    /**
+     * Return all block of arguments which match the list of predicates.
+     */
+    private List<List<Argument>> search(List<Argument> arguments) {
+      List<ArgumentResolution> resolvedArgument = arguments.stream().map(ArgumentResolution::of).collect(Collectors.toList());
```
nils-werner-sonarsource, line 85: I can imagine merging [SONARIAC-647](https://sonarsource.atlassian.net/browse/SONARIAC-647) could bring some value here. When the resolution contains the argument we don't have to resolve this every time.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public static Builder builder() {
+      return new Builder();
+    }
+
+    /**
+     * Return all block of arguments which match the list of predicates.
+     */
+    private List<List<Argument>> search(List<Argument> arguments) {
```
nils-werner-sonarsource, line 84: Instead of returning a list of lists, you could return a `Command` class that implements HasTextRange and only stores the arguments. So when calling `Command.textRange()` you can compute the range.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+      List<ArgumentResolution> resolvedArgument = arguments.stream().map(ArgumentResolution::of).collect(Collectors.toList());
+      List<List<Argument>> argumentBlockIssue = new ArrayList<>();
```
nils-werner-sonarsource, line 86: And this could be called `commands` when using a data class for the list of arguments per command argument block.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+      int argIndex = 0;
+      while (argIndex < arguments.size()) {
+        Integer sizeMatch = fullMatch(resolvedArgument, argIndex);
+        if (sizeMatch != null) {
+          argumentBlockIssue.add(arguments.subList(argIndex, argIndex + sizeMatch));
+          argIndex += sizeMatch;
+        } else {
+          argIndex++;
+        }
+      }
```
nils-werner-sonarsource, line 97: A while loop makes sense when having an iterator or queue. Here you are building a for loop with a while loop.
```
for (int argIndex = 0; argIndex < arguments.size(); argIndex++) {
        Integer sizeMatch = fullMatch(resolvedArgument, argIndex);
        if (sizeMatch != null) {
          argumentBlockIssue.add(arguments.subList(argIndex, argIndex + sizeMatch));
          argIndex += sizeMatch - 1;
        }
      }
```
IMO this loop is doing the same.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+    private List<List<Argument>> search(List<Argument> arguments) {
+      List<ArgumentResolution> resolvedArgument = arguments.stream().map(ArgumentResolution::of).collect(Collectors.toList());
+      List<List<Argument>> argumentBlockIssue = new ArrayList<>();
+
+      int argIndex = 0;
+      while (argIndex < arguments.size()) {
+        Integer sizeMatch = fullMatch(resolvedArgument, argIndex);
+        if (sizeMatch != null) {
+          argumentBlockIssue.add(arguments.subList(argIndex, argIndex + sizeMatch));
+          argIndex += sizeMatch;
```
nils-werner-sonarsource, line 93: This is not tested enough. When e.g. writing `argIndex += sizeMatch + 2;`, all tests are still passing.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+      return argumentBlockIssue;
+    }
+
+    /**
+     * If the provided list of resolved argument match with the list of predicates, return the size of matched predicates, ignoring optional predicates that didn't match.
+     * Otherwise, it will return null.
+     */
+    @CheckForNull
+    private Integer fullMatch(List<ArgumentResolution> resolvedArgument, int argIndex) {
+      int nbMatched = 0;
```
nils-werner-sonarsource, line 107: What does `nb` means?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+    sensitiveOpenSslSubcommand.search(runInstruction.arguments())
+      .forEach(sensitiveOpensslCall -> ctx.reportIssue(TextRanges.mergeRanges(sensitiveOpensslCall), MESSAGE));
+    sensitiveOpenSslDgst.search(runInstruction.arguments())
+      .forEach(sensitiveOpensslCall -> ctx.reportIssue(TextRanges.mergeRanges(sensitiveOpensslCall), MESSAGE));
+  }
+
+  private static class CommandDetector {
+
+    List<Predicate<String>> predicates;
+    List<Boolean> isOptionalPredicates;
```
nils-werner-sonarsource, line 70: Instead of having two lists just implement a class that extends the predicate which has a field that indicates, if the predicate is optional.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,151 @@
...
+          return null;
+        }
+      }
+      return nbMatched;
+    }
+
+    public static class Builder {
+
+      List<Predicate<String>> predicates = new ArrayList<>();
+      List<Boolean> isOptionalPredicates = new ArrayList<>();
```
nils-werner-sonarsource, line 129: See my comment above about the two strongly connected lists.

nils-werner-sonarsource (changes requested):
I like your approach. I was also drafting something similar. I've added a view comment which will harden the implementation. I am happy to talk about the comment.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
rudy-regazzoni-sonarsource: Yes :D I would like to move forward on this, it would really simplify things.
I did understand you already have the code done for this ? Otherwise I can also create the PR, let me know.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
rudy-regazzoni-sonarsource: That's exactly what I did first :D but then Sonar raise a major code smell on the `argIndex += sizeMatch - 1` with rule java:S127, as we are modifying the loop counter in the loop body.
I prefer the for loop indeed, should we ignore the code smell ? Or adapt differently ?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
nils-werner-sonarsource: https://github.com/SonarSource/sonar-iac/pull/613

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
rudy-regazzoni-sonarsource: Old french habbit ^^' Mean something like 'nombre', which could translate to 'quantity', I will change to a better name.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,164 @@
...
+
+  private static final String MESSAGE = "Using weak hashing algorithms is security-sensitive.";
+  private static final Set<String> OPENSSL_SENSITIVE_SUBCOMMAND = Set.of("md5", "sha1", "rmd160", "ripemd160");
+  private static final Set<String> OPENSSL_SENSITIVE_DGST_OPTION  = Set.of("-md2", "-md4", "-md5", "-sha1", "-ripemd160", "-ripemd", "-rmd160");
+
+  private static final CommandDetector sensitiveOpenSslSubcommand = CommandDetector.builder()
+    .with("openssl"::equals)
+    .with(OPENSSL_SENSITIVE_SUBCOMMAND::contains)
+    .build();
+  private static final CommandDetector sensitiveOpenSslDgst = CommandDetector.builder()
```
nils-werner-sonarsource, line 49: Constants are SCREAM_CASE

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
```diff
@@ -0,0 +1,164 @@
...
+     * Return all block of arguments which match the list of predicates.
+     */
+    private List<Command> search(List<ArgumentResolution> resolvedArguments) {
+      List<Command> commands = new ArrayList<>();
+
+      for (int argIndex = 0; argIndex < resolvedArguments.size(); argIndex++) {
+        Integer sizeCommand = fullMatch(resolvedArguments, argIndex);
+        if (sizeCommand != null) {
+          commands.add(new Command(resolvedArguments.subList(argIndex, argIndex + sizeCommand)));
+          argIndex += sizeCommand - 1;
```
nils-werner-sonarsource, line 103: This is still not covered by a test.
`argIndex += sizeCommand - 2;` does not lead to any failing test.

nils-werner-sonarsource (approved):
LGTM! You miss a comment to cover the increment of a the loop index in `WeakHashAlgorithmsCheck.CommandDetector#search`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WeakHashAlgorithmsCheck.java
rudy-regazzoni-sonarsource: Good point. New tests are covering the previous modification which was something like `argIndex += sizeCommand + 2;`, but not this modification with `- 2`.
This one is a bit more tricky as `argIndex += sizeCommand - 1;` is not required to make it work, it's more like an optimization to skip the full matched command. We would need a sensitive command which give us a sizeCommand of 1 to ensure this, which we don't have for now with OpenSSL (2+ arguments).

I created some specific test for CommandDetector class, for now it is in the class WeakHashAlgorithmsCheckTest, but once we extract CommandDetector, we should create a dedicated test class for it and move those test in it. I added a todo about this.

PR #612, opened 2023-03-16
SONARIAC-654 S6317 [Terraform] Update message and supports secondary …

navendu-barua-sonarsource:
…locations

### DIFF
--- a/iac-common/src/main/java/org/sonar/iac/common/checks/PrivilegeEscalationVector.java
+++ b/iac-common/src/main/java/org/sonar/iac/common/checks/PrivilegeEscalationVector.java
@@ -23,8 +23,11 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.sonar.iac.common.api.tree.Tree;
+import org.sonar.iac.common.checks.policy.Policy;
 
 public enum PrivilegeEscalationVector {
 
@@ -50,6 +53,8 @@ public enum PrivilegeEscalationVector {
   GLUE_DEVELOPMENT_ENDPOINT("Glue Development Endpoint", List.of("iam:PassRole", "glue:CreateDevEndpoint")),
   UPDATE_GLUE_DEV_ENDPOINT("Update Glue Dev Endpoint", List.of("glue:UpdateDevEndpoint")),
   UPDATE_LAMBDA_CODE("Update Lambda code", List.of("lambda:UpdateFunctionCode"));
+
+  private static final Pattern RESOURCE_NAME_PATTERN = Pattern.compile("arn:[^:]*:[^:]*:[^:]*:[^:]*:(role|user|group)/\\*");
 
   private final List<Permission.SimplePermission> permissions;
   private final String name;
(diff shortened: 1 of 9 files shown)


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/PrivilegeEscalationCheck.java
```diff
@@ -19,60 +19,79 @@
...
 
-  private static boolean isAllowEffect(Tree effect) {
-    return TextUtils.isValue(effect, "Allow").isTrue();
+      Optional<PrivilegeEscalationVector> vectorOpt = PrivilegeEscalationVector.getStatementEscalationVector(statement, actionTrees);
+      if (vectorOpt.isPresent()) {
+        PrivilegeEscalationVector vector = vectorOpt.get();
+        String vectorName = vector.getName();
+        List<SecondaryLocation> secondaryLocations = secondaryLocations(statement, vector, vectorName);
+        resourceSymbol.report(String.format(MESSAGE, vectorName), secondaryLocations);
+      }
```
karim-ouerghemmi-sonarsource, line 69: What do you think about moving line `63` to `69` in the above `if`. In general, I believe it only makes sense to look for a vector if the above condition is true `action.isPresent() && action.get() instanceof TupleTree`, and then we can avoid declaring `actionTrees` outside to an `emptyList`

iac-extensions/cloudformation/src/main/java/org/sonar/iac/cloudformation/checks/PrivilegeEscalationCheck.java
```diff
@@ -69,48 +81,23 @@ private static List<SecondaryLocation> secondaryLocations(Statement statement, P
     List<SecondaryLocation> secondaryLocations = new ArrayList<>();
     String actionsMsg = vector.getPermissions().size() == 1 ? String.format(MESSAGE_ACTION_SINGLE, vectorName) :
       String.format(MESSAGE_ACTION_MULTIPLE, vectorName);
+    addSecondaryLocationsFromAction(statement, vector, vectorName, secondaryLocations, actionsMsg);
+    addSecondaryLocationsFromResource(statement, secondaryLocations);
+    return secondaryLocations;
```
karim-ouerghemmi-sonarsource, line 86: See comment on this method in the Terraform check. I suggest doing the same changes here too.

iac-extensions/cloudformation/src/main/java/org/sonar/iac/cloudformation/checks/PrivilegeEscalationCheck.java
```diff
@@ -55,7 +58,16 @@ protected void checkResource(CheckContext ctx, Resource resource) {
 
   private static void checkPrivilegeEscalation(CheckContext ctx, Policy policy, Resource resource) {
     for (Statement statement : policy.statement()) {
-      Optional<PrivilegeEscalationVector> vectorOpt = getStatementEscalationVector(statement);
+      List<Tree> actionTrees = Collections.emptyList();
```
karim-ouerghemmi-sonarsource, line 61: See the suggestion on this method in the Terraform check. If you decide to go for it there, I suggest doing it here too.

iac-common/src/test/java/org/sonar/iac/common/checks/PrivilegeEscalationVectorTest.java
```diff
@@ -74,4 +83,96 @@ void all_vectors_have_a_non_empty_name_and_permissions_list() {
...
+  }
+
+  private static Policy.Statement statement(Map<String, String> properties) {
+    Tree[] trees = properties.entrySet().stream()
+      .map(e -> new PolicyTest.TestPropertyTree(e.getKey(), text(e.getValue())))
+      .toArray(Tree[]::new);
+    Tree statementTree = new PolicyTest.TestTree(trees);
+    Tree tree = new PolicyTest.TestTree(statementTree);
+    Policy policy = new Policy(tree, Tree::children);
+    return policy.statement().get(0);
```
karim-ouerghemmi-sonarsource, line 176:
```suggestion
  private static Policy.Statement statement(Map<String, String> properties) {
    PropertyTree[] propertyTrees = properties.entrySet().stream()
      .map(e -> attribute(e.getKey(), text(e.getValue())))
      .toArray(PropertyTree[]::new);
    TestPropertiesTree statement = properties(propertyTrees);
    return new Policy(null, tree -> List.of(statement)).statement().get(0);
```
Seeing this again, and the reported Code Smell, I believe we can simplify it as the suggestion. I found that it is done in a simpler way in `policy.PolicyValidatorTest` that uses common test helpers.

iac-common/src/main/java/org/sonar/iac/common/checks/policy/Policy.java
```diff
@@ -61,7 +61,7 @@ public static class Statement {
     private final Tree notResource;
     private final Tree condition;
 
-    private Statement(Tree statement) {
+    public Statement(Tree statement) {
```
karim-ouerghemmi-sonarsource, line 64:
```suggestion
    private Statement(Tree statement) {
```
I think, we can revert this.

iac-common/src/main/java/org/sonar/iac/common/checks/PrivilegeEscalationVector.java
```diff
@@ -71,16 +78,49 @@ public static boolean isSupersetOfAnEscalationVector(Stream<String> actionPermis
...
+      && statement.resource().filter(PrivilegeEscalationVector::isSensitiveResource).isPresent()
+      && statement.condition().isEmpty()
+      && statement.principal().isEmpty()
+      && statement.action().isPresent()) {
+      return getActionEscalationVector(actionTrees);
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<PrivilegeEscalationVector> getActionEscalationVector(List<Tree> testTree) {
```
karim-ouerghemmi-sonarsource, line 102:
```suggestion
  private static Optional<PrivilegeEscalationVector> getActionEscalationVector(List<Tree> actionTrees) {
```

karim-ouerghemmi-sonarsource (changes requested):
Functionality LGTM. Nice!
Added some refactoring suggestions. If something is unclear, let's have a call :slightly_smiling_face:

iac-common/src/test/java/org/sonar/iac/common/checks/policy/PolicyTest.java
karim-ouerghemmi-sonarsource, line 1: Visibility changes introduced with this PR in this file can be reverted as we do not need them anymore.

karim-ouerghemmi-sonarsource (approved):
LGTM. Nice!
See comment about reverting changes in `PolicyTest.java`.

PR #617, opened 2023-03-17
SONARIAC-656 Using weak hashing algorithms with shasum is security-sensitive

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM! Please see my comment

PR #618, opened 2023-03-17
SONARIAC-664 Make CommandDetector available as a utils class to be used by other checks

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM !

PR #620, opened 2023-03-20
Adapt CommandDetector with stack-based approach

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM ! Nice improvement of the Command API, I like that we don't have anymore the indexes to play with.

PR #621, opened 2023-03-20
SONARIAC-672 Rule S6500: Use new CommandDetector to improve check implementation

nils-werner-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -84,6 +86,9 @@ private List<ArgumentResolution> fullMatch(Deque<ArgumentResolution> argumentSta
 
       // Stop argument detection when argument list is empty or argument is unresolved to start new command detection
       if (resolution == null || resolution.is(UNRESOLVED)) {
+        if (resolution == null && remainingPredictsAreOptional(currentPredicate, predicateStack)) {
+          return commandArguments;
+        }
```
nils-werner-sonarsource, line 91: This is not nice but I don't have any more lean solutions so far.

rudy-regazzoni-sonarsource: Question : why not `return commandArguments` if resolution is `UNRESOLVED` and `remainingPredictsAreOptional(...)` is `true` ?
This seems valid to me, and we could then write the condition simpler.

rudy-regazzoni-sonarsource (approved):
LGTM ! Please see my comment.

jonas-wielage-sonarsource (approved):
LGTM! Like the new syntactical sugar.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
rudy-regazzoni-sonarsource: Ok I see, indeed it would possibly make it noisy, I forgot that `remainingPredictsAreOptional` exclude `MATCH`, which mean it include `NO_MATCH`.

PR #632, opened 2023-03-21
SONARIAC-678 Having executables in COPY/ADD not owned by root is security-sensitive

rudy-regazzoni-sonarsource:


sonar-iac-plugin/pom.xml
```diff
@@ -96,7 +96,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>1500000</maxsize>
+                  <maxsize>1600000</maxsize>
                   <minsize>950000</minsize>
```
rudy-regazzoni-sonarsource, line 100: We reached the limit, I incremented a bit.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ExecutableNotOwnedByRootCheck.java
```diff
@@ -0,0 +1,127 @@
...
+      ctx.reportIssue(primaryLocation, MESSAGE, secondaryLocations);
+    }
+  }
+
+  @CheckForNull
+  private static Flag getSensitiveChownFlag(TransferInstruction transferInstruction) {
+    return transferInstruction.options().stream()
+      .filter(f -> f.name().equals("chown"))
+      .filter(ExecutableNotOwnedByRootCheck::isSensitiveUser)
+      .findFirst().orElse(null);
```
mstachniuk, line 89: `.orElse(null)` should be probably in a new line - it will break master after merge if not fixed

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ExecutableNotOwnedByRootCheck.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  private static boolean isSensitiveUser(Flag chownFlag) {
```
mstachniuk, line 92: There are different method names patterns: `isSensitiveUser` vs. `isFileSensitive`, `isChmodWriteSensitive`, etc.
It will be nice to have one common convention. e.g. `Sensitive` is always at the end of the name

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ExecutableNotOwnedByRootCheck.java
```diff
@@ -0,0 +1,127 @@
...
+    ArgumentResolution resolvedArgArgument = ArgumentResolution.of(chownFlag.value());
+    return resolvedArgArgument.is(ArgumentResolution.Status.RESOLVED) && !"root".equals(resolvedArgArgument.value());
+  }
+
+  @CheckForNull
+  private static Chmod getChmod(TransferInstruction transferInstruction) {
+    return transferInstruction.options().stream()
+      .filter(f -> f.name().equals("chmod"))
+      .map(f -> ArgumentResolution.of(f.value()))
+      .filter(argResolved -> argResolved.is(ArgumentResolution.Status.RESOLVED))
```
mstachniuk, line 102: The call `argResolved.is(ArgumentResolution.Status.RESOLVED)` looks a little bit strange because the internal implementation detail is exposed outside the class. 
Consider adding methods `ArgumentResolution.isResolved()`, `ArgumentResolution.isUnresolved()` & `ArgumentResolution.isEmpty()` - then client code will be much better

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/Chmod.java
```diff
@@ -50,6 +50,12 @@ public Chmod(Argument chmodArg, Argument permissionsArg, String permissions) {
     this.permissions = parsePermissions(permissions);
   }
 
+  public Chmod(String permissions) {
+    this.chmodArg = null;
```
mstachniuk, line 54: It's better to call `this(null, null, permissions)` - then the construction logic is one place

iac-extensions/docker/src/test/resources/checks/ExecutableNotOwnedByRootCheck/add_copy.dockerfile
```diff
@@ -0,0 +1,99 @@
...
+ADD --chown=other              file.txt target/
+ADD --chown=other --chmod=u+x  file.sh  target/
+ADD --chown=other --chmod=u+x  file.txt target/
+# Noncompliant@+1
+ADD --chown=other --chmod=u+w  file.sh  target/
+ADD --chown=other --chmod=u+w  file.txt target/
+# Noncompliant@+1
+ADD --chown=other --chmod=u+wx file.sh  target/
+# Noncompliant@+1
+ADD --chown=other --chmod=u+wx file.txt target/
```
mstachniuk, line 30: I like the column formatting

mstachniuk (approved):
LGTM and please see my minor suggestions

iac-extensions/docker/src/test/resources/checks/ExecutableNotOwnedByRootCheck/add_copy.dockerfile
rudy-regazzoni-sonarsource: There are so much different use cases and the logic to differentiate compliant/sensitive is not straightforward.
There was a real need to be able to compare those cases to better understand which ones are sensitive and which ones are compliant.
Happy that you liked it :D

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ExecutableNotOwnedByRootCheck.java
nils-werner-sonarsource: Yeah, sounds good to me.

PR #637, opened 2023-03-23
SONARIAC-681 Rule S6437: Add secret generation detection for ssh-keygen

jonas-wielage-sonarsource:


jonas-wielage-sonarsource:
I just wanted to have some feedback for this PR.
Code Smells and coverage are something I will fix beforehand.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -78,44 +82,149 @@ public List<Command> search(List<ArgumentResolution> resolvedArguments) {
...
   private List<ArgumentResolution> fullMatch(Deque<ArgumentResolution> argumentStack) {
-    List<ArgumentResolution> commandArguments = new ArrayList<>();
-    Deque<CommandPredicate> predicateStack = new LinkedList<>(predicates);
-    while (!predicateStack.isEmpty()) {
-      CommandPredicate currentPredicate = predicateStack.pollFirst();
-      ArgumentResolution resolution = argumentStack.pollFirst();
+    CommandDetectionStatus detectionStatus = new CommandDetectionStatus(argumentStack, new LinkedList<>(predicates), new ArrayList<>());
+    while (!detectionStatus.predicateStack.isEmpty()) {
+
+      detectionStatus.setCurrentPredicate(detectionStatus.predicateStack.pollFirst());
```
nils-werner-sonarsource, line 90: This method does not need a parameter as the only provided argument is of the object itself

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -78,44 +82,149 @@ public List<Command> search(List<ArgumentResolution> resolvedArguments) {
...
   // Cognitive Complexity of methods should not be too high
   @SuppressWarnings("java:S3776")
+  // TODO: Rework commentary
   private List<ArgumentResolution> fullMatch(Deque<ArgumentResolution> argumentStack) {
-    List<ArgumentResolution> commandArguments = new ArrayList<>();
-    Deque<CommandPredicate> predicateStack = new LinkedList<>(predicates);
-    while (!predicateStack.isEmpty()) {
-      CommandPredicate currentPredicate = predicateStack.pollFirst();
-      ArgumentResolution resolution = argumentStack.pollFirst();
+    CommandDetectionStatus detectionStatus = new CommandDetectionStatus(argumentStack, new LinkedList<>(predicates), new ArrayList<>());
```
nils-werner-sonarsource, line 87: As the constructor is only used once, providing the stack and the predicates should be enough. The empty array as well as the List to LinkedList conversion can be handled in the constructor itself

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -78,44 +82,149 @@ public List<Command> search(List<ArgumentResolution> resolvedArguments) {
...
+    while (!detectionStatus.predicateStack.isEmpty()) {
+
+      detectionStatus.setCurrentPredicate(detectionStatus.predicateStack.pollFirst());
+      // resolution is removed from stack during match-methods, here it is peeked to see if it is null or UNRESOLVED
+      ArgumentResolution resolution = detectionStatus.argumentStack.peekFirst();
```
nils-werner-sonarsource, line 92: Implement a method instead of a field access

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -78,44 +82,149 @@ public List<Command> search(List<ArgumentResolution> resolvedArguments) {
...
 
       // Stop argument detection when argument list is empty
       if (resolution == null) {
-        return remainingPredictsAreOptional(currentPredicate, predicateStack) ? commandArguments : Collections.emptyList();
+        return remainingPredictsAreOptional(detectionStatus.getCurrentPredicate(), detectionStatus.predicateStack) ? detectionStatus.commandArguments : Collections.emptyList();
```
nils-werner-sonarsource, line 96: Do not mix getters and direct field access. I would recommend a getter setup.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -78,44 +82,149 @@ public List<Command> search(List<ArgumentResolution> resolvedArguments) {
...
-        // Re-add predicate to stack to be reevaluated on the next argument
-        if (currentPredicate.is(ZERO_OR_MORE)) {
-          predicateStack.addFirst(currentPredicate);
-        }
-        // Add matched argument to command
-        commandArguments.add(resolution);
-      } else if (currentPredicate.is(OPTIONAL, ZERO_OR_MORE, NO_MATCH)) {
-        // Re-add argument to be evaluated by the next predicate
-        argumentStack.addFirst(resolution);
+      if (detectionStatus.getCurrentPredicate() instanceof SingularPredicate) {
```
nils-werner-sonarsource, line 105: As you call the methods 7 times inside the loop maybe store its value in a variable as it is not expected to return different values withing a single loop iteration.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -78,44 +82,149 @@ public List<Command> search(List<ArgumentResolution> resolvedArguments) {
...
+      // above method should return empty
+      detectionStatus.setStatus(CommandDetectionStatus.Status.FOUND_NOTHING);
+      return;
+    }
+    detectionStatus.setStatus(CommandDetectionStatus.Status.CONTINUE);
+  }
+
+  private void matchPredicate(OptionPredicate optionPredicate, CommandDetectionStatus detectionStatus) {
+    matchPredicate(optionPredicate.flagPredicate, detectionStatus);
+    if (detectionStatus.is(CommandDetectionStatus.Status.ABORT, CommandDetectionStatus.Status.FOUND_NOTHING) || optionPredicate.withoutValue()) {
```
nils-werner-sonarsource, line 161: I would replace `optionPredicate.withoutValue()` with `optionPredicate.valuePredicate == null` as this helps analyzers to see that 166 is not null

nils-werner-sonarsource (commented):
A lot of changes! Nice approach.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandPredicate.java
```diff
@@ -0,0 +1,32 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.checks.utils;
+
+public interface CommandPredicate {
+
+  enum Type {
+    MATCH,
+    NO_MATCH,
+    OPTIONAL,
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: In the end I did not remove `OPTIONAL` : at the time I wanted to merge the PR which was supposed to remove it, @jonas-wielage-sonarsource introduced a PR for `ClearTextProtocolDowngradeCheck` which was using `withOptional()` a lot, so I kept it.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/CommandMatcher.java
```diff
@@ -0,0 +1,287 @@
...
+      }
+
+      // choose match-method depending on implementation of currentPredicate
+      if (currentPredicate instanceof SingularPredicate) {
+        matchPredicate((SingularPredicate) currentPredicate);
+      } else if (currentPredicate instanceof OptionPredicate) {
+        matchPredicate((OptionPredicate) currentPredicate);
+      } else if (currentPredicate instanceof MultipleUnorderedOptionsPredicate) {
+        matchPredicate((MultipleUnorderedOptionsPredicate) currentPredicate);
+      }
```
rudy-regazzoni-sonarsource, line 92: Seems a bit strange to me, I understand that we need a different logic to be applied depending of the predicate object, but this goes a bit against the OOP concept.

It would be great to be able to extract the specific logics into the related predicate class.

My default reflex would be to have the interface `CommandPredicate` define a method `match(...)` from which each inheriting class `SingularPredicate`/`OptionPredicate`/`MultipleUnorderedOptionsPredicate` would have an implementation. This implementation would be the current code you have in the different `matchPredicate()` method.
But this cannot be that simple since those depends heavily depend of the `CommandMatcher` context.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/CommandMatcher.java
```diff
@@ -0,0 +1,287 @@
...
+    this.status = status;
+  }
+
+  private ArgumentResolution pollFirstArgumentResolution() {
+    return argumentStack.pollFirst();
+  }
+
+  private ArgumentResolution peekFirstArgumentResolution() {
+    return argumentStack.peekFirst();
+  }
```
rudy-regazzoni-sonarsource, line 281: Not totally convinced that those method are necessary ^^ I don't have a strong opinion on this, feel free to keep them if you think they bring clarity.

rudy-regazzoni-sonarsource (commented):
There is a lot of work there !
We are moving toward what we wanted to avoid : having a really complex `CommandDetector` class ^^' The previous version indeed had strong limitation which were preventing from implementing this ticket properly.
Still because of our current approach with `RUN`/`CMD`, it is hard to parse command without risking misinterpreting.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/CommandMatcher.java
nils-werner-sonarsource: This was my suggestion, as it might be hard to understand what peekFirst means for argument handling. And I agree that this naming does not bring any big advantages as it duplicates the `argumentStack.peekFirst` wording.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/CommandMatcher.java
nils-werner-sonarsource: Sounds like a good and clean approach.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/CommandMatcher.java
jonas-wielage-sonarsource: Yeah, looking over it again, I'm also not convinced about those methods. Because we are using the argumentStack now as an argument in the CommandMatcher class, we can drop the extra method call. Or are there other suggestions?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/CommandMatcher.java
jonas-wielage-sonarsource: How would you suggest handling the context of the `CommandMatcher`? The original idea was to extract the methods and arguments to the `CommandMatcher` to not pass 4 arguments in each method call.

jonas-wielage-sonarsource: One option would be to pass an object to the methods, which wraps the other arguments and the status.
(Essentially an object of the class `CommandMatcher`, which doesn't have the `match()` methods and only works as a wrapper.
But that would be pretty similar to my first approach, where everything was handled in the CommandDetector, and I had an internal class wrapping the arguments. Except that now the `match()` methods would be contained in the implementations of `CommandPredicate` (which I think is a step in the right direction, but not quite ideal)

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/openssl.dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu:22.04
 
 # Noncompliant@+1 {{Revoke and change this secret, as it might be compromised.}}
 RUN openssl genrsa -des3 -passout pass:x -out server.pass.key 2048
-#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
rudy-regazzoni-sonarsource, line 5: I would add below test with highlighting :
```
RUN openssl genrsa -des3 -passout pass:x -out server.pass.key 2048 && other command
RUN openssl genrsa -des3 -passout pass:x -out server.pass.key 2048 | other command
RUN openssl genrsa -des3 -passout pass:x -out server.pass.key 2048; other command
```

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/sshKeygen.dockerfile
```diff
@@ -0,0 +1,109 @@
...
+# Noncompliant@+1
+RUN ssh-keygen -N "" -f rsync-key -b 1024 -t dsa
+
+# Noncompliant@+1
+RUN ssh-keygen -N "" -b 1024 -t dsa -f rsync-key
+
+# Noncompliant@+1
+RUN ssh-keygen -N "" -b 1024 -f rsync-key -t dsa
+
+RUN ssh-keygen -N -t dsa -b 1024 -f rsync-key
```
rudy-regazzoni-sonarsource, line 93: I would add these test case :
```
RUN ssh-keygen -N ''      -t dsa -b 1024 -f rsync-key
RUN ssh-keygen -N value   -t dsa -b 1024 -f rsync-key
RUN ssh-keygen -N "value" -t dsa -b 1024 -f rsync-key
RUN ssh-keygen -N ""value -t dsa -b 1024 -f rsync-key

ARG EMPTY=""
ARG NOT_EMPTY="value"
RUN ssh-keygen -N $EMPTY     -t dsa -b 1024 -f rsync-key
RUN ssh-keygen -N $NOT_EMPTY -t dsa -b 1024 -f rsync-key
RUN ssh-keygen -N $UNKNOWN   -t dsa -b 1024 -f rsync-key
```

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/sshKeygen.dockerfile
```diff
@@ -0,0 +1,109 @@
...
+RUN ssh-keygen -random1 -N "" -random2 -t dsa -random3 -b 1024 -random4 -f rsync-key -random5
+
+# Noncompliant@+1
+RUN ssh-keygen -random1 random1Value -N "" -random2 -t dsa -random3 =random3Value -b 1024 -random4 -f rsync-key -random5
+
+# Noncompliant@+1
+RUN ssh-keygen -random1 random1Value -N "" -random2 -t dsa -random3 =random3Value -b 1024 -random4 -f rsync-key -random5 random5Option
+
+
+# Noncompliant@+1
```
rudy-regazzoni-sonarsource, line 17: I would add this test case (not necessarily here) :
`RUN ssh-keygen -N "" -t dsa -f rsync-key -b 1024 $UNRESOLVED`

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/sshKeygen.dockerfile
```diff
@@ -0,0 +1,109 @@
+FROM ubuntu:22.04
+
+# Noncompliant@+1 {{Revoke and change this secret, as it might be compromised.}}
+RUN ssh-keygen -N "" -t dsa -b 1024 -f rsync-key
+#   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
rudy-regazzoni-sonarsource, line 6: I would add this test case (not necessarily here) :
`RUN ssh-keygen -N "" -t dsa -N "val" -b 1024 -f rsync-key`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/MultipleUnorderedOptionsPredicate.java
```diff
@@ -0,0 +1,169 @@
...
+    OptionPredicate anyOptionExceptExpectedPredicate = this.calculateAnyOptionMatchingExceptExpected();
+
+    // initial initialization to step into while, could be solved with do-while
+    boolean anythingMatched = true;
+
+    while (anythingMatched && !workingSet.isEmpty()) {
+
+      if (context.areNoArgumentsToHandle()) {
+        if (!context.remainingPredicatesAreOptional()) {
+          context.setStatus(ABORT);
```
rudy-regazzoni-sonarsource, line 97: The code should have more test : if I comment this line, I don't get any failed test.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/MultipleUnorderedOptionsPredicate.java
```diff
@@ -0,0 +1,169 @@
...
+
+    }
+
+    // calculate success or failure due to size of sets
+    if (fulfilledOptions.size() != expectedMatches) {
+      // not all expectedMatches are fulfilled so overall the MultipleUnorderedOptionsPredicate hasn't matched
+      context.setStatus(ABORT);
+      return;
+    }
+    context.setStatus(PredicateContext.Status.CONTINUE);
```
rudy-regazzoni-sonarsource, line 117: Same here.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/MultipleUnorderedOptionsPredicate.java
```diff
@@ -0,0 +1,169 @@
...
+          context.setStatus(ABORT);
+        }
+        return;
+      }
+
+      anythingMatched = matchExpectedOrAnyOption(context, fulfilledOptions, workingSet, anyOptionExceptExpectedPredicate);
+      if (context.is(ABORT)) {
+        // could also check for FOUND_NO_PREDICATE_MATCH, but renders the boolean anythingMatched useless which is added for better readability of
+        // the control flow
+        return;
```
rudy-regazzoni-sonarsource, line 106: Same here.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/MultipleUnorderedOptionsPredicate.java
```diff
@@ -0,0 +1,169 @@
...
+      }
+
+    }
+
+    // calculate success or failure due to size of sets
+    if (fulfilledOptions.size() != expectedMatches) {
+      // not all expectedMatches are fulfilled so overall the MultipleUnorderedOptionsPredicate hasn't matched
+      context.setStatus(ABORT);
+      return;
+    }
```
rudy-regazzoni-sonarsource, line 116: Same here: even If I comment this whole block, I don't get any failed test.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/OptionPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+      if (context.getCurrentPredicate() instanceof MultipleUnorderedOptionsPredicate && this.valuePredicate.hasType(MATCH)) {
+        context.setStatus(ABORT);
+      }
+      return;
+    }
+    this.valuePredicate.match(context);
+    if (context.is(FOUND_NO_PREDICATE_MATCH)) {
+      // should abort matching process, because the flag matched but the value didn't match, and there shouldn't be two flags that are identical
+      context.setStatus(ABORT);
+    }
```
rudy-regazzoni-sonarsource, line 78: Same here: If I comment this, no failed tests.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/OptionPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+      // further action is evaluated by the calling method
+      return;
+    }
+
+    // should not match on empty argumentStack
+    if (context.areNoArgumentsToHandle()) {
+      // in this special case the calling method doesn't handle it properly, so it should be aborted here
+      if (context.getCurrentPredicate() instanceof MultipleUnorderedOptionsPredicate && this.valuePredicate.hasType(MATCH)) {
+        context.setStatus(ABORT);
+      }
```
rudy-regazzoni-sonarsource, line 71: Same comment as for the previous file: the code should have more test: if I comment this line, I don't get any failed test.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/SingularPredicate.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public void match(PredicateContext context) {
+    ArgumentResolution resolution = context.getNextArgumentToHandleAndRemoveFromList();
+
+    if (resolution.isUnresolved()) {
+      context.setStatus(ABORT);
+      return;
+    }
```
rudy-regazzoni-sonarsource, line 61: Same as before: I can remove this block of code without breaking any test.

rudy-regazzoni-sonarsource (changes requested):
Nice improvement ! The split of the logic seems much better.
I pursued the review more in depth in test code. Either it need more test or it can be simplified a lot: we can remove several instructions and small block of code without breaking any test :D
Please see my comments, I'm also available if you want to discuss this directly.

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/openssl.dockerfile
jonas-wielage-sonarsource: In this case the other command would also be flagged.

rudy-regazzoni-sonarsource (approved):
LGTM ! Nice improvement of the `CommandDetector` API

PR #639, opened 2023-03-24
SONARIAC-682 Rule S6437: Add secret generation detection for keytool & openssl

jonas-wielage-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -0,0 +1,70 @@
...
+    .withOptionalRepeatingExcept(SENSITIVE_KEYTOOL_FLAGS)
+    .build();
+
+  private static final Set<String> SENSITIVE_OPENSSL_SUBCOMMANDS = Set.of("req", "genrsa", "rsa", "gendsa", "ec", "ecparam", "x509", "genpkey", "pkey");
+
+  private static final CommandDetector SENSITIVE_OPENSSL_COMMANDS = CommandDetector.builder()
+    .with("openssl")
+    .with(SENSITIVE_OPENSSL_SUBCOMMANDS)
+    // everything that comes after a MATCH of the sensitive subcommand should be flagged as well
+    .withOptionalRepeating(s -> true)
```
rudy-regazzoni-sonarsource, line 53: I understand the why, but by doing so you will highlight the full RUN instruction, even if it contain multiple command as below :
```
RUN openssl genrsa -des3 && other command && other
```
```
RUN <<EOF
  openssl genrsa -des3
  apt install wget
  wget http://myurl
EOF
```

It is probably less noisy to just highlight the `openssl` call with the subcommand.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -0,0 +1,70 @@
...
+  private static final String MESSAGE = "Revoke and change this secret, as it might be compromised.";
+
+  private static final Set<String> SENSITIVE_KEYTOOL_FLAGS = Set.of("-gencert", "-genkeypair", "-genseckey", "-genkey");
+
+  private static final CommandDetector KEYTOOL_DETECTOR = CommandDetector.builder()
+    .with("keytool")
+    .withOptionalRepeatingExcept(SENSITIVE_KEYTOOL_FLAGS)
+    .with(SENSITIVE_KEYTOOL_FLAGS)
+    .withOptionalRepeatingExcept(SENSITIVE_KEYTOOL_FLAGS)
+    .build();
```
rudy-regazzoni-sonarsource, line 45: Same things that I mentioned on @mstachniuk PR, here we are potentially matching over several commands, which in some case may throw strange issue or even FP.
```
RUN keytool -random1 && unrelated_command -genkeypair 
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This is indeed an edgy case here, but I don't exclude that something like that can happen.
We can keep it as it is for now and see if we get FP report later from users, but we need to keep in mind such thing as we usually like conservative approach.
Please add this use case in test file to document it.

Still without a real bash parser, we cannot do much better.

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/keytool.dockerfile
rudy-regazzoni-sonarsource, line 1: Please add a use case with the `ExecForm` and the `HeredocForm`, possibly including the highlight :
```
RUN ["keytool", "-genseckey", ...]
```
```
RUN <<EOF
  keytool -genseckey ...
EOF
```

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/keytool.dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Noncompliant@+1
+RUN keytool -random1 -genkeypair -random2
+
+# Noncompliant@+1
+RUN keytool -random1 opt1 -gencert -random2
+
+# Noncompliant@+1
+RUN keytool -random1 opt1 -genkey -random2
+
```
rudy-regazzoni-sonarsource, line 15: I would add a case with another command before + add the highlighting :
`RUN apt install wget && keytool -genkey`

rudy-regazzoni-sonarsource (approved):
LGTM ! Nice work, please see my comments.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
jonas-wielage-sonarsource: I will change to only Match on flags. Could be handled better with the updates to the CommandDetector in the other PR, so i will leave it for not to highlight not everything

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
jonas-wielage-sonarsource: I will add a FP, but i think the new CommandDetector can handle this aswell

acdha:
This caused a regression in some of our Dockerfiles because the list of sensitive OpenSSL commands is too broad. In our case we were converting an internal CA's public key into a different format using a command like this which is now flagged when linting Dockerfiles:

```
openssl x509 -inform der -in /etc/ssl/certs/example.crt -outform pem -out /etc/ssl/certs/example.pem
```

For the `x509` subcommand, I think the security concern would be satisfied by checking whether either `-passin`/`password` or `-passout` are specified using the `pass:` option which specifies the value in the arguments or simply at all (perhaps that's a blocker for the former and a warning for the latter?).

jonas-wielage-sonarsource:
Hello @acdha,
Thanks for reaching out to us to report the false positives you are having.

Our preferred way to report such issues is via the community forum, which you can find here [Sonar Community](https://community.sonarsource.com/).

For false positives, there is also a specific category you can find here [False-Positive / False-Negative Category](https://community.sonarsource.com/c/clean-code/fp/7).

In any case, I will make sure to forward your issue to the responsible team.

Best,
Jonas

PR #640, opened 2023-03-27
SONARIAC-687 Extract common code from CfnLintImporter

mstachniuk:


iac-common/src/main/java/org/sonar/iac/common/reports/AbstractJsonReportImporter.java
```diff
@@ -0,0 +1,125 @@
...
+      int failedToSaveIssues = saveIssues(issuesJson);
+      if (failedToSaveIssues > 0) {
+        addWarning(reportFile.getPath(), issuesJson.size(), failedToSaveIssues);
+      }
+    });
+  }
+
+  protected abstract void saveIssue(JSONObject issueJson);
+
+  protected Optional<JSONArray> parseJson(File reportFile) {
```
jonas-wielage-sonarsource, line 71: Looks like a method the `JSONParser` should contain itself.

iac-common/src/main/java/org/sonar/iac/common/reports/AbstractJsonReportImporter.java
```diff
@@ -0,0 +1,125 @@
...
+      String message = String.format("could not parse file as JSON %s", reportFile.getPath());
+      logWarning(message);
+    } catch (RuntimeException e) {
+      String message = String.format("file is expected to contain a JSON array but didn't %s", reportFile.getPath());
+      logWarning(message);
+    }
+    return Optional.ofNullable(issuesJson);
+  }
+
+  protected int saveIssues(JSONArray issuesJson) {
```
jonas-wielage-sonarsource, line 88: I would order it under the other `saveIssues(...) `method

jonas-wielage-sonarsource (approved):
LGTM!
But I'm wondering why the code coverage is so low? Shouldn't there be an IT that uses this?

jonas-wielage-sonarsource:
And another question. Do you know why we don't use a JSON Parser framework like Jackson?

iac-common/src/main/java/org/sonar/iac/common/reports/AbstractJsonReportImporter.java
```diff
@@ -0,0 +1,125 @@
...
+public abstract class AbstractJsonReportImporter {
+  private static final Logger LOG = Loggers.get(AbstractJsonReportImporter.class);
+  private static final JSONParser jsonParser = new JSONParser();
+
+  protected final SensorContext context;
+  private final AnalysisWarningsWrapper analysisWarnings;
+  private final String warningPrefix;
+  private Set<String> unresolvedPaths;
+
+  protected AbstractJsonReportImporter(SensorContext context, AnalysisWarningsWrapper analysisWarnings, String warningPrefix) {
```
jonas-wielage-sonarsource, line 47: I'm not sure about this, as i don't have that much experience with our naming convention.
Shouldn't we drop the `Abstract` in the class name?

iac-common/src/main/java/org/sonar/iac/common/reports/AbstractJsonReportImporter.java
mstachniuk: The `JSONParser` contains already the `parse` method. Here in this method, an error handling is encapsulated, so the main flow in `importReport` is easier to read & understand.

iac-common/src/main/java/org/sonar/iac/common/reports/AbstractJsonReportImporter.java
mstachniuk: We don't have a naming convention for that and for me, it's ok to use `Abstract` in the class name.

jonas-wielage-sonarsource:
Thanks for the clarification about the coverage, makes sense that the coverage is only calculated module based.

PR #649, opened 2023-03-31
SONARIAC-689 Define rules for tflint (Terraform)

mstachniuk:


jonas-wielage-sonarsource (approved):
please review if we really need these rules, as they are not really used (i think)

PR #654, opened 2023-04-03
SONARIAC-692 Import issues from Hadolint report

jonas-wielage-sonarsource:


its/plugin/projects/hadolint_project/docker/hadolint-report_sonarQube.json
mstachniuk, line 1: I would suggest not using uppercase in the filename as there already `-` & `_` are used as separators.

iac-common/src/main/java/org/sonar/iac/common/reports/AbstractJsonReportImporter.java
```diff
@@ -93,6 +93,10 @@ protected List<JSONArray> parseJson(File reportFile) {
     return List.of(issuesJson);
   }
 
+  protected JSONArray parseFileAsArray(File reportFile) throws IOException, ParseException {
```
mstachniuk, line 96: It can be also adopted by `TFLintImporter`

mstachniuk (approved):
LGTM and please see my comments and please fix a bug reported by SonarCloud

PR #660, opened 2023-04-21
Set released sonarlint-core version for integration test

nils-werner-sonarsource:
As we set an unreleased version we now fail to build the plugin. After 2 weeks the unreleased artifact will be removed. This will not affect the released version of sonar-iac but to have nightly builds of sonar-iac we have to set the used version to a released one which will not be deleted.


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #662, opened 2023-05-03
SONARIAC-753 S6505 should not raise issue if `yarn` is used as command option

jonas-wielage-sonarsource:


nils-werner-sonarsource (commented):
What about my suggestion to adapt the detector to check an object before? To detect

* `RUN yarn`
* `RUN foo bar && yarn` or `RUN foo bar; yarn`

nils-werner-sonarsource (approved):
LGTM! Please add the ticket number, which will improve the CommandDetector to handle valid standalone `yarn` usage cases.

PR #671, opened 2023-05-05
SONARIAC-752 Log filename when file identifier does not match in YamlSensor

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/extension/IacSensor.java
```diff
@@ -115,6 +115,13 @@ public void execute(SensorContext sensorContext) {
     statistics.log();
   }
 
+  private List<InputFile> inputFiles(SensorContext sensorContext) {
+    FileSystem fileSystem = sensorContext.fileSystem();
+    FilePredicate predicate = mainFilePredicate(sensorContext);
+    return StreamSupport.stream(fileSystem.inputFiles(predicate).spliterator(), false)
+      .collect(Collectors.toList());
+  }
```
nils-werner-sonarsource, line 123: That way the predicates are only applied once. Before there were applied in lines 101 and 158.

iac-extensions/cloudformation/src/test/java/org/sonar/iac/cloudformation/plugin/CloudformationSensorTest.java
```diff
@@ -35,6 +37,8 @@
 
 class CloudformationSensorTest extends ExtensionSensorTest {
 
+  @RegisterExtension
+  LogTesterJUnit5 logTester = new LogTesterJUnit5();
```
jonas-wielage-sonarsource, line 41: Should not be needed, as there is already a logtester in a parent class

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/plugin/KubernetesSensorTest.java
```diff
@@ -66,6 +66,11 @@ void shouldNotParseYamlFileWithoutIdentifiers() {
   void shouldNotParseYamlFileWithIncompleteIdentifiers() {
     analyse(sensor(), inputFile("apiVersion: ~\nkind: ~\nmetadata: ~\n"));
     asserNotSourceFileIsParsed();
+
+    var logs = logTester.logs(LoggerLevel.DEBUG);
```
jonas-wielage-sonarsource, line 70: The logtester could also be retrieved from the parent class

jonas-wielage-sonarsource (approved):
LGTM, only small nitpicks

iac-common/src/main/java/org/sonar/iac/common/extension/IacSensor.java
jonas-wielage-sonarsource: I don't see any duplicate applications before, but looks good nonetheless!

PR #673, opened 2023-05-08
SONARIAC-782 Invalid line offset while issue reporting should not lead to analysis failure

nils-werner-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM, but could lead to issues with the reporting that we miss

PR #681, opened 2023-05-23
SONARIAC-830 Add Azure Resource Manager Extensions: Provide AzureResourceManagementSensor

rudy-regazzoni-sonarsource:


rudy-regazzoni-sonarsource:
The missing code coverage can't reliably be achieved here from my perspective: we don't have specific tests for settings class, and this code will be tested through some other classes once we have other extensions set and defined in our ArmExtension class.

iac-common/src/test/java/org/sonar/iac/common/extension/FileIdentificationPredicateTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+class FileIdentificationPredicateTest {
+
+  @RegisterExtension
+  public LogTesterJUnit5 logTester = new LogTesterJUnit5();
+
+  private static final String IDENTIFIER = "myidentifier";
+
+  @Test
+  void emptyIdentifierAlwaysTrue() {
```
mstachniuk, line 44: Thanks for those tests

iac-extensions/arm/src/main/java/org/sonar/iac/arm/plugin/ArmExtension.java
```diff
@@ -23,10 +23,19 @@
...
 public class ArmExtension {
 
+  public static final String REPOSITORY_KEY = "azureresourcemanager";
+
   private ArmExtension() {
   }
 
   public static void define(Plugin.Context context) {
-    // empty for now
+    // TODO: uncomment once we have the other extensions available
```
mstachniuk, line 32: I think this can be uncommented now

iac-extensions/arm/src/test/java/org/sonar/iac/arm/plugin/ArmSensorTest.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Override
+  protected InputFile validFile() {
+    return inputFile("object.json", "{\"key\":\"value\"}");
+  }
+
+  @Override
+  protected void verifyDebugMessages(List<String> logs) {
+    // TODO
```
mstachniuk, line 104: You can simply add `assertThat(logTester.logs(LoggerLevel.DEBUG)).isEmpty()` or some equivalent.

mstachniuk (approved):

iac-extensions/arm/src/main/java/org/sonar/iac/arm/plugin/ArmExtension.java
rudy-regazzoni-sonarsource: Not yet, unfortunately: If I do so, I get an error message because it is expecting to have an associated profile extension, and since it cannot find it, it is not able to instantiate the bean and so crash.
Should be good in the second subticket!

iac-extensions/arm/src/test/java/org/sonar/iac/arm/plugin/ArmSensorTest.java
rudy-regazzoni-sonarsource: I replaced with proper values in the end, didn't saw that this part was indirectly already done and ready for test.

PR #683, opened 2023-05-23
SONARIAC-835 ARM Parse empty JSON to File

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmParser.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.annotation.Nullable;
+import org.sonar.iac.arm.tree.api.ArmTree;
+import org.sonar.iac.arm.tree.impl.json.FileImpl;
+import org.sonar.iac.common.extension.ParseException;
+import org.sonar.iac.common.extension.TreeParser;
+import org.sonar.iac.common.extension.visitors.InputFileContext;
+import org.sonar.iac.common.yaml.YamlParser;
+import org.sonar.iac.common.yaml.tree.FileTree;
+
+public class ArmParser implements TreeParser<ArmTree> {
```
rudy-regazzoni-sonarsource, line 32: I see you are later using a `YamlParser`.
I checked how `CloudFormationParser` is done, it is currently extending directly from the `YamlParser`.
Do you think it would make sense here? Seems a bit strange to instantiate a different parser in your parser.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmParserTest.java
```diff
@@ -0,0 +1,71 @@
...
+    File tree = (File) parser.parse("{}", null);
+    assertThat(tree.statements()).isEmpty();
+    assertThat(tree.children()).isEmpty();
+    assertThat(tree.parent()).isNull();
+    Throwable throwable = catchThrowable(tree::textRange);
+    assertThat(throwable).isInstanceOf(IllegalArgumentException.class);
+
+    // temp test for coverage
+    tree.setParent(tree);
+    assertThat(tree.parent()).isSameAs(tree);
```
rudy-regazzoni-sonarsource, line 49: Suggestion: if this is to be replaced later, maybe we should have this comment in the `TODO` form, to have it raised and kept in Sonar until we remove it?

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmParserTest.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @Test
+  void shouldThrowExceptionWhenParseError() {
+    Throwable throwable = catchThrowable(() -> parser.parse("{", null));
+    assertThat(throwable)
+      .isInstanceOf(ParseException.class)
+      .hasMessage("Failed to parse");
```
rudy-regazzoni-sonarsource, line 57:
```suggestion
  assertThatThrownBy(() -> parser.parse("{", null))
    .isInstanceOf(ParseException.class)
    .hasMessage("Failed to parse");
```
Small suggestion regarding assertions on throwable, I saw that we are doing it that way at other places, may be a bit simpler.

iac-extensions/cloudformation/src/main/java/org/sonar/iac/cloudformation/plugin/CloudformationSensor.java
```diff
@@ -62,7 +61,7 @@ public CloudformationSensor(SonarRuntime sonarRuntime, FileLinesContextFactory f
 
   @Override
   protected TreeParser<Tree> treeParser() {
-    return new YamlParser(new CloudformationConverter());
+    return new CloudformationParser();
   }
```
rudy-regazzoni-sonarsource, line 65: Good point!

rudy-regazzoni-sonarsource (approved):
LGTM! Nice work, please see my suggestions.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmParser.java
mstachniuk: The `CloudFormationParser` extends `YamlParser` to use `CloudformationConverter` to do some magic, we don't need it.
The second reason is that methods defined in `YamlParser` return types from `org.sonar.iac.common.yaml.tree` package and we would have another tree structure just for ARM.
Finally, it will be easier to extend it or refactor it in case of implementing Bicep parser in the future.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmParserTest.java
mstachniuk: ok

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmParserTest.java
mstachniuk: OK it is a bit simpler, I will change it here because those tests are small. IMO it is also ok to use both formats. In `catchThrowable` the tested method is in a single line, so it is easier to find it in long tests. In `assertThatThrownBy` putting a new line after to visually separate "given when then" part looks strange.

nils-werner-sonarsource (commented):
I've added two comments regarding an annotation about nullability. As well I think you can improve the coverage when implementing a FileImplTest, can't you?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/AbstractArmTreeImpl.java
```diff
@@ -0,0 +1,52 @@
...
+  protected ArmTree parent;
+
+  @Override
+  public TextRange textRange() {
+    if (textRange == null) {
+      List<TextRange> childRanges = children().stream().map(HasTextRange::textRange).collect(Collectors.toList());
+      textRange = TextRanges.merge(childRanges);
+    }
+    return textRange;
+  }
```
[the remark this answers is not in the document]
mstachniuk: I thought about `default` method in `Tree` interface, but then it needs to be recalculated every time because fields in Java interfaces are `final`. I'm not sure if introducing a next level in our already big class hierarchy to save 8 lines of code is a good idea 🤔  The benefit is small in comparison to complexity. 
Anyway if still needed it's better to handle it in separate PR

PR #686, opened 2023-05-26
SONARIAC-838 Convert JSON to Minimal ARM AST Model: ARM Parse Output

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmParser.java
```diff
@@ -99,31 +125,113 @@ private static List<ResourceDeclaration> convertResources(SequenceTree resource)
...
+   * Transform a collection of TupleTree into a Map of Property to easily find specific properties and process them.
+   * This is a recursive method which will also convert sub-object, adding a prefix at every level.
+   * Example :
+   * {
+   *   "key1" : "value1",
+   *   "key2 : {
+   *     "key3" : "value3"
+   *   }
+   * }
+   * => Will be translated into this Map {"key1":"value1", "key2.key3":"value3"}
```
mstachniuk, line 212: What if `key2` contains a dot? It's a valid string in JSON. I'm not sure what character might be good here as separator

mstachniuk (changes requested):
Please see my comments

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ResourceDeclarationTest.java
```diff
@@ -19,14 +19,12 @@
  */
 package org.sonar.iac.arm.parser;
```
mstachniuk, line 20: It should be moved to `arm.tree.impl.json` (I will rever my changes in this file

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/OutputDeclaration.java
```diff
@@ -19,5 +19,22 @@
...
+  String condition();
+
+  @CheckForNull
+  String copyCount();
+
+  @CheckForNull
+  String copyInput();
+
+  @CheckForNull
+  String value();
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: That's a good question, sometimes we return an `Expression`/`Identifier`, sometimes we return a `String`.
A good example is the `ResourceDeclaration` interface, we also have a mix of `String` and `Identifier`.
We should align, I would be tempted to return the `Expression`/`Identifier` because I'm suspecting that in Bicep we may get something really different, while in `JSON` we always have `String` (even if sometimes they are not "just string", for example: `"concatToVar": "[concat(variables('stringVar'), '-addtovar') ]"`).

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/utils/ArmAssertions.java
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.arm.parser.utils;
+
+import javax.annotation.Nullable;
+import org.sonar.iac.arm.tree.api.ArmTree;
+
+public class ArmAssertions {
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: I missed that, I will merge my version into the existing one.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmParser.java
rudy-regazzoni-sonarsource: :thinking_face: I didn't think about it.
Maybe something that can be tackled separately in the ticket I created [Improve Property to support array and object](https://sonarsource.atlassian.net/browse/SONARIAC-840).
I will check if I can change the approach quickly here otherwise.

mstachniuk (approved):
LGTM

PR #687, opened 2023-05-30
SONARIAC-840 Convert JSON to Minimal ARM AST Model: Improve Property to support array and object

rudy-regazzoni-sonarsource:
Changed a bit the approach regarding `Property`:
- `Property` as it was has been renamed to `SimpleProperty`: the key is an `Identifier` and the value is always an `Expression`
- Created a `PropertyValue` interface, with three implementations:
  - `Expression`, to represent simple value such as `"string"` or `5`
  - `ArrayExpression`, to represent a typical array such as `["value1", "value2"]`, where each entry is a `PropertyValue` also
  - `ObjectExpression`, which is an object like `{"key":"value"}`, represented as a `List<Property>` and some methods to use it as a `Map`
- `Property` still has `Identifier` as a key, but the value is now a `PropertyValue`

Added some useful methods regarding those changes, like a way to extract and/or convert elements to `SimpleProperty`, which are mainly used in our convertors for `ResourceDeclaration`, `ParameterDeclaration` and `OutputDeclaration`.

I also removed the method `checkMandatoryObject` for several reasons:
- it was a bit hacky
- it obfuscated the code and raised FP on some check-null rules for both sonarlint and intellij
- it would parse the whole object and fail at the end, once it discovers that one property was missing

The new approach is to require specific properties and to fail fast / as soon as we detect they are missing.

### DIFF
--- a/iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/json/ResourceDeclarationTest.java
+++ b/iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/json/ResourceDeclarationTest.java
@@ -20,15 +20,17 @@
 package org.sonar.iac.arm.parser.json;
 
 import java.util.List;
-import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 import org.sonar.iac.arm.parser.ArmParser;
 import org.sonar.iac.arm.tree.api.ArmTree;
+import org.sonar.iac.arm.tree.api.ArrayExpression;
 import org.sonar.iac.arm.tree.api.Expression;
 import org.sonar.iac.arm.tree.api.File;
+import org.sonar.iac.arm.tree.api.ObjectExpression;
 import org.sonar.iac.arm.tree.api.Property;
+import org.sonar.iac.arm.tree.api.PropertyValue;
 import org.sonar.iac.arm.tree.api.ResourceDeclaration;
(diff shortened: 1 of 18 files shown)


iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/json/ResourceDeclarationTest.java
```diff
@@ -71,23 +77,60 @@ void shouldParseResource() {
...
+    assertThat(properties.get(1).key().value()).isEqualTo("other properties 2");
+    assertThat(properties.get(1).value().is(ARRAY_EXPRESSION)).isTrue();
+    ArrayExpression arrayExpression = (ArrayExpression) properties.get(1).value();
+    assertThat(arrayExpression.values()).hasSize(1);
+    assertThat(arrayExpression.children()).hasSize(1);
+    PropertyValue arrValue = arrayExpression.values().get(0);
+    assertThat(arrValue.is(EXPRESSION)).isTrue();
+    assertThat(((Expression) arrValue).value()).isEqualTo("val");
+
+    IacCommonAssertions.assertThat(properties.get(0).textRange()).hasRange(7, 6, 7, 53);
```
mstachniuk, line 133: static import

rudy-regazzoni-sonarsource: Not possible:
```
Ambiguous method call. Both
assertThat (TextRange) in ArmAssertions and
assertThat (TextRange) in IacCommonAssertions match
```

PR #689, opened 2023-05-30
SONARIAC-846 Convert JSON to Minimal ARM AST Model: Refactor ArmParser

rudy-regazzoni-sonarsource:
Some modifications:
- moved most codes from ArmParser to ArmBaseConverter (base converter class with all basic general converter methods), ResourceDeclarationConverter, ParameterDeclarationConverter and OutputDeclarationConverter)
- did some modifications to ParameterDeclarationConverter: some methods were changed due to previous ticket SONARIAC-840 based on the same logic that was present in this method, changed the call to those new methods

### DIFF
--- a/iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
+++ b/iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
@@ -50,6 +51,7 @@ import org.sonar.iac.common.extension.visitors.InputFileContext;
 import org.sonar.iac.common.yaml.tree.MappingTree;
 import org.sonar.iac.common.yaml.tree.ScalarTree;
 import org.sonar.iac.common.yaml.tree.SequenceTree;
+import org.sonar.iac.common.yaml.tree.TupleTree;
 import org.sonar.iac.common.yaml.tree.YamlTree;
 import org.sonar.iac.common.yaml.tree.YamlTreeMetadata;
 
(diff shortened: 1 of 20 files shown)


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/api/PropertyValueAssert.java
nils-werner-sonarsource, line 1: I'm not happy to have such a specific custom assertion. 
I wonder, why we have assertions for a single kind. They are super-specific and will probably never be used somewhere else. IMO it is only creating overhead. I would be okay with an assertion like `isKind(ArmTree.Kind kind)`.
The only specific assertion is `hasValue(String value)` which is okay to have in such a class. The `isKind()` would make more sense in a generic `ArmTreeAssertion` class.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/api/PropertyValueAssert.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public PropertyValueAssert hasValue(String value) {
+    Assertions.assertThat(((Expression) actual).value()).isEqualTo(value);
+    return this;
+  }
+
+  public PropertyValueAssert hasRange(int startLine, int startLineOffset, int endLine, int endLineOffset) {
+    TextRangeAssert.assertThat(actual.textRange()).hasRange(startLine, startLineOffset, endLine, endLineOffset);
+    return this;
+  }
```
nils-werner-sonarsource, line 57: IMO it would make sense to move such an assertion to a common custom assertion class since this method is given via the `HasTextRange` interface. So having this in a dedicated common class would allow other tests even in other extensions to make use of it.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
```diff
@@ -0,0 +1,200 @@
...
+  // Cast methods
+  public MappingTree toMappingTree(YamlTree tree) {
+    if (!(tree instanceof MappingTree)) {
+      throw new ParseException("Expected MappingTree, got " + tree.getClass().getSimpleName(), new BasicTextPointer(tree.metadata().textRange()), null);
+    }
+    return (MappingTree) tree;
+  }
+
+  public Expression toExpression(PropertyValue propertyValue) {
+    throwErrorIfUnexpectedType("toExpression", ArmTree.Kind.EXPRESSION, propertyValue);
```
nils-werner-sonarsource, line 173: Any kind of Exception and Log message should start with an upper case.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/api/PropertyValueAssert.java
rudy-regazzoni-sonarsource: Thanks for your comment.
I created [a ticket](https://sonarsource.atlassian.net/browse/SONARIAC-848) to handle this separately.

PR #691, opened 2023-05-31
Update CI pipeline task names

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM!
To double-check: I see that tasks `Mend SCA` and `Build (Windows)` seems to not depend anymore on task `build`/`Build (Linux)`, is that normal? The PR name suggests that it is only about renaming tasks.

nils-werner-sonarsource:
> LGTM! To double-check: I see that tasks `Mend SCA` and `Build (Windows)` seems to not depend anymore on task `build`/`Build (Linux)`, is that normal? The PR name suggests that it is only about renaming tasks.

I simply removed the duplicate `depend-on`. Both tasks are using the `ONLY_IF_SONARSOURCE_QA` filter which includes the dependency on `Build (Linux)`. And I agree, I could have this added to the ticket.

PR #692, opened 2023-06-01
SONARIAC-845 Convert JSON to Minimal ARM AST Model: ARM Parse an array in property in resource declaration

rudy-regazzoni-sonarsource:
Supported without code modification since https://sonarsource.atlassian.net/browse/SONARIAC-840, just adding the test case.


nils-werner-sonarsource (approved):
LGTM!

PR #697, opened 2023-06-05
SONARIAC-853 Change ResourceDeclaration properties access and use PropertyUtils to extract properties

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
```diff
@@ -53,27 +53,50 @@
...
+  public ArrayExpression toArrayExpression(PropertyTree property) {
+    if (!(property.value() instanceof SequenceTree)) {
+      throw convertError(property, "ArrayExpression", "SequenceTree");
+    }
+    return toArrayExpression((SequenceTree) property.value());
   }
 
-  public Identifier convertToIdentifier(YamlTree tree) {
+  public Identifier toIdentifier(YamlTree tree) {
     return Optional.of(tree)
```
nils-werner-sonarsource, line 100: Sometime Optionals do not make it more readable IMO

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ResourceDeclarationConverter.java
```diff
@@ -59,12 +60,13 @@ private static List<MappingTree> mappingTreeOnly(List<YamlTree> yamlTrees) {
...
-    Property version = extractMandatoryProperty(tree.metadata(), properties, "apiVersion", ArmTree.Kind.STRING_LITERAL);
-    Property name = extractMandatoryProperty(tree.metadata(), properties, "name", ArmTree.Kind.STRING_LITERAL);
-    List<Property> otherProperties = new ArrayList<>(properties.values());
+    StringLiteral type = PropertyUtils.get(tree, "type").map(this::toStringLiteral).orElseThrow(() -> missingMandatoryAttributeError(tree, "type"));
+    StringLiteral version = PropertyUtils.get(tree, "apiVersion").map(this::toStringLiteral).orElseThrow(() -> missingMandatoryAttributeError(tree, "apiVersion"));
+    StringLiteral name = PropertyUtils.get(tree, "name").map(this::toStringLiteral).orElseThrow(() -> missingMandatoryAttributeError(tree, "name"));
+    List<Property> otherProperties = PropertyUtils.get(tree, "properties")
+      .map(PropertyTree::value)
+      .map(this::toProperties)
+      .orElse(Collections.emptyList());
```
nils-werner-sonarsource, line 69: I love it.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/OutputDeclaration.java
```diff
@@ -24,17 +24,17 @@
 public interface OutputDeclaration extends Statement {
   Identifier name();
 
-  Expression type();
+  String type();
```
nils-werner-sonarsource, line 27: I know this is part of the spec, and now I wonder if returning something like StringLiteral and NumericLiteral would make more sense. That way we would keep access to the range of the elements.

nils-werner-sonarsource (approved):
LGTM!

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/OutputDeclaration.java
rudy-regazzoni-sonarsource: Was thinking the same, It will probably make more sense for this reason.
It also seems more evident now that we have those literal classes, I will make the modification.

PR #702, opened 2023-06-14
Fix usage of StringUtils and fix ruling test

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
```diff
@@ -181,7 +180,7 @@ public String filenameAndPosition(TextRange textRange) {
     String position = textRange.start().line() + ":" + textRange.start().lineOffset();
     if (inputFileContext != null) {
       String filename = inputFileContext.inputFile.filename();
-      if (StringUtils.isNotBlank(filename)) {
+      if (filename != null) {
         return filename + ":" + position;
       }
```
rudy-regazzoni-sonarsource, line 185: I would be tempted to also check if the string is blank, like it was done with `StringUtils`.
```suggestion
      if (filename != null && !filename.isBlank()) {
        return filename + ":" + position;
      }
```

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my small comment.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
mstachniuk: The tests are green, but ok let's do it

rudy-regazzoni-sonarsource (approved):
LGTM!

PR #707, opened 2023-06-14
SONARIAC-872 Change OutputDeclaration API value() to return an Expression instead

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM

PR #708, opened 2023-06-15
SONARIAC-873 Change ResourceDeclaration API name() to return an Identifier instead

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM

PR #714, opened 2023-06-16
SONARIAC-877 Add ArchUnit to avoid common mistakes

mstachniuk:
Improve the check for public constructors in Sensors. Previously one public constructor was enough, now all constructors needs to be public.


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #715, opened 2023-06-16
SONARIAC-849 Refactor ArmConverter to reduce duplication

mstachniuk:


nils-werner-sonarsource (approved):
LGTM! Sometimes it would be clear when the different helper methods would accept the same type. Currently, there is a mixture of PropertyTree and YamlTree which makes it hard to understand. As well as the invocation - it's a mixture of `tree` and `tree.value()`. Nevertheless, it's an improvement and we can discuss further steps in the next iterations.

mstachniuk:
Yes, you are right. In some cases I needed to keep `tree.value`, because a `tree` was also used in the method, see `toStringLiteralOrException()`.

PR #716, opened 2023-06-16
SONARIAC-852 Move specific element test like into dedicated test classes instead of having everything in Parent class

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/ArrayExpressionImplTest.java
```diff
@@ -0,0 +1,60 @@
...
+    assertThat(arrayProperty.value())
+      .asArrayExpression()
+      .containsValuesExactly("val")
+      .hasRange(LINE_OFFSET + 1, 14, LINE_OFFSET + 1, 21);
+  }
+
+  @Test
+  void shouldParseEmptyArrayExpression() {
+    Property arrayProperty = parseProperty(parser, "\"array_prop\": []");
+    assertThat(arrayProperty.value()).asArrayExpression().containsValuesExactly();
```
mstachniuk, line 46: Maybe  a new method `isEmpty()` is needed instead of `containsValuesExactly()` - it looks a little bit strange.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/ArrayExpressionImplTest.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Test
+  void shouldParseArrayExpressionFilledWithDifferentTypesOfValues() {
+    Property arrayProperty = parseProperty(parser, "\"array_prop\": [\"val\", 5, {\"key\":\"val\"}, [\"val\"]]");
+    assertThat(arrayProperty.value()).asArrayExpression();
```
mstachniuk, line 52: IMO this line is not needed, the missing assignment looks strange (I know there is a check also under the hood).
If the value will be not an `ArrayExpression`, the exception from the next line will inform us about it.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/NumericLiteralImplTest.java
```diff
@@ -23,25 +23,24 @@
...
-    File tree = (File) parser.parse(code, null);
-
-    Property numericProperty = ((ResourceDeclaration) tree.statements().get(0)).properties().get(0);
-    assertThat(numericProperty.value()).asNumericLiteral().hasValue(7);
+    Property numericProperty = parseProperty(parser, "\"numeric\": 7");
+    assertThat(numericProperty.value())
+      .asNumericLiteral()
+      .hasValue(7)
+      .hasRange(LINE_OFFSET + 1, 11, LINE_OFFSET + 1, 12);
   }
```
mstachniuk, line 44: IMO more detailed tests for the following methods are needed here: `getKind()`, `children()`, and `parent()`. Similar in other test classes in this PR.

mstachniuk (approved):
LGTM and please see my suggestions.

PR #717, opened 2023-06-19
SONARIAC-880  Simplify ARM rule implementation with abstraction layer

nils-werner-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/IpRestrictedAdminAccessCheck.java
```diff
@@ -20,61 +20,69 @@
...
+      checkResourcePath("securityRules/*/properties"));
+
+    register("Microsoft.Network/virtualNetworks/subnets",
+      checkResourcePath("networkSecurityGroup/properties/securityRules/*/properties"));
+
+    register("Microsoft.Network/virtualNetworks",
+      checkResourcePath("subnets/*/properties/networkSecurityGroup/properties/securityRules/*/properties"));
+
+    register("Microsoft.Network/networkInterfaces",
+      checkResourcePath("ipConfigurations/*/properties/subnet/properties/networkSecurityGroup/properties/securityRules/*/properties"));
```
nils-werner-sonarsource, line 69: The change should make the code more readable. This is not super performant since the list will be generated on every resource match. However, we do not expect more than 10 of these resources per analysis, so the consumer may be called less than a dozen times.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/IpRestrictedAdminAccessCheck.java
```diff
@@ -24,57 +24,64 @@
...
+    register("Microsoft.Network/virtualNetworks/subnets",
+      (ctx, resource) -> checkResourcePath(ctx, resource, propertyPath("networkSecurityGroup/properties/securityRules/*/properties")));
+
+    register("Microsoft.Network/virtualNetworks",
+      (ctx, resource) -> checkResourcePath(ctx, resource, propertyPath("subnets/*/properties/networkSecurityGroup/properties/securityRules/*/properties")));
+
+    register("Microsoft.Network/networkInterfaces",
+      (ctx, resource) -> checkResourcePath(ctx, resource,
+        propertyPath("ipConfigurations/*/properties/subnet/properties/networkSecurityGroup/properties/securityRules/*/properties")));
+  }
```
nils-werner-sonarsource, line 71: The change should make the code more readable. This is not super performant since the list will be generated on every resource match. However, we do not expect more than 10 of these resources per analysis, so the consumer may be called less than a dozen times.

rudy-regazzoni-sonarsource: It's definitely more readable!
I was indeed at first worried a bit about the performance, but as you said there won't be that much resource per analysis, it's different from a typical programming language where we could potentially expect a lot of usage of specific instruction/keywords.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice improvement :muscle:

PR #719, opened 2023-06-19
SONARIAC-856 Rule S6321 ARM: add support for nested childs

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (approved):
LGTM! Even without a more complex bidirectional relation!

PR #723, opened 2023-06-20
SONARIAC-816 Rule S6382: ARM Disabling certificate-based authentication is security-sensitive - containerApps

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
```diff
@@ -19,35 +19,69 @@
...
+    }
   }
 
   private static boolean isBooleanFalse(Tree tree) {
     return ((ArmTree) tree).is(ArmTree.Kind.BOOLEAN_LITERAL) && !((BooleanLiteral) tree).value();
   }
+
+  private static boolean isStringValue(Tree tree, String value) {
+    return ((ArmTree) tree).is(ArmTree.Kind.STRING_LITERAL) && ((StringLiteral) tree).value().equals(value);
+  }
```
rudy-regazzoni-sonarsource, line 82:
```suggestion
  private static boolean isString(Tree tree, Predicate<String> predicate) {
    return TextUtils.matchesValue(tree, predicate).isTrue();
  }
```
FYI I replaced this later with a more generic method + added the usage of TextUtils which I tend to forget :sweat_smile:
I can also adapt this PR if you think it make sense.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
```diff
@@ -19,35 +19,69 @@
...
+          if (isBooleanFalse(property.value())) {
+            ctx.reportIssue(property, DISABLED_CERTIFICATE_MESSAGE);
           }
         },
-        () -> ctx.reportIssue(resource.type(), String.format(MISSING_CERTIFICATE, "negotiateClientCertificate")));
+        () -> ctx.reportIssue(resource.type(), String.format(MISSING_CERTIFICATE_MESSAGE, "negotiateClientCertificate")));
+  }
+
+  private static void checkContainerApps(CheckContext ctx, ResourceDeclaration resource) {
+    if (isResourceVersionBefore(resource, "2022-10-01")) {
```
nils-werner-sonarsource, line 59: In the ticket, it should be `2022-10-01` or newer. IMO the method name supposed that the opposite is tested.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
```diff
@@ -19,35 +19,69 @@
...
-          if (isBooleanFalse(value)) {
-            ctx.reportIssue(value, DISABLED_CERTIFICATE);
+        property -> {
+          if (isBooleanFalse(property.value())) {
+            ctx.reportIssue(property, DISABLED_CERTIFICATE_MESSAGE);
           }
         },
-        () -> ctx.reportIssue(resource.type(), String.format(MISSING_CERTIFICATE, "negotiateClientCertificate")));
+        () -> ctx.reportIssue(resource.type(), String.format(MISSING_CERTIFICATE_MESSAGE, "negotiateClientCertificate")));
+  }
```
nils-werner-sonarsource, line 56: When looking at this, I wonder if we actually use the DSL of the Terraform checks. That way we would write this logic like this:
```
resource.attribute("negotiateClientCertificate")
 .reportIf(isFalse(), DISABLED_CERTIFICATE_MESSAGE)
 .reportIfAbsent(MISSING_CERTIFICATE_MESSAGE)
```
Only a comment, no action is needed here. But maybe we can discuss it.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
```diff
@@ -19,35 +19,69 @@
...
+        PropertyUtils.get(resolvedTree, "clientCertificateMode")
+          .ifPresentOrElse(
+            property -> {
+              if (isStringValue(property.value(), "accept")) {
+                ctx.reportIssue(property, ALLOWING_NO_CERTIFICATE_MESSAGE);
+              } else if (isStringValue(property.value(), "ignore")) {
+                ctx.reportIssue(property, DISABLED_CERTIFICATE_MESSAGE);
+              }
+            },
+            () -> ctx.reportIssue(resolvedTree, String.format(MISSING_CERTIFICATE_MESSAGE, "clientCertificateMode")));
```
nils-werner-sonarsource, line 71: Same here. It would be like this:
```
container.attribute("clientCertificateMode")
 .reportIf(isValue("accept"), ALLOWING_NO_CERTIFICATE_MESSAGE)
 .reportIf(isValue("ignore"), DISABLED_CERTIFICATE_MESSAGE)
 .reportIfAbsent(MISSING_CERTIFICATE_MESSAGE)
```

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
nils-werner-sonarsource: My naming suggestion would be `isValue()` or `hasValue()` because you are not testing if the tree is of type string but if the tree is a string with an expected value.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
```diff
@@ -19,35 +19,69 @@
...
+            ctx.reportIssue(property, DISABLED_CERTIFICATE_MESSAGE);
           }
         },
-        () -> ctx.reportIssue(resource.type(), String.format(MISSING_CERTIFICATE, "negotiateClientCertificate")));
+        () -> ctx.reportIssue(resource.type(), String.format(MISSING_CERTIFICATE_MESSAGE, "negotiateClientCertificate")));
+  }
+
+  private static void checkContainerApps(CheckContext ctx, ResourceDeclaration resource) {
+    if (isResourceVersionBefore(resource, "2022-10-01")) {
+      List<Tree> resolvedTrees = ArmTreeUtils.resolveTreePath(resource, new LinkedList<>(PATH_CONTAINER_APPS));
```
nils-werner-sonarsource, line 60: I would prefer to choose a more descriptive name like `containers` or `configuration`. IMO `resolvedTrees` is super generic.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/ArmTreeUtils.java
nils-werner-sonarsource, line 1: My PR introduced the CheckUtils. I used to have such logic closer to the checks than to the trees. However, there is no rule on when to have a TreeUtils and when to have a CheckUtils class.

iac-extensions/arm/src/test/resources/checks/CertificateBasedAuthenticationCheck/Microsoft.App_containerApps/test.json
nils-werner-sonarsource, line 1: I would also prefer to have a more specific file name. That way, searching for the file will not end in a mess with dozens of `test.json` files. And I have to admit, that we already have a lot of similar files due to my contributions in the past. We can start to improve our test file naming over time.

nils-werner-sonarsource (approved):
LGTM! Please see my comments.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/CertificateBasedAuthenticationCheck.java
rudy-regazzoni-sonarsource: I'm currently not satisfied with the readability of this check.
The syntax you are proposing looks really well and easy to read, would love to be able to use it!

Still, I am working on the same main tickets and have implemented about 4 more other resources check.
I encountered several cases where we need to report an issue only when multiple conditions on different properties are true, like:
- `resource.properties.activities.*.type` is `WebActivity` or `WebHook`
- `resource.properties.activities.*.typeProperties.authenticationType` is `Basic` or `ServicePrincipal`

So we would need to be able to adapt it over time to support such use cases, not sure what are the current possibilities of DSL Terraform checks currently in place. But definitely it would bring a lot of improvement in writing checks.

PR #731, opened 2023-06-21
SONARIAC-791 [Multiple resource types] properties/publicNetworkAccess

mstachniuk:
I hope you like the way of testing all these resources. Preparing a separate file for each will be a nightmare.


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
```diff
@@ -32,14 +35,128 @@ public class PublicNetworkAccessCheck extends AbstractArmResourceCheck {
...
+  // Method has 101 lines, which is greater than 100 authorized - it will be refactored when SONARIAC-890 will be done
+  @SuppressWarnings("java:S138")
   protected void registerResourceConsumer() {
-    register("Microsoft.DesktopVirtualization/hostPools",
-      (ctx, resource) -> PropertyUtils.value(resource, "publicNetworkAccess")
-        .filter(PublicNetworkAccessCheck::isSensitiveStringLiteral)
-        .ifPresent(value -> ctx.reportIssue(value, "Make sure allowing public network access is safe here.")));
+    register("Microsoft.DesktopVirtualization/hostPools", checkPublicNetworkAccess());
+
+    register("Microsoft.AgFoodPlatform/farmBeats", checkPublicNetworkAccessSimplified());
```
rudy-regazzoni-sonarsource, line 43: A bit heavy to read :sweat_smile: Maybe put all those resources in a list and have a `.forEach()` call on it, WDYT?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
```diff
@@ -32,14 +35,128 @@ public class PublicNetworkAccessCheck extends AbstractArmResourceCheck {
...
+      .ifPresent(value -> ctx.reportIssue(value, "Make sure allowing public network access is safe here."));
+  }
+
+  private static boolean isSensitivePublicNetworkAccess(Tree tree) {
     return ((ArmTree) tree).is(ArmTree.Kind.STRING_LITERAL) && SENSITIVE_VALUES.contains(((StringLiteral) tree).value());
   }
+
+  private static boolean isSensitivePublicNetworkAccessSimplified(Tree tree) {
+    return ((ArmTree) tree).is(ArmTree.Kind.STRING_LITERAL) && "Enabled".equals(((StringLiteral) tree).value());
+  }
```
rudy-regazzoni-sonarsource, line 161:
```suggestion
  private static boolean isSensitivePublicNetworkAccess(Tree tree) {
    return TextUtils.matchesValue(tree, SENSITIVE_VALUES::contains).isTrue();
  }

  private static boolean isSensitivePublicNetworkAccessSimplified(Tree tree) {
    return TextUtils.matchesValue(tree, "Enabled"::contains).isTrue();
  }
```
Suggesting the use of `TextUtils`: our `StringLiteral` complies to its contract and it simplify the checks.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
rudy-regazzoni-sonarsource, line 1: It is missing a test of a resource that would not be in the list of checked resources.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
```diff
@@ -19,19 +19,173 @@
...
       issue(26, 31, 26, 54));
   }
+
+  static Stream<Arguments> shouldCheckPublicNetworkAccessSimplified() {
+    return Stream.of(
+      of("Microsoft.AgFoodPlatform/farmBeats", "2021-09-01-preview"),
+      of("Microsoft.ApiManagement/service", "2021-08-01"),
+      of("Microsoft.ApiManagement/service", "2021-04-01-preview"),
+      of("Microsoft.AppConfiguration/configurationStores", "2020-06-01"),
+      of("Microsoft.AppConfiguration/configurationStores", "2019-11-01-preview"),
```
rudy-regazzoni-sonarsource, line 57: There is no test on the version, wondering why we are putting it in the test.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice job, please see my few comments.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
mstachniuk: The new method `protected void register(List<String> resourceTypes, BiConsumer<CheckContext, ResourceDeclaration> consumer)` will be provided by Nils, and I will use it. But ok I can do it for now.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
mstachniuk: I put them here to generate a valid, deployable input. If you think it doesn't make sense I can replace it by "dummy" apiVersion. WDYT?

rudy-regazzoni-sonarsource: I would tend to remove it from there and directly inject your dummy apiVersion in the template if it has no impact on the rule detection logic.
As of now it seems we are injecting very specific version for each resource while in fact this is not used.
FYI on some rules we do have a check on the API version (Ex: only apply if version is 2021-10-01 or later), so that's why I was wondering here.

PR #737, opened 2023-06-22
SONARIAC-892 Build an ARM Domain Specific Language to simplify rule implementation (Property & Object symbol)

nils-werner-sonarsource:
Removed the MapSymbol from the common package. It does not provide any value but only more complexity. Sorry for the number of files. However, it should not introduce more logic.

### DIFF
--- a/iac-common/src/test/java/org/sonar/iac/common/dsl/SymbolTest.java
+++ b/iac-common/src/test/java/org/sonar/iac/common/dsl/SymbolTest.java
@@ -1,0 +1,114 @@
+/*
+ * SonarQube IaC Plugin
+ * Copyright (C) 2021-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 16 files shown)


iac-common/src/test/java/org/sonar/iac/common/dsl/SymbolTest.java
```diff
@@ -0,0 +1,114 @@
...
+    assertThat(absent.toSecondary("secondary")).isNull();
+  }
+
+  @Test
+  void reportIfAbsent() {
+    present.reportIfAbsent("present");
+    absent.reportIfAbsent("absent");
+
+    verify(ctx, times(1)).reportIssue(parentTree, "absent", Collections.emptyList());
+  }
```
rudy-regazzoni-sonarsource, line 85: Same as above, you are calling `reportIfAbsent` on both `present` and `absent`, but you are only checking there is one call with `"absent"`, you should also check that there is no call with `"present"`.

rudy-regazzoni-sonarsource (approved):
LGTM! Really great code! The use of generic is really great here and this code is a masterpiece, good job!
Please see my few comments.

PR #738, opened 2023-06-23
SONARIAC-801 [Multiple resource types] properties/startIpAddress and endIpAddress

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
```diff
@@ -0,0 +1,207 @@
...
+        try {
+          int x = Integer.parseInt(s);
+          if (x < 0 || x > 255) {
+            return false;
+          }
+        } catch (NumberFormatException e) {
+          return false;
+        }
+      }
+      return true;
```
rudy-regazzoni-sonarsource, line 154: Alternatively, you could use the regex from OWASP:
`^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$`
It's a bit heavy and I'm not sure how fast/slow it is compared to your implementation.
It could also be used at the same time to capture each group.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
```diff
@@ -0,0 +1,207 @@
...
+    // The IP Address can be represented as 32 bit int, but to not have negative integers, the long in necessary here.
+    long toLong() {
+      String[] split = address.split("\\.");
+      long x1 = Long.parseLong(split[0]);
+      long x2 = Long.parseLong(split[1]);
+      long x3 = Long.parseLong(split[2]);
+      long x4 = Long.parseLong(split[3]);
+
+      return x1 * BASE3 + x2 * BASE2 + x3 * BASE1 + x4;
+    }
```
rudy-regazzoni-sonarsource, line 167: I like this approach!

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
```diff
@@ -178,11 +180,56 @@ void shouldRaiseNoIssuesForUnknownType() {
...
   }
 
-  private static String readTemplateAndReplace(String type) {
-    Path filePath = Paths.get("src", "test", "resources", "checks", "PublicNetworkAccessCheckTest", "publicNetworkAccess-Simplifed").resolve("template.json");
+  static Stream<String> shouldCheckRangePublicIPAddress() {
+    return Stream.of(
+      "Microsoft.DBForMySql/flexibleServers/firewallRules",
+      "Microsoft.DBForPostgreSql/flexibleServers/firewallRules",
+      "Microsoft.DBforMariaDB/servers/firewallRules",
+      "Microsoft.DBforMySQL/flexibleServers/firewallRules",
```
rudy-regazzoni-sonarsource, line 188: It is missing a test on a resource not present in the list with a public range IP that does not raise an issue.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
rudy-regazzoni-sonarsource, line 1: I would have made an IP address validator not related to the `ArmTree` elements but expecting `String` instead so that it could be used in a different scope and not necessarily in the ARM Parser.
No need to change this now.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice job, please see my few comments.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
```diff
@@ -0,0 +1,207 @@
...
+  private final ArmTree endIpAddress;
+
+  private IpAddressValidator(@Nullable ArmTree startIpAddress, @Nullable ArmTree endIpAddress) {
+    this.startIpAddress = startIpAddress;
+    this.endIpAddress = endIpAddress;
+  }
+
+  public static IpAddressValidator fromStartToEnd(@Nullable ArmTree startIpAddress, @Nullable ArmTree endIpAddress) {
+    return new IpAddressValidator(startIpAddress, endIpAddress);
+  }
```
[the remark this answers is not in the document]
mstachniuk: It's a valid point since the parameters are in the same order. Maybe in the future, it will have more sense, so ok I will fix it now.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
```diff
@@ -0,0 +1,207 @@
...
+      return new ValidationResult(false, null, null);
+    }
+    String start = null;
+    String end = null;
+    if (startIpAddress != null && startIpAddress.is(ArmTree.Kind.STRING_LITERAL)) {
+      start = TextUtils.getValue(startIpAddress).orElse(DEFAULT_START_IP);
+    }
+    if (endIpAddress != null && endIpAddress.is(ArmTree.Kind.STRING_LITERAL)) {
+      end = TextUtils.getValue(endIpAddress).orElse(DEFAULT_END_IP);
+    }
```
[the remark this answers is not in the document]
mstachniuk: ok, extracted to `valueOrElseDefaultIfStringLiteralOrNull`, but I'm not sure if it is better now

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
mstachniuk: I'm sure that my impl will be faster - think about the whole state machine behind this regex.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
mstachniuk: good catch, I had it in mind but forgot about

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ipaddress/IpAddressValidator.java
mstachniuk: It can be refactored later. Other technologies use `1.2.3.4/32` instead of the stard+end range.

PR #745, opened 2023-06-26
SONARIAC-795 [Microsoft.Web/*] properties/siteConfig/publicNetworkAccess

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
```diff
@@ -132,6 +132,11 @@ public class PublicNetworkAccessCheck extends AbstractArmResourceCheck {
     "Microsoft.DocumentDB/mongoClusters/firewallRules",
     "Microsoft.Sql/servers/firewallRules",
     "Microsoft.Synapse/workspaces/firewallRules");
+
+  private static final List<String> PUBLIC_NETWORK_ACCESS_SIMPLIFIED_IN_SITE_CONFIG = List.of(
```
jonas-wielage-sonarsource, line 136: Why is this also `SIMPLIFIED`? Shouldn't this only be `PUBLIC_NETWORK_ACCESS_SIMPLIFIED_IN_SITE_CONFIG `? Same for all related methods and test files

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
```diff
@@ -236,12 +236,31 @@ void shouldCheckDbForMySqlFlexibleServers() {
...
   }
 
+  static Stream<String> shouldCheckPublicNetworkAccessSimplifiedInSiteConfig() {
+    return Stream.of(
+      "Microsoft.Web/sites",
+      "Microsoft.Web/sites/slots");
+  }
+
+  @MethodSource
+  @ParameterizedTest(name = "[${index}] should check range public IP Address for type {0}")
```
jonas-wielage-sonarsource, line 252: I didn't know that you could define custom names for the tests, looks quite nice!

jonas-wielage-sonarsource (approved):
LGTM! See my comments before merging

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
mstachniuk: The `publicNetworkAccess` property in some cases can have the following sensitive values: `"Enabled", "EnabledForSessionHostsOnly", or "EnabledForClientsOnly" ` and in other cases only `Enabled` is sensitive. I call the second case `SIMPLIFIED`. If you have a better name for it, I'm happy to refactor it.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheckTest.java
mstachniuk: It's very useful for Parameterized Test

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
jonas-wielage-sonarsource: Ah ok, I thought it had something to do with how deep the `enabled` is inside the properties.

PR #752, opened 2023-06-27
SONARIAC-802 [Microsoft.Blockchain/*] properties/firewallRules/startI…

mstachniuk:
…pAddress and endIpAddress

### DIFF
--- a/iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
+++ b/iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
@@ -20,18 +20,17 @@
 package org.sonar.iac.arm.checks;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.Set;
-import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import org.sonar.check.Rule;
+import org.sonar.iac.arm.checkdsl.ContextualArray;
+import org.sonar.iac.arm.checkdsl.ContextualMap;
+import org.sonar.iac.arm.checkdsl.ContextualProperty;
 import org.sonar.iac.arm.checkdsl.ContextualResource;
 import org.sonar.iac.arm.checks.ipaddress.IpAddressValidator;
-import org.sonar.iac.arm.tree.api.ArmTree;
-import org.sonar.iac.arm.tree.api.ResourceDeclaration;
-import org.sonar.iac.common.api.checks.CheckContext;
+import org.sonar.iac.arm.tree.api.Expression;
+import org.sonar.iac.common.api.tree.HasProperties;
 import org.sonar.iac.common.api.tree.Tree;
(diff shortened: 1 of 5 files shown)


iac-extensions/arm/src/test/resources/checks/PublicNetworkAccessCheckTest/rangePublicIPAddress-firewallRules/unknown-type.json
rudy-regazzoni-sonarsource, line 1: I would have limited the check with unknown type to only one case :sweat_smile:

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
```diff
@@ -179,12 +183,19 @@ private static boolean isSensitivePublicNetworkAccessSimplified(Tree tree) {
     return TextUtils.matchesValue(tree, "Enabled"::contains).isTrue();
   }
 
-  private static BiConsumer<CheckContext, ResourceDeclaration> checkIpRange() {
-    return (ctx, resource) -> {
-      Optional<Tree> startIpAddress = PropertyUtils.value(resource, "startIpAddress");
-      Optional<Tree> endIpAddress = PropertyUtils.value(resource, "endIpAddress");
-      IpAddressValidator validator = new IpAddressValidator((ArmTree) startIpAddress.orElse(null), (ArmTree) endIpAddress.orElse(null));
-      validator.reportIssueIfPublicIPAddress(ctx, PUBLIC_IP_ADDRESS_MESSAGE, PUBLIC_IP_ADDRESS_MESSAGE_SECONDARY_LOCATION);
+  private static <S extends ContextualMap<S, T>, T extends HasProperties & Tree> void checkIpRange(ContextualMap<S, T> resource) {
```
rudy-regazzoni-sonarsource, line 186: I wish there was a way to write this simpler, but I don't think there is.
The only way I found was to write it this way:
`private static void checkIpRange(ContextualMap<?, ?> resource) {`
I'm not convinced it's a good idea, I let you decide on this.

rudy-regazzoni-sonarsource (approved):
LGTM! Very complete test cases, please see my few comments!

iac-extensions/arm/src/test/resources/checks/PublicNetworkAccessCheckTest/rangePublicIPAddress-firewallRules/unknown-type.json
mstachniuk: In other cases for other types, there are checks for all cases, so I would keep it.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PublicNetworkAccessCheck.java
mstachniuk: But then the generic checks are weaker. I will keep it as it is.

PR #756, opened 2023-06-28
SONARIAC-775 Rule S5332: ARM Using clear-text protocols is security-sensitive; Microsoft.Storage/storageAccounts

jonas-wielage-sonarsource:


iac-extensions/arm/src/test/resources/checks/ClearTextProtocolsCheck/Microsoft.Storage_storageAccounts/test.json
rudy-regazzoni-sonarsource, line 1: I didn't saw that before, but I would encourage the use of more meaningful name for the json file, as explained in [this discussion](https://github.com/SonarSource/sonar-iac/pull/723#discussion_r1236600968).

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my comment.

PR #759, opened 2023-06-28
SONARIAC-800 Rule S6383: ARM Disabling Role-Based Access Control on Azure resources is security-sensitive

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/RbacAzureResourceCheck.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package org.sonar.iac.arm.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.iac.arm.checkdsl.ContextualResource;
+
+import static org.sonar.iac.arm.checks.utils.CheckUtils.isFalse;
+
+@Rule(key = "S6383")
+public class RbacAzureResourceCheck extends AbstractArmResourceCheck {
```
nils-werner-sonarsource, line 28: I would suggest naming this check `RoleBasedAccessControlCheck`. This is also the name of the check which is implemented for [Terraform](https://github.com/SonarSource/sonar-iac/blob/dda5f3a5cf7514fb6e5f75be38c2a16feabf51fd/iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/RoleBasedAccessControlCheck.java#L33).

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/RbacAzureResourceCheck.java
```diff
@@ -0,0 +1,53 @@
...
+  private static final String MESSAGE_MISSING_PARAMETER = "Omitting %s disables role-based access control for this resource. Make sure it is safe here.";
```
nils-werner-sonarsource, line 29:
```suggestion
  private static final String MESSAGE_MISSING_PARAMETER = "Omitting '%s' disables role-based access control for this resource. Make sure it is safe here.";
```
We use to have single or double quotes around these properties. Feel free to fix it for checks where you identify missing quotes.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/RbacAzureResourceCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+class RbacAzureResourceCheckTest {
+  @Test
+  void checkRbacPresentAndEnabledForManagedClusters() {
+    verify("RbacAzureResourceCheck/Microsoft.ContainerService_managedClusters/test.json",
+      new RbacAzureResourceCheck(),
+      issue(28, 10, 28, 34, "Make sure that disabling role-based access control is safe here."),
+      issue(30, 8, 30, 27, "Make sure that disabling role-based access control is safe here."),
+      issue(39, 10, 39, 34, "Make sure that disabling role-based access control is safe here."),
+      issue(52, 8, 52, 27, "Make sure that disabling role-based access control is safe here."));
```
nils-werner-sonarsource, line 35: You can, but you don't need to test always the message when there is no expected difference from the previous issue.

iac-extensions/arm/src/main/resources/org/sonar/l10n/azureresourcemanager/rules/azureresourcemanager/S6383.html
```diff
@@ -0,0 +1,105 @@
...
+<pre data-diff-id="1" data-diff-type="noncompliant">
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "resources": [
+    {
+      "type": "Microsoft.ContainerService/managedClusters",
+      "apiVersion": "2023-03-01",
+      "properties": {
+        "aadProfile": {
```
nils-werner-sonarsource, line 33: At first, I thought this is a type but it seems to be the acronym for `Azure Active Directory` :smile:

nils-werner-sonarsource (approved):
LGTM! Please see my comments.

PR #760, opened 2023-06-28
SONARIAC-778 Rule S5332: ARM Using clear-text protocols is security-sensitive; Microsoft.Cdn/profiles/endpoints

jonas-wielage-sonarsource:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/LogRetentionCheckTest.java
jonas-wielage-sonarsource, line 1: I deleted the line mentioning the schema in the json test files, so every issue line moved one. Same for the other test.

rudy-regazzoni-sonarsource: I would avoid doing so: the line about schema is a mandatory property, and I'm currently adding it as it is required for some rules to detect the use case :sweat_smile: (there are different schemas and the rule must raise only on few of them)

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my comment about adding back the `$schema` unless there is a concern I didn't get.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/LogRetentionCheckTest.java
jonas-wielage-sonarsource: Summary of discussion: This should not be a mandatory attribute.

PR #762, opened 2023-06-28
SONARIAC-780 Rule S5332: ARM Using clear-text protocols is security-sensitive; Microsoft.DBfor*

jonas-wielage-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #763, opened 2023-06-28
SONARIAC-924 ContextualMap should resolve objects by path

nils-werner-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualMap.java
```diff
@@ -50,6 +53,13 @@ public ContextualObject object(String name) {
       .orElse(ContextualObject.fromAbsent(ctx, name, this));
   }
 
+  public List<ContextualObject> objectsByPath(String path) {
+    return ArmTreeUtils.resolveProperties(path, tree).stream()
```
rudy-regazzoni-sonarsource, line 57: As `tree` can be `null` in this context and the method supports it, I would suggest adding the annotation `@Nullable` on both `resolveProperties` methods.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice improvement, please see my comment.

PR #764, opened 2023-06-28
SONARIAC-923 Add targetScope in File

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/FileImpl.java
```diff
@@ -39,6 +45,29 @@ public List<Tree> children() {
...
   }
 
+  @Override
+  public Scope targetScope() {
+    if (targetScope == null) {
+      return Scope.NOT_SET;
+    } else if (targetScope.value().endsWith("/managementGroupDeploymentTemplate.json#")) {
+      return Scope.MANAGEMENT_GROUP;
+    } else if (targetScope.value().endsWith("/deploymentTemplate.json#")) {
+      return Scope.RESOURCE_GROUP;
```
mstachniuk, line 55: Why `RESOURCE_GROUP`? Is `DEPLOYMENT` not a better choice?

mstachniuk (approved):
LGTM and please see my comments.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/FileImpl.java
rudy-regazzoni-sonarsource: This is coming from the [ARM documentation](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/syntax).
And I agree: the schema name poorly reflects this :sweat_smile:

PR #765, opened 2023-06-28
Rule S6387: ARM Azure role assignments that grant access to all resources of a subscription are security-sensitive - Detect only on base template

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SubscriptionRoleAssignmentCheck.java
```diff
@@ -0,0 +1,50 @@
...
+import org.sonar.iac.arm.tree.api.File;
+import org.sonar.iac.common.api.checks.SecondaryLocation;
+
+@Rule(key = "S6387")
+public class SubscriptionRoleAssignmentCheck extends AbstractArmResourceCheck {
+
+  private static final String MESSAGE = "Make sure assigning this role with a %s scope is safe here.";
+  private static final Map<File.Scope, String> SENSITIVE_SCOPE_WITH_NAME = Map.of(
+    File.Scope.SUBSCRIPTION, "Subscription",
+    File.Scope.MANAGEMENT_GROUP, "Management Group");
```
nils-werner-sonarsource, line 34: When you add the word "scope" to both strings. You can reduce the message and the secondary location message by the concat. Maybe it makes it a bit more readable.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SubscriptionRoleAssignmentCheck.java
```diff
@@ -0,0 +1,50 @@
...
+import org.sonar.check.Rule;
+import org.sonar.iac.arm.checkdsl.ContextualResource;
+import org.sonar.iac.arm.tree.api.File;
+import org.sonar.iac.common.api.checks.SecondaryLocation;
+
+@Rule(key = "S6387")
+public class SubscriptionRoleAssignmentCheck extends AbstractArmResourceCheck {
+
+  private static final String MESSAGE = "Make sure assigning this role with a %s scope is safe here.";
+  private static final Map<File.Scope, String> SENSITIVE_SCOPE_WITH_NAME = Map.of(
```
nils-werner-sonarsource, line 32: IMO a static import of Scope is also fine since it is not used in another context. I will make it more readable.

nils-werner-sonarsource (approved):
LGTM! I left two nitpicky comments. In addition, I had a look at the `FileImpl::targetScope` this can be improved IMO. It is not efficient nor easy to read. However, this is not related to this PR.

nils-werner-sonarsource:
```
  @Override
  public Scope targetScope() {
    if (targetScope == null) {
      return Scope.NOT_SET;
    }

    String scopeAsString = targetScope.value();
    String scopeSuffix = scopeAsString.substring(scopeAsString.lastIndexOf('/'));

    switch (scopeSuffix) {
      case "/managementGroupDeploymentTemplate.json#":
        return Scope.MANAGEMENT_GROUP;
      case "/deploymentTemplate.json#":
        return Scope.RESOURCE_GROUP;
      case "/subscriptionDeploymentTemplate.json#":
        return Scope.SUBSCRIPTION;
      case "/tenantDeploymentTemplate.json#":
        return Scope.TENANT;
      default:
        return Scope.UNKNOWN;
    }
  }
```

PR #766, opened 2023-06-28
SONARIAC-914 Rule S6388: ARM Using unencrypted cloud storages is security-sensitive (Omitting encryption property)

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM! Nice use of `objectsByPath`

PR #767, opened 2023-06-28
SONARIAC-772 Rule S6385: ARM Azure custom roles should not grant subscription Owner capabilities

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SubscriptionOwnerCapabilitiesCheck.java
```diff
@@ -0,0 +1,87 @@
...
+  private static Stream<Expression> findArbitraryActions(ContextualObject permissionObject) {
+    return permissionObject.list("actions").getItemIf(
+      action -> isValue(action, "*").isTrue());
+  }
+
+  private static boolean isSensitiveScope(Tree scope) {
+    boolean hasSensitiveCall = isValue(scope, "[managementGroup().id]").isTrue() ||
+      isValue(scope, "[subscription().id]").isTrue();
+    boolean referencesSensitiveScope = matchesValue(scope, s -> s.startsWith("/subscriptions/")).isTrue() &&
+      matchesValue(scope, s -> RESOURCE_GROUP_SCOPE_PATTERN.matcher(s).matches()).isFalse() ||
```
rudy-regazzoni-sonarsource, line 83: I would use the same approach as for the terraform rule, with this regex: `^/subscriptions/[^/]+/?$`
That way you don't need to perform two checks on the same string.

iac-extensions/arm/src/test/resources/checks/SubscriptionOwnerCapabilitiesCheck/Microsoft.Authorization_roleDefinitions/test.json
```diff
@@ -0,0 +1,54 @@
...
+      "name": "Compliant - specific actions are specified",
+      "properties": {
+        "permissions": [
+          {
+            "actions": ["Microsoft.Compute/*"],
+            "notActions": []
+          }
+        ],
+        "assignableScopes": [
+          "[subscription().id]"
```
rudy-regazzoni-sonarsource, line 17:
```suggestion
          "${type}"
```
You could replace this with the injected sensitive value directly.

iac-extensions/arm/src/test/resources/checks/SubscriptionOwnerCapabilitiesCheck/Microsoft.Authorization_roleDefinitions/test.json
```diff
@@ -0,0 +1,54 @@
...
+      "name": "Compliant - scope is limited to a resource group",
+      "properties": {
+        "permissions": [
+          {
+            "actions": ["*"],
+            "notActions": []
+          }
+        ],
+        "assignableScopes": [
+          "/subscriptions/b24988ac-6180-42a0-ab88-20f7382dd24c/resourceGroups/b24988ac-6180-42a0-ab88-20f7382dd24c"
```
rudy-regazzoni-sonarsource, line 33:
```suggestion
          "${type}"
```
Same here.

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my comments.

iac-extensions/arm/src/test/resources/checks/SubscriptionOwnerCapabilitiesCheck/Microsoft.Authorization_roleDefinitions/test.json
petertrr: Here that's a compliant example with insensitive scope, but the test is parameterized with different types of sensitive scopes.

PR #771, opened 2023-06-29
SONARIAC-807 Rule S6381: ARM Assigning high privileges Azure Resource…

mstachniuk:
… Manager built-in roles is security-sensitive


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
```diff
@@ -39,6 +39,14 @@ public static Predicate<Expression> isEqual(String targetString) {
...
   }
 
+  public static Predicate<Expression> contains(String targetString) {
+    return expr -> TextUtils.matchesValue(expr, containsPredicate(targetString)).isTrue();
+  }
+
+  private static Predicate<String> containsPredicate(String targetString) {
+    return predicate -> predicate.contains(targetString);
+  }
+
```
rudy-regazzoni-sonarsource, line 49:
```suggestion
  public static Predicate<Expression> contains(String targetString) {
    return expr -> TextUtils.matchesValue(expr, str -> str.contains(targetString)).isTrue();
  }
```
I would simplify it this way.

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my comment.

PR #774, opened 2023-06-30
SONARIAC-916 S6388: Omitting encryptionState or infrastructureEncryption

petertrr:
* Implemented part of the rule S6388
* Fixed the message according to the RSPEC


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #775, opened 2023-06-30
SONARIAC-913 S6288: Omitting or setting encryption to false

rudy-regazzoni-sonarsource:
Not sure what to do about the code smells.
For resource type, a possible solution would be to define an enum with the associated path and reference them everywhere, but this would add a layer of abstraction when reading the rule.


nils-werner-sonarsource (approved):
LGTM!

PR #785, opened 2023-07-05
SONARIAC-944 ARM Bicep support: create IDENTIFIER(name)

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepLexicalGrammar.java
```diff
@@ -45,13 +46,15 @@ public enum BicepLexicalGrammar implements GrammarRuleKey {
    * Statements
    */
   STATEMENT,
+  IDENTIFIER,
```
petertrr, line 49: Can identifier be a statement on its own? Maybe it should be moved to expressions?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepLexicalGrammar.java
```diff
@@ -69,7 +72,8 @@ public static LexerlessGrammarBuilder createGrammarBuilder() {
   }
 
   private static void punctuators(LexerlessGrammarBuilder b) {
-    b.rule(Punctuator.EQU).is(Punctuator.EQU.getValue());
+    Stream.of(Punctuator.EQU).forEach(
+      p -> b.rule(p).is(SPACING, p.getValue()).skip());
```
petertrr, line 76: Not quite a comment, but rather a question. Does this line mean that spaces surrounding equality signs will be attached to them in the AST? If so, shouldn't it be done for other tokens, e.g. for keywords?

petertrr (approved):
LGTM! Just a couple of questions

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepLexicalGrammar.java
mstachniuk: No, an identifier can't be a statement on its own.
Let's move it.

mstachniuk: IMO those comments don't have much value, sometimes it's hard to decide what is what.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepLexicalGrammar.java
mstachniuk: No, spaces will be not attached to AST. Usually, we don't need spacing in AST (exception: Docker).
For now, there is no need to do it for keywords - baby steps.

petertrr: So `skip()` basically means that the equality sign along with optional spaces will be a `SyntaxToken` and not a dedicated class of AST node?

PR #790, opened 2023-07-07
SONARIAC-934 Extending of LiteralValue in grammar should return the proper Expression implementation

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepLexicalGrammar.java
```diff
@@ -57,10 +57,21 @@ public enum BicepLexicalGrammar implements GrammarRuleKey {
   LITERAL_VALUE,
   ALPHA_NUMERAL_STRING,
   STRING_LITERAL,
```
petertrr, line 59: Is `STRING_LITERAL` still used anywhere?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -84,20 +87,36 @@ public Expression EXPRESSION() {
...
+  public NullLiteral NULL_LITERAL() {
+    return b.<NullLiteral>nonterminal(BicepLexicalGrammar.NULL_LITERAL).is(
+      f.nullLiteral(
+        b.token(BicepLexicalGrammar.NULL_LITERAL_VALUE)));
   }
 
   public StringLiteral STRING_LITERAL_VALUE() {
     return b.<StringLiteral>nonterminal().is(
-      f.stringLiteral(b.token(BicepLexicalGrammar.STRING_LITERAL)));
+      f.stringLiteral(b.token(BicepLexicalGrammar.STRING_LITERAL_VALUE)));
```
petertrr, line 119: What about single quotes? Should we include them as arguments in `f.stringLiteral`?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -84,20 +87,36 @@ public Expression EXPRESSION() {
...
+          b.token(BicepLexicalGrammar.FALSE_LITERAL_VALUE))));
+  }
+
+  public NullLiteral NULL_LITERAL() {
+    return b.<NullLiteral>nonterminal(BicepLexicalGrammar.NULL_LITERAL).is(
+      f.nullLiteral(
+        b.token(BicepLexicalGrammar.NULL_LITERAL_VALUE)));
   }
 
   public StringLiteral STRING_LITERAL_VALUE() {
```
petertrr, line 117: Should it be `STRING_LITERAL` or `STRING_LITERAL_VALUE`?

On a side note, I feel like naming for strings is a bit confusing, as the grammar doesn't have `stringLiteral` and `stringValue`, but rather has `stringComplete` as a variant of `interpString` without any interpolation. In this sense our `STRING_LITERAL_VALUE` is `stringComplete`

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/BooleanLiteralImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+  private final SyntaxToken token;
+
+  public BooleanLiteralImpl(SyntaxToken token) {
+    this.token = token;
+  }
+
+  @Override
+  public boolean value() {
+    return "true".equals(token.value());
```
petertrr, line 38: Is it worth storing this value in a field, both here and in `NumericLiteralImpl`, so that it isn't recalculated on every access?

petertrr (approved):
LGTM, please see my comments

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
rudy-regazzoni-sonarsource: This part is more like temporary support of string in our `Expression`.
I expect it to be progressively removed as we further implement elements such as `expression` or `interpString`.
I added a comment in this sense.
The question you are raising should definitely be tackled at those points.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
rudy-regazzoni-sonarsource: Good point, I renamed it to `STRING_LITERAL`.
I agree, the name is not great and is not reflected in the grammar. It could be interpreted as an equivalent of `stringComplete`, but as mentioned earlier this is more like a temp approach to support string, that will be removed later in favor of a more complete version aligned to the Bicep grammar.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/BooleanLiteralImpl.java
rudy-regazzoni-sonarsource: That's a fair point.
Currently, we usually choose this basic approach: the method simply returns the value we are looking for, computing it every time.
I believe we do it for readability: it's easier to see what the method does instead of going through an additional layer of cache.
I mostly agree with this approach, as I don't expect a big win by caching the value: although it may happen that we call it a few times in our checks, I'm not expecting this to be more than 10 or at worst case 100 times, and the computation time wouldn't be that big.

This is also something you will find in the `children()` method that you can find in some tree elements, where we every time rebuild the list (See for example in the class HealthCheckInstructionImpl in the docker parser). We could theoretically replace it with a unique read-only List that would be computed once.

PR #795, opened 2023-07-07
Arm Sensor should analyze bicep files & setup Ruling tests with ARM bicep

nils-werner-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM! Will be great to see all those parse error been fixed progressively!

PR #796, opened 2023-07-10
SONARIAC-947 ARM Bicep support: add support for functionCall

jonas-wielage-sonarsource:


iac-common/src/test/java/org/sonar/iac/common/AbstractTestCommonSyntaxToken.java
```diff
@@ -17,39 +17,33 @@
...
+  public TextRange textRange() {
+    return null;
   }
 
   @Override
-  public List<SyntaxToken> separators() {
-    return separators;
+  public String value() {
+    return "";
   }
```
rudy-regazzoni-sonarsource, line 43: I would remove those methods, to me they are invalid and must be overridden in any case, so by removing them from there we ensure that any implementation done later does not forget them.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice improvement regarding the generalization of `SeparatedList`! Please see my few comments.

iac-common/src/main/java/org/sonar/iac/common/api/tree/CommonSyntaxToken.java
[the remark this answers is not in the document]
jonas-wielage-sonarsource: @rudy-regazzoni-sonarsource Good catch, I was wondering how i could remove this!

iac-common/src/test/java/org/sonar/iac/common/AbstractTestCommonSyntaxToken.java
jonas-wielage-sonarsource: I based my implementation on the TestTree, but I agree and will change it.

PR #797, opened 2023-07-10
SONARIAC-956 ARM Bicep support: create outputDecl - basic implementation

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -80,6 +82,26 @@ public TypeDeclaration TYPE_DECLARATION() {
...
+          IDENTIFIER(),
+          IDENTIFIER(),
+          b.token(Punctuator.EQU),
+          EXPRESSION()),
+        f.outputDeclaration(
+          b.token(BicepKeyword.OUTPUT),
+          IDENTIFIER(),
+          b.token(BicepKeyword.RESOURCE),
+          // TODO SONARIAC-958 Put in place interpString
+          STRING_LITERAL(),
```
petertrr, line 100: I think you can already use the rule which is called `INTERPOLATED_STRING_TYPE` in the current master, which is simply a quoted string literal.

petertrr, line 100: I think you can already use the rule which is called `INTERPOLATED_STRING_TYPE` in the current master, which is simply a quoted string literal.

petertrr (approved):
LGTM!

PR #798, opened 2023-07-10
SONARIAC-965 ARM Bicep support: create ambientTypeReference

mstachniuk:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/AmbientTypeReferenceTest.java
```diff
@@ -0,0 +1,47 @@
...
+      .matches("int")
+      .matches("bool")
+      .matches("array")
+      .matches("object")
+
+      .notMatches("String")
+      .notMatches("STRING")
+      .notMatches("foo")
+      .notMatches("123")
+      .notMatches("variable a = 10")
```
rudy-regazzoni-sonarsource, line 43: I would also add a usecase like `stringeee` to ensure that we don't match it.

rudy-regazzoni-sonarsource (changes requested):
This seems a bit over-engineered just to match 5 different strings.
Here you have created 5 identical classes/interfaces, for which their only purpose is to have a different Kind.
The test classes enforce this idea because they test the same things.

I would simplify it by only keeping the `AmbientTypeReference` class/interface.
The token value could be enforced by a regex like `(string|int|bool|array|object)`, with probably a look behind to ensure it's not part of a bigger word.
Also, we could think about a method/enum in the interface to identify the type.

WDYT?
Happy to discuss it.

mstachniuk:
> Here you have created 5 identical classes...

Yes, I know, I wasn't sure what is better in this case, I just followed how it's done for `literalValue -> NUMBER | "true" | "false" | "null"`.
Sure let's simplify it to `AmbientTypeReference`

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/AmbientTypeReferenceImplTest.java
rudy-regazzoni-sonarsource, line 1: I would add another method that parses `intelligent` (or any word starting with a valid ambient type reference) but trigger a parser error.
```
@Test
void shouldNotParseBiggerWord() {
  assertThatThrownBy(() -> parse("intelligent", BicepLexicalGrammar.AMBIENT_TYPE_REFERENCE))
    .isInstanceOf(RecognitionException.class)
    .hasMessageStartingWith("Parse error at line 1 column 1");
}
```
Currently, your tests are valid, but this does not show that we should not parse such word as a `ambientTypeReference`.
The `notMatches` only work because it detects that this doesn't match the full string as it, not because it triggers a parser error.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice improvements! I added a few more modifications to consider.
Please also check the code smell, I believe I saw duplicate import to be fixed.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/ArmAssertions.java
```diff
@@ -68,4 +74,16 @@ public static ObjectExpressionAssert assertThat(ObjectExpression actual) {
...
+    return new RuleAssert(actual);
+  }
+
+  public static ParserAssert assertThat(GrammarRuleKey rule) {
+    return new ParserAssert(BicepParser.create(rule));
+  }
+
+  public static org.assertj.core.api.AbstractComparableAssert<?, ArmTree.Kind> assertThat(ArmTree.Kind kind) {
+    return Assertions.assertThat(kind);
+  }
```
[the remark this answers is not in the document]
mstachniuk: Fixed only returned type. Static import of `Assertions` will cause: 
`Method 'assertThat()' recurses infinitely, and can only end by throwing an exception`

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/AmbientTypeReferenceImplTest.java
mstachniuk: Good catch

PR #799, opened 2023-07-10
SONARIAC-858 ARM Bicep support: create importDecl

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ImportDeclarationImpl.java
```diff
@@ -0,0 +1,67 @@
...
+  private final SyntaxToken newLine;
+
+  public ImportDeclarationImpl(SyntaxToken keyword, InterpolatedString specification, Optional<ImportWithClause> withClause, Optional<ImportAsClause> asClause,
+    SyntaxToken newLine) {
+    this.keyword = keyword;
+    this.specification = specification;
+    this.withClause = withClause.orNull();
+    this.asClause = asClause.orNull();
+    this.newLine = newLine;
+  }
```
rudy-regazzoni-sonarsource, line 51: We usually prefer to pass `@Nullable` parameter instead of `Optional`.
Some elements of the [why here](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments), although everything can be arguable.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ImportDeclarationImplTest.java
```diff
@@ -0,0 +1,55 @@
...
+      .matches("import 'foo'")
+      .matches("import 'foo' as bar")
+      .matches("import 'foo' with {}")
+      .matches("import 'foo' with {} as bar")
+
+      .notMatches("import")
+      .notMatches("import with {}")
+      .notMatches("import as bar")
+      .notMatches("import 'foo' as");
+  }
```
rudy-regazzoni-sonarsource, line 42: I would add  a few more tests:
```
import 'foo' as bar with {}
import 'foo' with as bar
```

rudy-regazzoni-sonarsource (approved):

nils-werner-sonarsource:
@petertrr and @rudy-regazzoni-sonarsource why no interface describing getters were added for the `ImportAsClause` and `ImportWithClause`?

rudy-regazzoni-sonarsource:
> @petertrr and @rudy-regazzoni-sonarsource why no interface describing getters were added for the `ImportAsClause` and `ImportWithClause`?

Good question... I don't see any reasons honestly, I would say it's a miss, I will create a ticket to fix and take care of it.

PR #800, opened 2023-07-10
SONARIAC-961 ARM Bicep support: create parameterDecl - basic implementation

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -111,6 +114,26 @@ public TargetScopeDeclaration TARGET_SCOPE_DECLARATION() {
...
+  public ParameterDeclaration PARAMETER_DECLARATION() {
+    return b.<ParameterDeclaration>nonterminal(BicepLexicalGrammar.PARAMETER_DECLARATION).is(
+      b.firstOf(
+        f.parameterDeclaration(
+          b.token(BicepKeyword.PARAMETER),
+          IDENTIFIER(),
+          b.token(BicepKeyword.RESOURCE),
+          INTERPOLATED_STRING_TYPE(),
+          b.optional(b.token(Punctuator.EQU)),
+          b.optional(EXPRESSION())),
```
mstachniuk, line 127: 1. The `EQU` and `EXPRESSION()` as a whole need to be optional. Currently, the following input will be accepted:
`parameter id resource iterString =` (with equals but without expression at the end).
2. Extraction to separate method will help - just to be as much as possible with original grammar, and reuse `parameterDefaultValue` also in lines 133-134

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ParameterDeclarationImpl.java
```diff
@@ -0,0 +1,156 @@
...
+    this.keyword = keyword;
+    this.name = name;
+    this.typeExpression = typeExpression;
+    this.equ = equ;
+    this.defaultValue = defaultValue;
+    this.resource = null;
+    this.typeInterp = null;
+  }
+
+  public ParameterDeclarationImpl(SyntaxToken keyword, Identifier name, @Nullable SyntaxToken resource, @Nullable InterpolatedString typeInterp, @Nullable SyntaxToken equ,
```
mstachniuk, line 65: Nitpic. Lines 55 and 65-66 are long, maybe they will look better when each argument will be in a separate line?

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/utils/Assertions.java
```diff
@@ -53,7 +53,7 @@ private void parseTillEof(String input) {
       TokenLocation loc = new TokenLocation(1, 0, input);
       if (!tree.textRange().end().equals(new TextPointer(loc.endLine(), loc.endLineOffset()))) {
         throw new RecognitionException(
-          0, "Did not match till EOF, but till line " + tree.textRange().end().line());
+          0, "Did not match till EOF, but till line " + tree.textRange().end().line() + " line offset: " + tree.textRange().end().lineOffset());
```
mstachniuk, line 56: I like it.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ParameterDeclarationImplTest.java
```diff
@@ -0,0 +1,95 @@
...
+      .matches("parameter myParam int")
+      .matches("parameter myParam int = 123")
+      .matches("parameter myParam int=123")
+      .matches("parameter myParam resource 'interpolated string'")
+      .matches("parameter myParam resource 'interpolated string' = 123")
+      // invalid code that it still accepted by our parser
+      .matches("parameter myParam int 123")
+      .matches("parameter myParam resource interpolatedString")
+
+      .notMatches("parameter")
```
mstachniuk, line 49: please also add 
`.notMatches("parameter myParam resource 'interpolated string' = ")`

mstachniuk (approved):

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
rudy-regazzoni-sonarsource: Not necessary, I got the same use case with Docker and Nils suggested that we don't need to make them whole (see `Flag` from Docker for example).
We tend to enforce the grammar/syntax, but we are not writing a validator :smile:
When writing our parser we suppose that the provided code is valid, so we don't need to ensure/enforce this. There would be no use case where a valid code would be parsed into an invalid AST or trigger a parse error here.

I wanted to avoid introducing `parameterDefaultValue` grammar element, from my opinion it was just adding a layer of complexity (+ method in grammar + classes to handle it + tests) for not much added value.

WDYT?

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/utils/Assertions.java
rudy-regazzoni-sonarsource: No much pride on this, I just copy pasted it from Docker :grin:

PR #806, opened 2023-07-11
SONARIAC-866 ARM Bicep support: add 'existing' boolean in resources

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/ResourceDeclaration.java
```diff
@@ -31,4 +31,8 @@ public interface ResourceDeclaration extends Statement, HasProperties {
   StringLiteral type();
 
   List<Property> properties();
+
+  default boolean existing() {
+    return false;
```
petertrr, line 36: There is a way to express the notion of `existing` for JSON: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/conditional-resource-deployment#new-or-existing-resource
But I'm not sure if it makes sense to support a piece of logic that requires that much evaluation

mstachniuk: Please create a ticket for that in the backlog to not forget. The implementation should not be so hard for JSON, but for sure it's not a priority for now.

mstachniuk (approved):

PR #810, opened 2023-07-11
SONARIAC-973 ARM Bicep add ifCondition to resourceDecl

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -188,7 +188,9 @@ public ResourceDeclaration RESOURCE_DECLARATION() {
         INTERPOLATED_STRING(),
         b.optional(b.token(BicepKeyword.EXISTING)),
         b.token(Punctuator.EQU),
-        OBJECT_EXPRESSION(),
+        b.firstOf(
+          OBJECT_EXPRESSION(),
+          IF_EXPRESSION()),
```
petertrr, line 193: We also already have `FOR_EXPRESSION`, should it be supported in here in a separate PR or in this one as well?

petertrr (approved):
LGTM!

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
rudy-regazzoni-sonarsource: That's a fair point, I actually tried to implement both `IF_EXPRESSION` and `FOR_EXPRESSION`, but I couldn't make it work properly as there was something missing on `FOR_EXPRESSION` part, it would have required an additional modification later.
So in the end I didn't want to partially do it for `ForExpression` and just kept separate ticket :smile:

PR #786, opened 2023-07-06
SONARIAC-935 ARM Bicep support: create simplified resourceDecl

mstachniuk:

PR #815, opened 2023-07-12
SONARIAC-968 ARM Bicep support: create tupleType

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -371,6 +373,24 @@ public UnaryOperator UNARY_OPERATOR() {
...
+        b.token(Punctuator.RBRACKET)));
+  }
+
+  public TupleItem TUPLE_ITEM() {
+    return b.<TupleItem>nonterminal(BicepLexicalGrammar.TUPLE_ITEM).is(
+      f.tupleItem(
+        b.zeroOrMore(DECORATOR()),
+        // TODO replace by typeExpression in SONARIAC-969 ARM Bicep support: create typeExpression
+        STRING_LITERAL(),
+        b.token(BicepLexicalGrammar.EOL)));
```
petertrr, line 391: In another ticket I also have this `NL+` structure and I attempted to describe it like this:
https://github.com/SonarSource/sonar-iac/blob/0628ffd793bcd7da0fadfcef7978b47ba3847792/iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepLexicalGrammar.java#L136
because I believe there could be multiple EOLs and leading whitespace. WDYT?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -371,6 +373,24 @@ public UnaryOperator UNARY_OPERATOR() {
       f.unaryOperator(b.token(BicepLexicalGrammar.UNARY_OPERATOR_VALUE)));
   }
 
+  public TupleType TUPLE_TYPE() {
+    return b.<TupleType>nonterminal(BicepLexicalGrammar.TUPLE_TYPE).is(
+      f.tupleType(
+        b.token(Punctuator.LBRACKET),
+        b.zeroOrMore(
```
petertrr, line 380: I think there should always be a newline before the first element, but I can't find proper examples in docs or samples

petertrr (approved):
LGTM, see my suggestions below

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
mstachniuk: We should try to avoid spacing and new lines as tokens in our grammar, see: https://github.com/SonarSource/sonar-iac/pull/786#discussion_r1255489954 
All punctators, keywords, and literal values allow leading space.
In most cases, a new line can be also skipped (I'm not sure if it is the case here). 
And as a remainder: we don't implement a language validator but a parser that expects valid files as input.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
mstachniuk: Yes, there should be a new line before the first element. Important is that the current implementation allows that, even if it's not implicity defined in the grammar. See test cases, the `"[typeExpr]" ` is not accepted (but it will be also not a big issue to parse it as well IMO).

PR #818, opened 2023-07-13
SONARIAC-988 Put in place Expression skeleton

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
```diff
@@ -138,7 +138,10 @@ public NumericLiteral toNumericLiteral(PropertyTree property) {
       throw convertError(property, value, NumericLiteral.class.getSimpleName(), "ScalarTree.Style.PLAIN");
     }
     try {
-      return new NumericLiteralImpl(Float.parseFloat(value.value()), value.metadata());
+      // for validation
+      Double.parseDouble(value.value());
```
nils-werner-sonarsource, line 142: Is there no smarter validation? Why does the grammar is not able to validate this? Why the factory should try to create a numericLiteral tree which is not valid?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -391,6 +391,36 @@ public TupleItem TUPLE_ITEM() {
         b.token(BicepLexicalGrammar.EOL)));
   }
 
+  public Expression EXPRESSION() {
+    return b.<Expression>nonterminal(BicepLexicalGrammar.EXPRESSION).is(
+      f.expression(
```
nils-werner-sonarsource, line 396: Why this is calling a factory method and not a `firstOf`?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/NumericLiteralImpl.java
```diff
@@ -34,12 +34,17 @@ public NumericLiteralImpl(SyntaxToken token) {
   }
 
   @Override
-  public double value() {
+  public double doubleValue() {
```
nils-werner-sonarsource, line 37: I would prefer `asDouble()`

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmBaseConverter.java
jonas-wielage-sonarsource: Well the way you are supposed to do it (based on documentation) is to copy / paste a giant regex into your code and validate strings with it.
I'm not sure that is better.
This is inside the ArmBaseConverter for Json, so there is no grammar.

PR #820, opened 2023-07-13
SONARIAC-990 Implement Relational Expression

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -408,7 +408,18 @@ public Expression EQUALITY_EXPRESSION() {
 
   public Expression RELATIONAL_EXPRESSION() {
     return b.<Expression>nonterminal(BicepLexicalGrammar.RELATIONAL_EXPRESSION).is(
-      b.firstOf(ADDITIVE_EXPRESSION()));
+      b.firstOf(
```
petertrr, line 411: Should there actually be `firstOf`, if the case with only one `ADDITIVE_EXPRESSION` could be covered by `zeroOrMore` in `relationalExpression` instead of `oneOrMore`? Or is it to make `TreeFactory#relationalExpression` less complicated?

petertrr (approved):
LGTM! Please see one question

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
rudy-regazzoni-sonarsource: That's a good question! Here is the reasoning we had with Jonas.
In our initial commit from our squad programming session, we had this `firstOf`, which would represent this grammar:
```
multiplicativeExpression -> 
  unaryExpression (("*" | "/" | "%") unaryExpression)* |
  unaryExpression
```
It was actually never matching the second part `unaryExpression`, because the first expression with the zeroOrMore part would match it first.

So at first, I removed it, which was giving this grammar:
```
multiplicativeExpression -> 
  unaryExpression (("*" | "/" | "%") unaryExpression)*
```
But as you saw in the `TreeFactory` class, this would give us a multiplicativeExpression method which would either return an Expression or a MultiplicativeExpression of the presence of the optional second part.
So part of the grammar logic was moved to the `TreeFactory` class, which is not great.

To fix this, we put back this `firstOf` and replaced the `zeroOrMore` with a `oneOrMore`, giving us this grammar:
```
multiplicativeExpression -> 
  unaryExpression (("*" | "/" | "%") unaryExpression)+ |
  unaryExpression
```
(notice the `+` instead of the `*`)

So now we easily see the two logic:
- we first try to greedy match a `multiplicativeExpression` with an operator (`*/%`)
- if we cannot match it, then we rely on `unaryExpression` alone.

Hope this is clear, let me know if you want to discuss it.

PR #821, opened 2023-07-13
SONARIAC-991 Implement Equality Expression

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -447,7 +447,16 @@ public Expression BINARY_EXPRESSION() {
...
+      b.firstOf(
+        f.equalityExpression(
+          RELATIONAL_EXPRESSION(),
+          b.oneOrMore(
+            f.tuple(
+              b.firstOf(
+                b.token(Punctuator.EQUAL),
+                b.token(Punctuator.NOT_EQUAL)),
+              RELATIONAL_EXPRESSION()))),
+        RELATIONAL_EXPRESSION()));
```
mstachniuk, line 459: Why not something like this:

```java
f.equalityExpression(
  RELATIONAL_EXPRESSION(),
  b.zeroOrMore(
  f.tuple(
  b.firstOf(
  b.token(Punctuator.EQUAL),
  b.token(Punctuator.NOT_EQUAL)),
  RELATIONAL_EXPRESSION())
  )
  )
```

There is common prefix. Later in `f.equalityExpression()` if the second argument is empty you can just return first argument. Now the SSLR needs to evaluate `RELATIONAL_EXPRESSION()` twice.

mstachniuk (approved):
LGTM and please see my comment. If my suggestion is not applicable, please just describe why and merge.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
rudy-regazzoni-sonarsource: Yes, that's true. We discussed it with Jonas and preferred the current solution.

I initially implemented the way you mentioned, but it is like having part of the grammar logic in the `TreeFactory`.
It's also kind of strange to have the method `f.equalityExpression()` either do nothing or build a proper `EqualityExpression` object.

The `firstOf()` approach seems cleaner to us: we first try a greedy match and build an `EqualityExpression` out of it, and if it doesn't work, we pass it to the `RELATIONAL_EXPRESSION` grammar. This is something we can read and understand directly from the grammar, no need to check what magic happens on the TreeFactory side.

I agree that it would probably require SSLR more evaluation that way, but can't tell at which level.
@nils-werner-sonarsource WDYT? It may be a fair concern, as we can expect several expression per files, and this logic would be applied at each level: `expression`, `binaryExpression`, `equalityExpression`, `relationalExpression`, `additiveExpression`, `multiplicativeExpression`, `unaryExpression` and `memberExpression`.

nils-werner-sonarsource: I see both sides. I prefer to have as much logic as possible in the grammar and not in the factory. However, this comes at some performance cost. For me it is not important which solution you choose, it is more important that the approach is always the same in the parser. Mixing approaches will lead to confusion in the future.

nils-werner-sonarsource: Not a blocker. However, should be discussed with the team.

PR #822, opened 2023-07-13
SONARIAC-992 Implement Binary Expression

rudy-regazzoni-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/parser/grammar/Punctuator.java
```diff
@@ -36,6 +36,7 @@ public enum Punctuator implements GrammarRuleKey {
   RPARENTHESIS(")"),
   OR("||"),
   AND("&&"),
+  COALESCE("??"),
```
jonas-wielage-sonarsource, line 39: TIL that this has a name...

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/TreeFactory.java
```diff
@@ -400,4 +402,8 @@ public RelationalExpression relationalExpression(Expression expression, List<Tup
   public EqualityExpression equalityExpression(Expression expression, List<Tuple<SyntaxToken, Expression>> list) {
     return new EqualityExpressionImpl(SeparatedListImpl.separatedList(expression, list));
   }
+
+  public BinaryExpression binaryExpression(Expression expression, List<Tuple<SyntaxToken, Expression>> list) {
+    return new BinaryExpressionImpl(SeparatedListImpl.separatedList(expression, list));
```
jonas-wielage-sonarsource, line 407: I would add a static import for this.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/expression/BinaryExpressionImpl.java
```diff
@@ -0,0 +1,32 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.arm.tree.impl.bicep.expression;
+
+import org.sonar.iac.arm.tree.api.Expression;
+import org.sonar.iac.arm.tree.api.bicep.SyntaxToken;
+import org.sonar.iac.arm.tree.api.bicep.expression.BinaryExpression;
+import org.sonar.iac.common.api.tree.SeparatedList;
+
+public class BinaryExpressionImpl extends BinaryOperationExpressionImpl implements BinaryExpression {
```
jonas-wielage-sonarsource, line 27: After thinking about it, IMO I would do the following:

1. Remove BinaryOperationExpression 
2. Rename BinaryOperationExpressionImpl to AbstractBinaryOperation

I don't see that much of a value of the BinaryOperationExpression interface.

WDYT?

jonas-wielage-sonarsource (approved):
LGTM! Please see my comments.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/expression/BinaryExpressionImpl.java
rudy-regazzoni-sonarsource: I agree and tried to remove the interface, but it is needed to make other interface extends from it.
Since we are mostly working with interfaces, we cannot remove it otherwise we won't have access to the `SeparatedList`.
I still kept the renaming which is much better.

iac-common/src/main/java/org/sonar/iac/common/parser/grammar/Punctuator.java
rudy-regazzoni-sonarsource: To be honest I found this yesterday in Bicep azure documentation :sweat_smile:

PR #827, opened 2023-07-14
SONARIAC-978 Extend decoratorExpression

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/bicep/Decorator.java
```diff
@@ -20,8 +20,9 @@
 package org.sonar.iac.arm.tree.api.bicep;
 
 import org.sonar.iac.arm.tree.api.ArmTree;
+import org.sonar.iac.arm.tree.api.Expression;
 
 public interface Decorator extends ArmTree {
 
-  FunctionCall functionCall();
+  Expression expression();
```
petertrr, line 27: I think for later when we'll need to analyze decorators and retrieve some values, it would be useful to have an interface `DecoratorExpression` with an accessor for function call

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/TreeFactory.java
```diff
@@ -255,29 +255,29 @@ public IfExpression ifExpression(SyntaxToken keyword, ParenthesizedExpression co
     return new IfExpressionImpl(keyword, condition, object);
   }
 
-  public Expression memberExpression(Expression value, List<MemberExpression> memberExpressionComponent) {
+  public MemberExpression memberExpression(Expression value, List<MemberExpression> memberExpressionComponent) {
```
petertrr, line 258:
```suggestion
  public MemberExpression memberExpression(Expression value, List<MemberExpression> memberExpressionComponents) {
```
I think that having variables for collections named as plural somewhat improves readability, WDYT?

petertrr (approved):
LGTM! See my suggestions below

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/TreeFactory.java
jonas-wielage-sonarsource: Definitely

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/bicep/Decorator.java
jonas-wielage-sonarsource: As we are not sure if rules would really use it, for now I would leave it as it is. Can be later added when we have Checks than would benefit from it IMO

PR #829, opened 2023-07-17
SONARIAC-949 Put in place typeExpression instead of StringLiteral for…

mstachniuk:
… typeDecl


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #831, opened 2023-07-17
SONARIAC-994 Replace Primary Expression with 'root' Expression

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM and I would expect more changes in other test files. In Grammar, there are changes in 11 types, but only 3 tests are modified.

PR #833, opened 2023-07-18
SONARIAC-986 Bicep Arrays: support comma-separated arrays

rudy-regazzoni-sonarsource:
Bicep array value can either be specified alone on their line, or on a single line with comma separator.
```
array1 = [
  'a'
  'b'
]
array2 = ['c', 'd']
```
It also supports to mix those forms:
```
array_mix = [
  'a'
  'b'
  'c', 'd'
]
```

To enable this, I put SeparatedList in our `ArrayExpressionImpl`.
I modified `SeparatedList` to allow to store nullable value, which in the end get skipped when retrieving elements with separators.
This is done in order to keep the right order when retrieving the list of elements and separators in `children()` methods.
In the above last example, we would have those lists:
```
// list of element
['a', 'b', 'c', 'd']
// list of separator
[null, null, ',']
// when calling 'elementsAndSeparators'
elements = ['a', 'b', 'c', ',', 'd']
```

### DIFF
--- a/iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ArrayExpressionImplTest.java
+++ b/iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ArrayExpressionImplTest.java
@@ -25,7 +25,10 @@ import org.sonar.iac.arm.ArmAssertions;
 import org.sonar.iac.arm.parser.bicep.BicepLexicalGrammar;
 import org.sonar.iac.arm.tree.api.ArmTree;
 import org.sonar.iac.arm.tree.api.ArrayExpression;
+import org.sonar.iac.arm.tree.api.bicep.StringComplete;
 
+import static org.fest.assertions.Assertions.assertThat;
+import static org.sonar.iac.arm.ArmTestUtils.recursiveTransformationOfTreeChildrenToStrings;
 import static org.sonar.iac.common.testing.IacTestUtils.code;
 
 class ArrayExpressionImplTest extends BicepTreeModelTest {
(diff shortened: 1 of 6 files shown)


nils-werner-sonarsource (changes requested):
Please do not locate TreeFactory-related logic to tree implementations.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ArrayExpressionImplTest.java
```diff
@@ -47,13 +52,33 @@ void shouldParseMultilineArrays() {
...
+
+    assertThat(tree.getKind()).isEqualTo(ArmTree.Kind.ARRAY_EXPRESSION);
+    assertThat(tree.elements()).hasSize(3);
+    StringComplete a = (StringComplete) tree.elements().get(0);
+    StringComplete b = (StringComplete) tree.elements().get(1);
+    StringComplete c = (StringComplete) tree.elements().get(2);
+    ArmAssertions.assertThat(a.content()).hasValue("a");
+    ArmAssertions.assertThat(b.content()).hasValue("b");
+    ArmAssertions.assertThat(c.content()).hasValue("c");
+
```
mstachniuk, line 80: please add assertions for `separators()`

mstachniuk (approved):
LGTM

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ArrayExpressionImplTest.java
rudy-regazzoni-sonarsource: Not possible here: `ArrayExpression` does not provide access to the `SeparatedList` object.
I added another test in `SeparatedListImplTest` about it.

PR #835, opened 2023-07-19
SONARIAC-1000 ARM Bicep primaryExpression should accept string complete instead of alpha numeral string

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM

PR #836, opened 2023-07-19
SONARIAC-1011 Bicep grammar: remove EOL, align expressions orders

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepKeyword.java
```diff
@@ -38,7 +38,10 @@ public enum BicepKeyword implements GrammarRuleKey {
   IMPORT("import"),
   WITH("with"),
   AS("as"),
-  MODULE("module");
+  MODULE("module"),
+  TRUE("true"),
+  FALSE("false"),
+  NULL("null");
```
mstachniuk, line 44: Is it used? I don't see it in this PR, we have TRUE_LITERAL_VALUE and similar to detect `literalValue`

mstachniuk (approved):

mstachniuk:
> Couldn't fully respect the order from the original Bicep grammar. `InterString` would match `MultilineString` first and then fail: `'''multiline string'''` and `'interpolated string'`
> 
> We could prevent this by using some negative lookahead regex, but not sure if it would be worth it.

For me, it's ok to have them in a different order. The official Bicep Grammar looks like it is not "production ready" and fully correct.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepKeyword.java
rudy-regazzoni-sonarsource: It is not used directly, but by adding them in `BicepKeyword`, they get injected into the regex exception of `Identifier`, so that we avoid having them be considered as valid `Identifier` instead of a literal value.
Although, now that I see it, I don't think `true`/`false`/`null` follow the definition of a keyword.
I will remove them from the list and instead use those literals in the regex build method.

PR #838, opened 2023-07-19
SONARIAC-1005 Bicep: Fix ResourceDeclarationImpl properties() to retu…

mstachniuk:
…rn elements inside properties field

### DIFF
--- a/iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImpl.java
+++ b/iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImpl.java
@@ -37,6 +37,7 @@ import org.sonar.iac.arm.tree.api.bicep.IfCondition;
 import org.sonar.iac.arm.tree.api.bicep.InterpolatedString;
 import org.sonar.iac.arm.tree.api.bicep.SyntaxToken;
 import org.sonar.iac.arm.tree.impl.AbstractArmTreeImpl;
+import org.sonar.iac.common.api.tree.TextTree;
 import org.sonar.iac.common.api.tree.Tree;
 import org.sonar.iac.common.api.tree.impl.TextRange;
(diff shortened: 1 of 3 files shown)


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImpl.java
```diff
@@ -141,7 +142,11 @@ public StringLiteral type() {
   @Override
   public List<Property> properties() {
     if (body.is(Kind.OBJECT_EXPRESSION)) {
-      return ((ObjectExpression) body).properties();
+      return ((ObjectExpression) body).properties().stream()
+        .filter(propertyTree -> "properties".equals(((TextTree) propertyTree.key()).value()))
+        .map(p -> (Collections.<Property>unmodifiableList(((ObjectExpression) ((Property) p).value()).properties())))
+        .findFirst()
+        .orElse(List.of());
```
rudy-regazzoni-sonarsource, line 149: I believe this behavior should also be applied In case of `ForExpression` and `IfCondition`.
In both cases, the object body represents a resource, and so we should provide the `properties` field.

Example from Azure bicep documentation:
```
resource vmName_omsOnboarding 'Microsoft.Compute/virtualMachines/extensions@2023-03-01' = if (!empty(logAnalytics)) {
  name: '${vmName}/omsOnboarding'
  location: location
  properties: {
    publisher: 'Microsoft.EnterpriseCloud.Monitoring'
    type: 'MicrosoftMonitoringAgent'
    typeHandlerVersion: '1.0'
    autoUpgradeMinorVersion: true
    /* ... */
  }
}
```

rudy-regazzoni-sonarsource (changes requested):
The code is good to me, only thing that is missing is to apply the same for for/if resource, please see my comment.

rudy-regazzoni-sonarsource (approved):
LGTM!

PR #839, opened 2023-07-20
SONARIAC-1017 ARM Bicep: add support for nested resources

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImpl.java
```diff
@@ -115,11 +116,12 @@ public StringLiteral version() {
         return new StringLiteralImpl(token);
       }
     }
-    throw new UnsupportedOperationException("The typeAndVersion field should contains `type@version` value, but it's " + text);
+    return typeAndVersion;
```
rudy-regazzoni-sonarsource, line 119: I had to remove this exception throw because it currently crashes the parser and prevents it from parsing the file + all projects.
It is valid in bicep to not have a version in a resource declaration apparently, tickets created to fix this: [SONARIAC-1018](https://sonarsource.atlassian.net/browse/SONARIAC-1018) and [SONARIAC-1019](https://sonarsource.atlassian.net/browse/SONARIAC-1019).

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/ObjectExpressionImpl.java
```diff
@@ -42,6 +43,11 @@ public List<Property> properties() {
     return Collections.unmodifiableList(properties);
   }
 
+  @Override
+  public List<ResourceDeclaration> nestedResources() {
+    return Collections.emptyList();
```
mstachniuk, line 48: Why it returns empty list? I would expect some logic here to return nested resources 🤔

mstachniuk (approved):
LGTM and please see my question

PR #840, opened 2023-07-20
SONARIAC-1020 ARM Bicep: support operator :: for output declaration

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/MemberExpressionImplTest.java
```diff
@@ -137,6 +143,21 @@ void shouldParseMemberExpressionWithIdentifier() {
     assertThat(((TextTree) tree.children().get(1)).value()).isEqualTo(":");
   }
 
+  @Test
+  void shouldParseMemberExpressionDoubleColonWithIdentifier() {
+    String code = code("virtualNetwork::subnet1");
```
mstachniuk, line 148: Nitpic. Calling a method `code()` doesn't have sense here ;-) (and in other lines in this file too)

mstachniuk (approved):

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/MemberExpressionImplTest.java
rudy-regazzoni-sonarsource: Good catch :sweat_smile: I made a small refactoring for this class, it was also using a manually created parser, removed it.

PR #841, opened 2023-07-20
SONARIAC-1021 ARM bicep: Add support to nested child resources

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImplTest.java
```diff
@@ -210,4 +210,21 @@ void shouldProvideEmptyPropertiesWithForBodyNotObject() {
     ResourceDeclaration tree = parse(code, BicepLexicalGrammar.RESOURCE_DECLARATION);
     assertThat(tree.properties()).isEmpty();
   }
+
+  @Test
+  void shouldParseNestedResource() {
```
mstachniuk, line 215: Please also take into consideration that child resources can have only a part of the type (and taking a full type by concatenation of parent and child type). https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/child-resource-name-type

mstachniuk (approved):

PR #843, opened 2023-07-21
SONARIAC-1016 Parse error on object attribute when named "type"

mstachniuk:


its/ruling/src/test/resources/expected/azureresourcemanager/azureresourcemanager-S2260.json
rudy-regazzoni-sonarsource, line 1: It seems we get more parse error than before, do you know why? Should we create another ticket to fix this?

rudy-regazzoni-sonarsource (approved):
LGTM! I'm a bit worried about what seems to be new parsing issues, please see my comments.

its/ruling/src/test/resources/expected/azureresourcemanager/azureresourcemanager-S2260.json
mstachniuk: I plan to handle them as separate ticket (baby steps): https://sonarsource.atlassian.net/browse/SONARIAC-1025

PR #845, opened 2023-07-21
SONARIAC-1027 ARM Bicep ResourceDeclaration properties should return …

mstachniuk:
…list of both ternary expression


rudy-regazzoni-sonarsource (changes requested):
This sound too risky for me, because we are going to merge two resource declarations into a single mixed resource, we could raise FP on those mixed properties, which would be really strange to the user. There is too much uncertainty to perform analyses on.

An alternative would be instead to execute the checks on the two different `ObjectProperty`. Not sure o the complexity to implement this though.

nils-werner-sonarsource (commented):
I agree with Rudy's concerns. We merge two different states. Use case: Raise an issue if properly 1 and property 2 are set to true. How do you ensure, that that both properties are set in the same state? IMO we should not process `TernaryExpression` for now within checks. The problem is part of path-sensitive control flow analysis and should be handled with care. Let's discuss a better solution at a given time. Right now, it does not seem like a blocker that creates FPs.

mstachniuk:
Fixed in https://github.com/SonarSource/sonar-iac/pull/857

PR #849, opened 2023-07-25
SONARIAC-1028 ARM Bicep: exception since ResourceDeclaration.version() can return null

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualResource.java
```diff
@@ -33,7 +33,7 @@ public final class ContextualResource extends ContextualMap<ContextualResource,
   private ContextualResource(CheckContext ctx, ResourceDeclaration tree, String type) {
     super(ctx, tree, tree.name().value(), null);
     this.type = type;
-    this.version = tree.version().value();
+    this.version = tree.version() != null ? tree.version().value() : "";
```
petertrr, line 36: Is it worth storing `null` here instead of an empty string? I can see that we only actually access this version in one check, and it still works with an empty string, but I wonder if Azure has default versions for resources that may lead us to false negatives if we simply ignore the missing version.

petertrr (approved):
LGTM, but please see my consideration

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualResource.java
rudy-regazzoni-sonarsource: That's a good point.
I made this method return null in case there is no version because it does not return a simple `String` but a `TextTree` object, which also expects a `textRange`, which we cannot really provide for an empty value.

The default version for azure is a valid concern that we may need to handle. We probably will not be able to determine it during the analysis as it can evolve over time, so we would need to choose to not raise (conservative approach) or to raise anyway (possible FP, in case of hotspot could be viable).

PR #852, opened 2023-07-25
SONARIAC-889 Disabling Managed Identities for Azure resources is secu…

mstachniuk:
…rity-sensitive

Implementing the part for JSON
Adding "resource properties" to ResourceDeclaration Small improvements in tests


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ResourceDeclarationConverter.java
```diff
@@ -65,28 +67,47 @@ public ResourceDeclaration convertToResourceDeclaration(MappingTree tree) {
...
-    return new ResourceDeclarationImpl(name, version, type, otherProperties, Collections.emptyList());
+  private List<Property> toResourceProperties(MappingTree tree) {
+    return tree.elements().stream()
+      .map(tupleTree -> {
+        Identifier key = toIdentifier(tupleTree.key());
+        Expression value = toExpression(tupleTree.value());
+        return new PropertyImpl(key, value);
+      })
+      .collect(Collectors.toList());
+  }
```
rudy-regazzoni-sonarsource, line 89: Will this `resourceProperties` contain also `properties`?
If so, should we think about deprecating our old `properties`, and in the near future replace it with `resourceProperties`?

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/api/ExpressionAssert.java
rudy-regazzoni-sonarsource, line 1: Nice improvement 👍

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/ManagedIdentityCheckTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+  @MethodSource(value = "managedIdentityTypes")
+  @ParameterizedTest(name = "[{index}] Bicep should check managed identity check for type {0}")
+  void shouldCheckManagedIdentityCheckBicep(String type) {
+    String content = ArmTestUtils.readTemplateAndReplace("ManagedIdentityCheck/managedIdentityCheck_template.bicep", type);
+
+    BicepVerifier.verifyContent(content, check,
+      issue(1, 9, 1, 39, "Omitting the \"identity\" block disables Azure Managed Identities. Make sure it is safe here."),
+      issue(18, 10, 18, 16, "Make sure that disabling Azure Managed Identities is safe here."),
+      issue(33, 2, 35, 3, "Omitting the \"type\" in \"identity\" block disables Azure Managed Identities. Make sure it is safe here."));
```
rudy-regazzoni-sonarsource, line 232: Why not define the expected issues directly in the file as comment?
Easier to define and to read afterwards.

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my few suggestions.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/ManagedIdentityCheckTest.java
mstachniuk: Because it's for different types, I have to first read the content, replace `${type}` and then check issues. Other methods expect a filename as argument.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ResourceDeclarationConverter.java
rudy-regazzoni-sonarsource: Yes, that's why I wanted to propose to keep your code as it is and just add a `@Deprecated` tag on the method properties(), as now with `resourceProperties()` we have two ways to access the same data, and the one you are proposing is more complete.
I don't want to change everything now, just potentially encourage writers of next rules to use this `resourceProperties()` instead of `properties()`.

PR #855, opened 2023-07-25
BUILD-3415 tasks must depend on build

julien-carsique-sonarsource:
Any task calling `source cirrus-env <TASK_TYPE>` with `TASK_TYPE!=BUILD*` must depend on the task calling `source cirrus-env BUILD`.

If you want to parallelize the build and the scan tasks, then you have to introduce a common root task to call `source cirrus-env BUILD` first, and then change the `<TASK_TYPE>` of the build task.
See https://github.com/SonarSource/re-ci-images/blob/master/docker/bin/cirrus-env#L24

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -114,8 +114,8 @@ build_task:
   cleanup_before_cache_script: cleanup_maven_repository
 
 qa_os_win_task:
+  <<: *QA_TASK_FILTER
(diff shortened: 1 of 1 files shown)


nils-werner-sonarsource (approved):
LGTM! Thanks for the explanation. It was me removing the `depends_on` for the sca task. I was wrong in assuming that the `mvn` command in the task is enough.

PR #856, opened 2023-07-26
Fix master red: fix bug and few code smell

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualResource.java
```diff
@@ -33,7 +35,7 @@ public final class ContextualResource extends ContextualMap<ContextualResource,
   private ContextualResource(CheckContext ctx, ResourceDeclaration tree, String type) {
     super(ctx, tree, tree.name().value(), null);
     this.type = type;
-    this.version = tree.version() != null ? tree.version().value() : "";
+    this.version = Optional.ofNullable(tree.version()).map(TextTree::value).orElse("");
```
petertrr, line 38: That's really strange that this line wasn't highlighted as a bug in the PR, but was highlighted as soon as it was merged. Do we have different quality profiles for branches?

petertrr (approved):
LGTM!

PR #857, opened 2023-07-26
SONARIAC-1033 ARM Bicep: fix parse error on non-ObjectExpression for resource properties

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImpl.java
```diff
@@ -197,6 +197,7 @@ public List<Decorator> decorators() {
   private static List<Property> propertiesOrEmpty(List<PropertyTree> properties) {
     return properties.stream()
       .filter(propertyTree -> "properties".equals(((TextTree) propertyTree.key()).value()))
+      .filter(propertyTree -> ((Property) propertyTree).value().is(Kind.OBJECT_EXPRESSION))
```
petertrr, line 200: Do we need or already have a ticket to support other types of expressions? Should it be added here as a todo comment to explain this filtering?

petertrr (approved):

PR #858, opened 2023-07-27
SONARIAC-1031 ARM Bicep: set parents after building the AST

rudy-regazzoni-sonarsource:


its/ruling/src/test/resources/expected/azureresourcemanager/azureresourcemanager-S6387.json
```diff
@@ -0,0 +1,5 @@
+{
+"azureresourcemanager-project:sources/azureresourcemanager/azure-docs-bicep-samples/samples/scenarios-rbac/scope-default.bicep": [
+6
```
petertrr, line 3: Cool, that this change made an existing rule work for Bicep!

petertrr (approved):
LGTM!

its/ruling/src/test/resources/expected/azureresourcemanager/azureresourcemanager-S6387.json
rudy-regazzoni-sonarsource: Yes! I was a bit surprised at first to see our QA ruling fail, then I saw that this rule was indeed using `parent()` method :smile:

PR #859, opened 2023-07-27
SONARIAC-896 Rule S6648 for ARM: Secure strings and objects should not have default values - basic implementation

rudy-regazzoni-sonarsource:
Refer to [SONARIAC-1041](https://sonarsource.atlassian.net/browse/SONARIAC-1041).
Does not implement check regarding parameters as mentioned in the ticket.

[SONARIAC-1041]: https://sonarsource.atlassian.net/browse/SONARIAC-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualParameter.java
```diff
@@ -0,0 +1,64 @@
...
+    if (defaultValue != null && predicate.test(defaultValue)) {
+      return (S) report(message, List.of(secondaries));
+    }
+    return (S) this;
+  }
+
+  @Override
+  @CheckForNull
+  protected HasTextRange toHighlight() {
+    return tree.defaultValue();
```
petertrr, line 62: A question regarding this method in general. I saw that we are only calling it inside `report` methods and raise an issue if `highlight != null`. It seems that implementations like this imply that we will only ever raise issues for parameters if they have default values, otherwise this will return `null` and `report` will ignore this location. Do I understand this correctly? Do we need to add fallback values in `toHighlight`?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
```diff
@@ -50,4 +57,29 @@ public static Predicate<Expression> isTrue() {
...
+  public static Predicate<Expression> isEmptyArray() {
+    return expr -> expr.is(ArmTree.Kind.ARRAY_EXPRESSION) && ((ArrayExpression) expr).elements().isEmpty();
+  }
+
+  public static Predicate<Expression> isEmptyObject() {
+    return expr -> expr.is(ArmTree.Kind.OBJECT_EXPRESSION) && ((ObjectExpression) expr).properties().isEmpty();
+  }
+
+  public static Predicate<Expression> isFunctionCall(String functionName) {
+    // TODO SONARIAC-10038 ARM Json: parse expression in string and build the AST to be same as Bicep equivalent
```
petertrr, line 78:
```suggestion
    // TODO SONARIAC-1038 ARM Json: parse expression in string and build the AST to be same as Bicep equivalent
```

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
```diff
@@ -50,4 +57,29 @@ public static Predicate<Expression> isTrue() {
...
+    // Here we detect functionCall in two ways:
+    // - in Bicep we expect a FunctionCall object
+    // - in Json we expect a StringLiteral with this format: "[functionName(...)]"
+    return expr -> (expr.is(ArmTree.Kind.STRING_LITERAL) && isRegexMatch("^\\[" + functionName + "\\(.*\\)\\]$").test(expr)) // ARM Json
```
petertrr, line 82: Does it allow spacing after `[` or before `]`?

petertrr (approved):
LGTM!

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualParameter.java
rudy-regazzoni-sonarsource: That's a fair point! I will add a check and return the `tree` itself if there is no `defaultValue` + add test for this.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
rudy-regazzoni-sonarsource: That's a good question.
I tried to decompile some json file into bicep, the result is that your string must expressly start with `[` and end with `]`.
Otherwise, the expression in the brackets can contain spaces anywhere.
The examples I tried:
```
"[uniqueString('secret')]"    -> uniqueString('secret')
"   [uniqueString('secret')]" -> '   [uniqueString(\'secret\')]'
"[uniqueString('secret')]   " -> '[uniqueString(\'secret\')]   '
"[   uniqueString('secret')]" -> uniqueString('secret')
"[uniqueString('secret')   ]" -> uniqueString('secret')
"[uniqueString(   'secret')]" -> uniqueString('secret')
"[uniqueString('secret'   )]" -> uniqueString('secret')
```
I will adapt the regex to support this 👍

PR #860, opened 2023-07-27
SONARIAC-1043 Align `ResourceDeclaration#name()` between JSON and Bicep

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ManagedIdentityCheck.java
```diff
@@ -224,7 +224,7 @@ private static BiConsumer<CheckContext, ResourceDeclaration> checkManagedIdentit
     return (ctx, resource) -> {
       Optional<Property> identityProperty = resource.resourceProperties().stream().filter(p -> "identity".equals(p.key().value())).findFirst();
       if (identityProperty.isEmpty()) {
-        ctx.reportIssue(resource.name(), OMITTING_IDENTITY_MESSAGE);
+        ctx.reportIssue(resource.symbolicName() != null ? resource.symbolicName() : resource.name(), OMITTING_IDENTITY_MESSAGE);
```
petertrr, line 227: I changed it like this to keep the previous behavior. I'm not sure, why here it was done with `name`, as we usually simply highlight the entire resource

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/ResourceDeclarationImpl.java
```diff
@@ -66,10 +67,15 @@ public List<Tree> children() {
...
   @Override
-  public Identifier name() {
+  public StringLiteral name() {
     return name;
   }
 
+  @Override
+  public @CheckForNull Identifier symbolicName() {
+    return null;
+  }
```
rudy-regazzoni-sonarsource, line 77:
```suggestion
  @CheckForNull 
  @Override
  public Identifier symbolicName() {
    return null;
  }
```
Nitpick

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImpl.java
```diff
@@ -95,7 +97,18 @@ public Kind getKind() {
...
   @Override
-  public Identifier name() {
+  public StringLiteral name() {
+    return resourceProperties().stream()
+      .filter(prop -> TextUtils.isValue(prop.key(), "name").isTrue())
+      .filter(prop -> prop.value().is(Kind.STRING_COMPLETE))
+      .findFirst()
+      .map(prop -> ((StringComplete) prop.value()).content())
+      .orElse(null);
+  }
```
rudy-regazzoni-sonarsource, line 107: We should add a `@CheckForNull` on this method, as now we can expect it to sometimes return a null value.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ResourceDeclarationImplTest.java
rudy-regazzoni-sonarsource, line 1: It is missing a new test where `.name()` would return `null` for a resource, either because it is not specified, or because it is an expression we currently don't handle (`TernaryExpression` for example).

rudy-regazzoni-sonarsource (approved):

PR #861, opened 2023-07-28
SONARIAC-1042 Rule S6648 for ARM: Secure strings and objects should not have default values - remove secure parameter reference case

rudy-regazzoni-sonarsource:


petertrr (approved):
LGTM

PR #862, opened 2023-07-28
SONARIAC-901 Rule S6380: Authorizing anonymous access to Azure resources is security-sensitive - `ApiManagement/service`

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
```diff
@@ -50,12 +56,34 @@ private static void checkWebSites(ContextualResource resource) {
...
+    if (!isEqual(APIMGMT_PORTALSETTINGS_SIGNIN_RESOURCE_NAME).test(resource.tree.name())) {
+      return;
+    }
+
+    resource.property("enabled").reportIf(isFalse(), APIMGMT_PORTAL_SETTINGS_DISABLED_MESSAGE);
+
+    resource.childResourceBy("apis", r -> true)
+      .property("authenticationSettings")
+      .reportIfAbsent(APIMGMT_AUTHENTICATION_SETTINGS_NOT_SET_MESSAGE);
+  }
```
mstachniuk, line 88: Consider move this logic to separate method. IMO it can produce FP for `Microsoft.ApiManagement/service/portalsettings/signIn/apis` (I'm not sure if such resource is valid).
Also the logic in code should be similar as described in ticket. Now `checkApiManagementPortalSettings` is called for 2 different types, where the ticket descriptions mention `authenticationSettings` only once.

mstachniuk (approved):

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
```diff
@@ -29,18 +29,24 @@
 
 @Rule(key = "S6380")
 public class AnonymousAccessToResourceCheck extends AbstractArmResourceCheck {
-  private static final String AUTH_SETTINGS_V2_RESOURCE_NAME = "authsettingsV2";
+  private static final String WEBSITES_CONFIG_AUTH_SETTINGS_V2_RESOURCE_NAME = "authsettingsV2";
   private static final String WEBSITES_MISSING_AUTH_SETTINGS_MESSAGE = "Omitting authsettingsV2 disables authentication. Make sure it is safe here.";
   private static final String WEBSITES_DISABLED_AUTH_MESSAGE = "Make sure that disabling authentication is safe here.";
+  private static final String APIMGMT_PORTALSETTINGS_SIGNIN_RESOURCE_NAME = "signin";
```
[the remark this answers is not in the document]
petertrr: There are a lot of constants in this class (especially after all of the 5 tickets are added), so I decided to make the names as descriptive, as possible

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
petertrr: Good catch, I placed this check in an incorrect place. Moved and added more test cases.

> Now `checkApiManagementPortalSettings` is called for 2 different types

It's actually called for the same type, but the type can be either standalone or nested. I believe our `AbstractArmResourceCheck#register` doesn't handle nested types

PR #866, opened 2023-07-31
SONARIAC-917 Whitespace line is counted as code when prefixed with code

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/yaml/visitors/YamlMetricsVisitor.java
```diff
@@ -35,26 +36,33 @@ public YamlMetricsVisitor(FileLinesContextFactory fileLinesContextFactory, NoSon
...
-    linesOfCode().add(tree.textRange().start().line());
-    linesOfCode().add(endLine(tree));
+  /**
+   * Analyze line by line and do not use nodes from YAML parser since their text range contains trailing lines with whitespaces which should
+   * not be counted as LOC
+   */
+  private void analyzeFileContentForLoc(InputFile inputFile) throws IOException {
+    int lineNumber = 1;
+    String[] lines = inputFile.contents().split("(\\r\\n|\\r|\\n)");
+    for (String line : lines) {
```
petertrr, line 55: Would iterating over lines without reading the entire file into memory make sense here? Or will the difference be negligible, because we read it later anyway for parsing?

petertrr (approved):
LGTM!

iac-common/src/main/java/org/sonar/iac/common/yaml/visitors/YamlMetricsVisitor.java
nils-werner-sonarsource: Thank you for the suggestion. I was wondering if `InputFile` provides a proper API for this, but I was overlooking the `InputFile::inputStream` method.

PR #867, opened 2023-08-01
SONARIAC-930 Rule S6656: ARM Template evaluation should not expose secure values

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SecureValuesExposureCheck.java
```diff
@@ -0,0 +1,85 @@
...
+    if (sensitiveParameters.isEmpty()) {
+      return;
+    }
+
+    ContextualObject expressionEvaluationOptions = resource.object("expressionEvaluationOptions");
+    ContextualProperty scope = expressionEvaluationOptions.property("scope");
+    if (scope.isAbsent() || TextUtils.isValue(scope.valueOrNull(), "Inner").isFalse()) {
+      Set<String> sensitiveParameterNames = sensitiveParameters.keySet();
+      List<SecondaryLocation> sensitiveParameterUsages = resource.object("template").list("resources").objects()
+        .filter(ContextualTree::isPresent)
```
rudy-regazzoni-sonarsource, line 69:
```suggestion
```
Technically useless, `objects()` method return only `ContextualTree` elements that are not absent, or empty if there is none.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SecureValuesExposureCheck.java
```diff
@@ -0,0 +1,85 @@
...
+      .filter(it -> it.getValue().type() == ParameterType.SECURE_OBJECT || it.getValue().type() == ParameterType.SECURE_STRING)
+      .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+    if (sensitiveParameters.isEmpty()) {
+      return;
+    }
+
+    ContextualObject expressionEvaluationOptions = resource.object("expressionEvaluationOptions");
+    ContextualProperty scope = expressionEvaluationOptions.property("scope");
+    if (scope.isAbsent() || TextUtils.isValue(scope.valueOrNull(), "Inner").isFalse()) {
```
rudy-regazzoni-sonarsource, line 66:
```suggestion
    if (!TextUtils.isValue(scope.valueOrNull(), "Inner").isTrue()) {
```
Suggestion, we want to raise an issue only if the scope is not set to `Inner`, can be written this way.
Keep the one you prefer.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/SecureValuesExposureCheckTest.java
rudy-regazzoni-sonarsource, line 1: I would split the tests methods by files, as we usually like to split out test with clearly defined goal.
A compliant file test should be in it own method `testJsonCompliantSecureScope` for example.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/SecureValuesExposureCheckTest.java
```diff
@@ -0,0 +1,41 @@
...
+    ArmVerifier.verify("SecureValuesExposureCheck/Microsoft.Resources_deployments_noncompliant.json", new SecureValuesExposureCheck(),
+      Verifier.issue(12, 14, 12, 47, "Change this code to not use an outer expression evaluation scope in nested templates."));
+  }
+
+  @Test
+  void testBicep() {
+    BicepVerifier.verifyNoIssue("SecureValuesExposureCheck/Microsoft.Resources_deployments_compliant.bicep", new SecureValuesExposureCheck());
+    BicepVerifier.verifyNoIssue("SecureValuesExposureCheck/Microsoft.Resources_deployments_compliant_no_usages.bicep", new SecureValuesExposureCheck());
+    BicepVerifier.verify("SecureValuesExposureCheck/Microsoft.Resources_deployments_noncompliant.bicep", new SecureValuesExposureCheck());
+    // TODO: files `main.bicep` and `vm.bicep` are a preferred fix for the issue, but we don't support multiple files analysis yet
```
rudy-regazzoni-sonarsource, line 39: You could write a different test method for main.bicep and vm.bicep, and disable this test with @disabled annotation with text parameters your todo.

iac-extensions/arm/src/test/resources/checks/SecureValuesExposureCheck/Microsoft.Resources_deployments_noncompliant.bicep
rudy-regazzoni-sonarsource, line 1: It is missing a test where the parameters would not be `@secure()`.
Same with equivalent for json.
I would also add a test where you refer a parameter that is not defined, to see/document how our rule reacts to it (in both json and bicep).

rudy-regazzoni-sonarsource (approved):
LGTM! Nice work! Please see my few comments. Depending on how you plan to tackle the nested template part, you may want to create a ticket about it and refer it in your todo.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SecureValuesExposureCheck.java
petertrr: On the next line I'm accessing `ContextualObject#tree`, so I added this check just to make sure

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/SecureValuesExposureCheck.java
```diff
@@ -0,0 +1,85 @@
...
+      Set<String> sensitiveParameterNames = sensitiveParameters.keySet();
+      List<SecondaryLocation> sensitiveParameterUsages = resource.object("template").list("resources").objects()
+        .filter(ContextualTree::isPresent)
+        .flatMap(o -> o.tree.allPropertiesFlattened())
+        .map(Property::value)
+        .filter(containsParameterReference(sensitiveParameterNames))
+        .map(value -> new SecondaryLocation(value.textRange(), SECONDARY_MESSAGE))
+        .collect(Collectors.toList());
+
+      // TODO: also check nested templates
```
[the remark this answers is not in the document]
petertrr: Yes. I've clarified this with AppSec; when deployment is performed, values are evaluated based on the parameters and `scope` from the root template, but can be substituted in deeply nested `deployments` resources. So I'll need to recursively collect all usages in nested resources here.

PR #869, opened 2023-08-02
SONARIAC-1056 Improve IaC assertions for issues

mstachniuk:


iac-common/src/test/java/org/sonar/iac/common/testing/Verifier.java
```diff
@@ -294,12 +294,12 @@ private static void compare(List<Issue> actualIssues, List<Issue> expectedIssues
   }
 
   private static class Tuple {
-    private static final String NO_ISSUE = "* [NO_ISSUE] Expected but no issue on range %s.\n\n";
-    private static final String WRONG_MESSAGE = "* [WRONG_MESSAGE] Issue at %s : \nexpected: %s\nbut was: %s";
-    private static final String UNEXPECTED_ISSUE = "* [UNEXPECTED_ISSUE] at %s with a message: \"%s\"\n\n";
+    private static final String NO_ISSUE = "* [NO_ISSUE] %s\n\n";
+    private static final String WRONG_MESSAGE = "* [WRONG_MESSAGE]\nexpected: %s\nbut was:  %s";
```
petertrr, line 298:
```suggestion
    private static final String WRONG_MESSAGE = "* [WRONG_MESSAGE]\nexpected: %s\nbut was: %s";
```
Is this extra space intentional?

petertrr (approved):
LGTM!

iac-common/src/test/java/org/sonar/iac/common/testing/Verifier.java
mstachniuk: Yes, because then both values are vertically aligned so it's easier to find the difference.

PR #870, opened 2023-08-02
SONARIAC-1055 Add metrics for Bicep files

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmParser.java
```diff
@@ -32,9 +32,13 @@ public class ArmParser implements TreeParser<Tree> {
...
   public ArmTree parse(String source, @Nullable InputFileContext inputFileContext) {
-    if (inputFileContext != null && inputFileContext.inputFile.filename().endsWith(".bicep")) {
+    if (inputFileContext != null && isBicepFile(inputFileContext)) {
       return bicepParser.parse(source, inputFileContext);
     }
     return jsonParser.parse(source, inputFileContext);
   }
+
+  public static boolean isBicepFile(InputFileContext inputFileContext) {
+    return inputFileContext.inputFile.filename().endsWith(".bicep");
```
nils-werner-sonarsource, line 42:
```suggestion
    return ArmLanguage.KEY.equals(inputFileContext.inputFile.language());
```
This is not related to your changes, but we should test the language of a file and not a suffix. We would need to update the suffix here, in case the suffix for the language can be customized.

nils-werner-sonarsource: And I would prefer to have this logic in the ArmSensor and not in the parser since the ArmMetricsVisitor should not have a direct dependency on the parser.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/plugin/ArmMetricsVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+  protected void languageSpecificMetrics() {
+    register(SyntaxToken.class, (ctx, token) -> {
+      if (!(token.value().trim().isEmpty())) {
+        TextRange range = token.textRange();
+        for (int i = range.start().line(); i <= range.end().line(); i++) {
+          linesOfCode().add(i);
+        }
+      }
+      addCommentLines(token.comments());
+    });
```
nils-werner-sonarsource, line 50: As this is a duplicate do you think we can move the BiConsumer to `MetricsVisitor`?

iac-extensions/arm/src/test/java/org/sonar/iac/arm/plugin/JsonMetricsVisitorTest.java
```diff
@@ -0,0 +1,68 @@
...
+      "  {",
+      "    \"type\": \"Microsoft.ContainerService/managedClusters\",",
+      "    \"apiVersion\": \"2023-03-01\",",
+      "    \"name\": \"Compliant\",",
+      "    \"properties\": {",
+      "    \"enableRBAC\": true",
+      "    }",
+      "  }",
+      "]",
+      "}"));
```
nils-werner-sonarsource, line 65: I'm looking forward to use Text Blocks :smile:

iac-common/src/test/java/org/sonar/iac/common/testing/AbstractMetricsTest.java
```diff
@@ -71,12 +71,16 @@ protected MetricsVisitor scan(String... codeLines) {
...
       .setContents(code)
       .build();
     InputFileContext ctx = new InputFileContext(sensorContext, inputFile);
     visitor.scan(ctx, parser.parse(code, null));
     return visitor;
   }
+
+  protected String testFileName() {
+    return "file";
+  }
```
nils-werner-sonarsource, line 85: You wouldn't need this if you decide on the language (see next comment). However, then you have to provide a language for the test file.

nils-werner-sonarsource (approved):
LGTM! Please see my comments.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmParser.java
petertrr: Thanks for the suggestion, I wasn't aware that this information is available in file context. However, it feels somewhat counter-intuitive: it's `ArmLanguage` and not `BicepLanguage`, as if we are implying that there could be more formats countable as ARM sources.

PR #874, opened 2023-08-03
SONARIAC-1050 ARM Bicep S6329 Allowing public network access to cloud resources is security-sensitive

rudy-regazzoni-sonarsource:


petertrr (approved):
LGTM, just one question

PR #881, opened 2023-08-04
SONARIAC-1052 ARM Bicep S6388 Using unencrypted cloud storages is security-sensitive

petertrr:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/UnencryptedCloudServicesCheckTest.java
```diff
@@ -159,13 +214,35 @@ void testResourcesWithInfrastructureEncryption(String resourceType) {
...
+  void testResourcesWithInfrastructureEncryptionBicep(String resourceType) {
+    String content = ArmTestUtils.readTemplateAndReplace("UnencryptedCloudServicesCheck/MultiUnencryptedInfrastructureEncrypted_template.bicep", resourceType);
+
+    int resourceTypeLength = resourceType.length();
+    BicepVerifier.verifyContent(content, check,
+      issue(18, 4, 18, 31, "Make sure that using unencrypted cloud storage is safe here."),
+      issue(25, 4, 25, 25, "Make sure that using unencrypted cloud storage is safe here."),
+      issue(29, 48, 29, 80, "Omitting \"encryptionState\" enables clear-text storage. Make sure it is safe here."),
+      issue(36, 4, 36, 40, "Make sure that using unencrypted cloud storage is safe here."),
+      issue(40, 52, 40, 52 + resourceTypeLength, "Omitting \"infrastructureEncryption\" enables clear-text storage. Make sure it is safe here."));
```
rudy-regazzoni-sonarsource, line 231: I believe you can still specify the issue directly in the bicep file and not provide the list of issues here, would be easier to read afterward (and also probably to setup).

iac-extensions/arm/src/test/resources/checks/UnencryptedCloudServicesCheck/Compute_disk_and_snapshots.bicep
```diff
@@ -0,0 +1,146 @@
+// Noncompliant@+1 {{Omitting "encryption.diskEncryptionSetId", "encryptionSettingsCollection" or "securityProfile.secureVMDiskEncryptionSetId" enables clear-text storage. Make sure it is safe here.}}
+resource sensitiveDiskEncryptionSetIdEncryptionSettingsCollectionAndSecureVMDiskEncryptionSetIdAreNotSet 'Microsoft.Compute/disks@2022-07-02' = {
```
rudy-regazzoni-sonarsource, line 2:
```suggestion
resource noncompliant1 'Microsoft.Compute/disks@2022-07-02' = {
```
Suggestion, no need to keep the default decompiled symbolic name, it is not really readable and we already have the full information in the name of the resource.
I usually put `noncompliantX` / `compliantX` instead, so that we quickly identify in which case we are and then can read it if we want to know more.

rudy-regazzoni-sonarsource (approved):
LGTM! A big rule with lot of tests! Please see my few comments.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/UnencryptedCloudServicesCheckTest.java
petertrr: Actually, the way we are using `SingleFileVerifier` in `verifyConent` now doesn't seem to allow this: we never call `SingleFileVerifier#assertOneOrMoreIssues` from inside `verifyContent`

PR #882, opened 2023-08-07
SONARIAC-926 ARM Bicep S6385 Azure custom roles should not grant subscription Owner capabilities

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
```diff
@@ -93,6 +94,28 @@ public static Predicate<Expression> isFunctionCall(String functionName) {
       || (expr.is(ArmTree.Kind.FUNCTION_CALL) && ((FunctionCall) expr).name().value().equals(functionName));// ARM Bicep
   }
 
+  // Detect if the provided expression is a call to a specific function combined with access to a specific property. Example: myFunc().myProp
```
petertrr, line 97:
```suggestion
  /**
   * Detect if the provided expression is a call to a specific function combined with access to a specific property. 
   Example: myFunc().myProp
   */
```

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
```diff
@@ -93,6 +94,28 @@ public static Predicate<Expression> isFunctionCall(String functionName) {
...
+  public static Predicate<Expression> isFunctionCallWithPropertyAccess(String functionName, String propertyName) {
+    // TODO SONARIAC-1038 ARM Json: parse expression in string and build the AST to be same as Bicep equivalent
+    // Here we detect functionCall in two ways:
+    // - in Bicep we expect a MemberExpression with following attributes (separatingToken=".", memberAccess=FunctionCall,
+    // expression={"propertyName" expression})
+    // - in Json we expect a StringLiteral with this format: "[functionName(...).propertyName]"
+    return expr -> {
+      // ARM Json
+      if (expr.is(ArmTree.Kind.STRING_LITERAL) && isRegexMatch("^\\[\\s*+" + functionName + "\\(.*\\)\\s*+\\." + propertyName + "\\s*+\\]$").test(expr)) {
```
petertrr, line 106: Maybe we could introduce some kind of function to generate ARM JSON expression patterns, in order to at least not duplicate `[...]` part? Like, `armExpressionPattern(functionName)` would return `"^\[\s*" + functionName + "\s*\]$`, and it could be reused already here and in `isFunctionCall`.

petertrr (approved):
LGTM!

PR #885, opened 2023-08-07
SONARIAC-918 ARM Bicep S6387 Azure role assignments that grant access to all resources of a subscription are security-sensitive

petertrr:


rudy-regazzoni-sonarsource (approved):
LGTM! Very nice and clean PR!

PR #886, opened 2023-08-07
SONARIAC-907 ARM Bicep S6364 ARM Defining a short backup retention duration is security-sensitive

rudy-regazzoni-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ShortBackupRetentionCheck.java
```diff
@@ -138,7 +139,18 @@ private static Double computeRetentionInDays(ContextualObject retentionPolicyObj
...
   private static Predicate<Expression> isNumericValue(DoublePredicate predicate) {
-    return expr -> expr.is(ArmTree.Kind.NUMERIC_LITERAL) && predicate.test(((NumericLiteral) expr).asDouble());
+    return expr -> {
+      if (expr.is(ArmTree.Kind.NUMERIC_LITERAL)) {
+        return predicate.test(((NumericLiteral) expr).asDouble());
+      } else if (expr.is(ArmTree.Kind.UNARY_EXPRESSION)) {
+        UnaryExpression unaryExpression = (UnaryExpression) expr;
+        if (unaryExpression.expression().is(ArmTree.Kind.NUMERIC_LITERAL)) {
+          double factor = unaryExpression.operator().value().equals("-") ? -1 : 1;
+          return predicate.test(((NumericLiteral) unaryExpression.expression()).asDouble() * factor);
```
petertrr, line 149: Maybe it could have been a method somewhere in tree utils like `asNumericValueOrNull(UnaryExpression expression)` that would perform kind check and extract the factor?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ShortBackupRetentionCheck.java
```diff
@@ -156,8 +168,8 @@ private static String toString(@Nullable Tree tree) {
     return Optional.ofNullable(tree)
```
petertrr, line 168: It's not actually part of this PR, but for me, a more clear name for this method would be something like `propertyValueAsString` or `getValueAsString`. `toString` feels a bit too generic.
Also, there is some overlapping with methods from `PropertyUtils`, but I couldn't think of a way to use them here, since we don't have a common supertype for `StringLiteral` and `StringComplete`

iac-extensions/arm/src/test/resources/checks/ShortBackupRetentionCheck/Microsoft.Web_sites.bicep
```diff
@@ -0,0 +1,129 @@
+resource noncompliant1 'Microsoft.Web/sites/config@2022-03-01' = {
+  name: 'backup'
+  properties: {
+    __sonar__: 'Sensitive: retention period is less than 30'
```
petertrr, line 4: I have never seen this field before, but apparently, we have it in a couple of tests. What's the difference compared to putting description into `name`?

petertrr (approved):
LGTM, see a couple of my suggestions

iac-extensions/arm/src/test/resources/checks/ShortBackupRetentionCheck/Microsoft.Web_sites.bicep
rudy-regazzoni-sonarsource: In bicep, a resource type can have restrictions on which names are allowed, and depending on the name used, It can define how the resource object behaves.

This is the case with `Microsoft.Web/sites/config`, where one of the names expected is `backup`.
In the original version of how this check S6364 was written, we ensured that the name of the resource was backup, that's also why we have a few tests about it.
This check was later dropped: the other check we performs in the rule are sufficient, there was no upside to ensuring the name was `backup`.

What you see here is mostly an artifact, although real bicep code is indeed expected this resource to have the name `backup`.

PR #890, opened 2023-08-08
Add ITs for rules S6364 and S6385

rudy-regazzoni-sonarsource:


its/ruling/src/test/resources/sources/azureresourcemanager/rules/S6385.bicep
```diff
@@ -0,0 +1,33 @@
...
+      }
+    ]
+    assignableScopes: [
+      subscription().id
+    ]
+  }
+}
+
+resource compliant 'Microsoft.Authorization/roleDefinitions@2022-04-01' = {
+  name: 'Compliant - specific actions are specified'
```
petertrr, line 19: I was thinking - maybe for ITs we should place compliant example before non-compliant? Since we are not checking issues in detail and only the first line, it would make sense to verify this way that we are not triggering on compliant examples.

petertrr (approved):
LGTM, see my comment

PR #891, opened 2023-08-08
SONARIAC-1060 Fix FP on S6648 : do not raise on null value or empty string

rudy-regazzoni-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM, please see my comment before merging

PR #895, opened 2023-08-09
Sonariac 933 ARM Resource names are case-insensitive

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/ResourceDeclarationImpl.java
```diff
@@ -67,6 +68,11 @@ public List<Tree> children() {
     return children;
   }
 
+  /**
+   * The name of a resource is case-insensitive. Comparisons to this field should always respect this property.
+   * An easy way to do this is via {@link TextUtils#isValue(Tree, String)}.
+   * @see <a href=”https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules”>Microsoft - Naming rules and restrictions for Azure resources</a>
+   */
```
petertrr, line 75: Should this description be put in the interface?

petertrr (approved):

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/ResourceDeclarationImpl.java
jonas-wielage-sonarsource: You're right. I originally thought this would only apply for json, but it applies for bicep aswell. So the description should move aswell.

PR #896, opened 2023-08-10
SONARIAC-1070 ARM JSON Support reading required fields by case-insensitive names

petertrr:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/OutputDeclarationImplTest.java
```diff
@@ -208,10 +208,10 @@ void shouldFailOnMissingMandatoryAttribute() {
       "  }",
       "}");
     ParseException parseException = catchThrowableOfType(() -> parser.parse(code, null), ParseException.class);
-    assertThat(parseException).hasMessage("Missing mandatory attribute 'type' at null:3:4");
+      assertThat(parseException).hasMessage("Missing mandatory attribute 'type' at null:3:21");
```
petertrr, line 211: I tried to merge two overloads of `Arm(Json)BaseConverter#toStringLiteralOrException` into one; not sure if it had a lot of meaning to highlight the entire `TupleTree` in case of an exception instead of `TupleTree#value`

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmJsonBaseConverter.java
```diff
@@ -57,34 +57,28 @@
...
-    return PropertyUtils.get(tree, key)
+    return PropertyUtils.get(tree, key::equalsIgnoreCase)
       .map(this::toStringLiteral)
       .orElse(null);
   }
 
-  public StringLiteral toStringLiteralOrException(TupleTree tree, String key) {
-    return PropertyUtils.get(tree.value(), key)
+  public StringLiteral toStringLiteralOrException(YamlTree tree, String key) {
+    return PropertyUtils.get(tree, key::equalsIgnoreCase)
```
jonas-wielage-sonarsource, line 76: I like

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/OutputDeclarationImplTest.java
jonas-wielage-sonarsource, line 1: Could maybe also benefit form case insensitive test.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/OutputDeclarationImplTest.java
jonas-wielage-sonarsource: Fine for me

jonas-wielage-sonarsource (approved):
LGTM, please see my comments.
Also shouldn't there also be a test for case-insensitive VariableDeclarationImpl?

PR #901, opened 2023-08-17
SONARIAC-1072 Update ARM documentation to include bicep files

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/plugin/ArmSettings.java
```diff
@@ -55,10 +58,21 @@ public static List<PropertyDefinition> getGeneralProperties() {
...
-        .name("File Identifier")
-        .description("Files without the identifier are excluded from the analysis. The identifier can be anywhere in the file.")
+        .name("JSON Template File Identifier")
+        .description("ARM JSON templates without the identifier are excluded from the analysis. The identifier can be anywhere in the file.")
+        .onQualifiers(Qualifiers.PROJECT)
+        .category(ARM_CATEGORY)
+        .subCategory(GENERAL_SUBCATEGORY)
+        .build(),
+
+      PropertyDefinition.builder(FILE_SUFFIXES_KEY)
```
nils-werner-sonarsource, line 68: Why did you add this property? It is not taken into account in `ArmLanguage`. It's not described in the ticket.

nils-werner-sonarsource (changes requested):
Please not add a property which has no effect on the analysis

PR #909, opened 2023-09-11
SONARIAC-744 S6596: Specific version tag for image should be used

jonas-wielage-sonarsource:


iac-extensions/docker/src/test/resources/checks/SpecificVersionTagCheck/specificVersionTagCheck.dockerfile
```diff
@@ -0,0 +1,79 @@
...
+
+# Noncompliant@+2
+FROM \
+foobar:latest
+
+# Compliant
+FROM my-image:1.2.3 as baseWithTag
+FROM my-image:1.2.3-alpine
+FROM ubuntu:18.04 as local-toolchain-ubuntu18.04-manylinux2010
+FROM foobar@12313423 # Compliant usage of digest
```
rudy-regazzoni-sonarsource, line 61: [Dockerfile comments cannot be put after an instruction](https://docs.docker.com/engine/reference/builder/#format), they must be alone on their line.

> Docker treats lines that begin with # as a comment, unless the line is a valid [parser directive](https://docs.docker.com/engine/reference/builder/#parser-directives). A # marker anywhere else in a line is treated as an argument.

rudy-regazzoni-sonarsource, line 61: I didn't saw in the ticket nor in the rspec PR that using a digest is considered as compliant, to me we should still raise an issue in that case.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpecificVersionTagCheck.java
```diff
@@ -0,0 +1,89 @@
...
+      // raise an issue if the version tag is "latest"
+      String[] splitImageName = fullImageName.split(":");
+      return splitImageName.length >= 2 && splitImageName[1].equals("latest");
+    } else {
+      // no version tag specified, docker assumes "latest"
+      return true;
+    }
+  }
+
+  private static boolean imageNameIsUsedAsAliasInPreviousImages(FromInstruction currentFromInstruction, String currentImageName) {
```
rudy-regazzoni-sonarsource, line 68: Not necessarily a big issue here because I don't expect dockerfile to contain a ton of docker image definitions, but this approach has an n^2 complexity, as every time you add a docker image, this will perform an additional iteration looping itself on all previously defined image.
A simpler way would be to have a local cache in this class that retains visited docker images (or even just their alias), but we need to carefully clean this cache or make it available per dockerfile.
I honestly don't have an exact idea about how to implement this or if a similar mechanism already exist.

rudy-regazzoni-sonarsource (changes requested):
Pretty good, just requesting a few adaptations, but otherwise everything seems good to me!
Feel free to reach out if you want to discuss some of my comments.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpecificVersionTagCheck.java
jonas-wielage-sonarsource: Yeah, I'm not sure what would be the best way either.
Normally we don't want to save the state in between checks, so I figured because we don't expect a lot of FROM instructions in a docker file, that this is save.

iac-extensions/docker/src/test/resources/checks/SpecificVersionTagCheck/specificVersionTagCheck.dockerfile
jonas-wielage-sonarsource: Well a digest is not a version tag, so we shouldn't raise here.
The ticket only describes to raise an issue on a missing version tag.
And it wouldn't make sense to raise an issue here on every digest, because we shouldn't discourage from using it.

jonas-wielage-sonarsource: Fair point, the comments should not be on the same level here

TyIsI:
Hey! Sorry to bump this old PR but Jira is not being insightful.

Can you advise me if SONARIAC-1323 deals with the `scratch` image matching? (`FROM scratch` matches.)
Or else where I should report this issue?

Thanks!

jonas-wielage-sonarsource:
Hello @TyIsI:
Currently, we don't have any logic in place to handle `scratch`.
But there is already a ticket for Docker that aims to deal with `scratch` images: [SONARIAC-1260](https://sonarsource.atlassian.net/browse/SONARIAC-1260). I bumped the priority on the ticket, so it should be handled faster.

The one you mentioned ([SONARIAC-1323](https://sonarsource.atlassian.net/browse/SONARIAC-1323)) refers to Kubernetes and Helm analysis.

Hope that clarifies your question.

Also you can report False positives in our community, which you find here: [SonarSource Community - How to report a false positive / false negative](https://community.sonarsource.com/t/how-to-report-a-false-positive-false-negative/37022)

Best,
Jonas

TyIsI:
Copy. Thank you!

PR #910, opened 2023-09-11
SONARIAC-728 S6579: Access variable which is not available in the current scope

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ArgDefinedOutsideOfScopeCheck.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  private static void checkArgUsagesIn(DockerImage image, List<String> globalArgNames, CheckContext ctx) {
+    List<String> argNamesInStage = collectArgNames(image.instructions().stream().filter(ArgInstruction.class::isInstance).map(ArgInstruction.class::cast));
+    List<String> notRedeclaredArgsNames = new ArrayList<>(globalArgNames);
+    notRedeclaredArgsNames.removeAll(argNamesInStage);
+
+    List<Variable> usedVariables = image.instructions().stream()
+      .filter(RunInstruction.class::isInstance)
+      .flatMap(run -> ((RunInstruction) run).arguments().stream())
```
rudy-regazzoni-sonarsource, line 69: Should we only check `RunInstruction`? What about references in `CmdInstruction` or somewhere else?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ArgDefinedOutsideOfScopeCheck.java
```diff
@@ -0,0 +1,81 @@
...
+public class ArgDefinedOutsideOfScopeCheck implements IacCheck {
+  private static final String MESSAGE = "Include the ARG instruction in the build stage where it is used.";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(Body.class, ArgDefinedOutsideOfScopeCheck::checkReferencedVariablesAreInScope);
+  }
+
+  private static void checkReferencedVariablesAreInScope(CheckContext ctx, Body body) {
+    List<String> globalArgsNames = collectArgNames(body.globalArgs().stream());
```
rudy-regazzoni-sonarsource, line 48: I see what you are doing here.

It may not be necessary to do this, because in our docker analyzer, we have a Symbol table for variables.
Currently, thanks to the `DockerSymbolVisitor`, each variable has a Symbol associated, and this Symbol give you direct access to all reference of the variable.

Typically, the first reference is the "declaration" of the variable, and even better, it has a scope associated, which is either `Global` (`ARG` instruction at the beginning of the file, before any `FROM`) or `Image` (`ARG` instruction in a docker image, so after a `FROM` instruction). So exactly what you are looking for in your rule :smile:

rudy-regazzoni-sonarsource (changes requested):
Suggesting different approaches + challenging the scope of the rule, please see my comments.
In case we want to extend this rule to instruction other than RUN, a solution could to attach a visitor to `Variable` directly.
Feel free to reach me out if you want to discuss it.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ArgDefinedOutsideOfScopeCheck.java
petertrr: I wasn't able to find a comprehensive list of where ARGs can be used. The [docs](https://docs.docker.com/engine/reference/builder/#arg) say

> You can use an ARG or an ENV instruction to specify variables that are available to the RUN instruction.

However, they show example also with `USER`. I think at least it should be covered for all kinds of `CommandInstruction`

petertrr: But I guess the logic for everything other than `FROM` is the same so this can be indeed generalized

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ArgDefinedOutsideOfScopeCheck.java
```diff
@@ -0,0 +1,61 @@
...
+    boolean isDefinedInImageScope = false;
+    boolean isDefinedInGlobalScope = false;
+    for (Usage usage : variable.symbol().usages()) {
+      if (usage.kind().equals(Usage.Kind.ASSIGNMENT)) {
+        isDefinedInImageScope = usage.scope().kind().equals(Scope.Kind.IMAGE);
+        isDefinedInGlobalScope = usage.scope().kind().equals(Scope.Kind.GLOBAL);
+      }
+    }
+    boolean isDefinedOnlyInGlobalScope = isDefinedInGlobalScope && !isDefinedInImageScope;
+    if (isDefinedOnlyInGlobalScope) {
```
rudy-regazzoni-sonarsource, line 57: This code is a bit strange.

You are overriding `isDefinedInImageScope` and `isDefinedInGlobalScope` at every loop iteration, which also means they can never have the same value, so the boolean `isDefinedOnlyInGlobalScope` is useless.

From my understanding, you are trying to check if a variable is only defined with a `Global` scope.
You don't really need to do this, but only to check if the last assignment is `Global`, as they are kept in the line order.
```suggestion
    
    List<Usage> usages = variable.symbol().usages();
    boolean isGlobalScope = false;
    for (int i = usages.size() - 1; i >= 0; i--) {
      Usage usage = usages.get(i);
      if (usage.kind().equals(Usage.Kind.ASSIGNMENT)) {
        isGlobalScope = usage.scope().kind().equals(Scope.Kind.GLOBAL);
        break; // we stop at the first assignment we find
      }
    }

    if (isGlobalScope) {
```
(Feel free to improve the suggested code, I couldn't find a better way to write it).

iac-extensions/docker/src/test/resources/checks/ArgDefinedOutsideOfScopeCheck/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+FROM busybox
+# Noncompliant@+1 {{Include the ARG instruction in the build stage where it is used.}}
+RUN ./run/setup $SETTINGS
+#               ^^^^^^^^^
+
+FROM busybox
+ARG SETTINGS
+RUN ./run/setup $SETTINGS
+
```
rudy-regazzoni-sonarsource, line 17: You could add here another occurence of this:
```
FROM busybox
# Noncompliant@+1
RUN ./run/setup $SETTINGS
```
It would show that the algorithm is working for this rule, as the `Symbol` table is not mixing this variable symbol in `RUN` with the one redeclared at line 15 in `ARG`.
I would add a comment about this also.

rudy-regazzoni-sonarsource (changes requested):
Sorry to push back again, for me there is an issue with the algorithm: while it is working, it seems to show a different intention from what it's really doing. Please see my comment, feel free to come back to me if you want to discuss it!

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ArgDefinedOutsideOfScopeCheck.java
petertrr: Thanks for catching this, my code is definitely incorrect here :sweat_smile: 
However, I think that relying on the order in which usages are present couples the check logic too tightly with symbol table implementation. I would rather fix my code or add an accessor `Symbol#getDeclarationScope`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/symbols/Symbol.java
```diff
@@ -56,4 +56,16 @@ public List<Usage> usages() {
...
+    for (int i = usages.size() - 1; i >= 0; i--) {
+      Usage usage = usages.get(i);
+      if (usage.kind().equals(Usage.Kind.ASSIGNMENT)) {
+        declarationScope = usage.scope().kind();
+        // we stop at the first assignment we find, as we store usages in the same order they appear in Dockerfile
+        break;
+      }
+    }
+    return declarationScope;
+  }
```
rudy-regazzoni-sonarsource, line 70: I'm wondering what result do you get for a variable that is used but never declared? Do we have a symbol for it in that case? Do this methis return an `IMAGE` scope in that case? Shouldn't we get null instead? (+add @CheckForNull)

rudy-regazzoni-sonarsource (approved):
LGTM! I like this rule implementation, it's clear and simple! I added one last comment.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/symbols/Symbol.java
petertrr: We don't create symbols for such variables, AFAIU. In `DockerSymbolVisitor` we only add symbols when visiting assignment nodes, that's why we can assume that if there is no symbol for a variable, then it's not a Dockerfile-level variable, but probably a shell env.

PR #911, opened 2023-09-11
SONARIAC-1092 S6506 should raise an issue if sensitive https request is encapsulated in quotes

jonas-wielage-sonarsource:


iac-extensions/docker/src/test/resources/checks/ClearTextProtocolDowngradeCheck/wget.dockerfile
```diff
@@ -15,6 +17,20 @@ RUN wget https://might-redirect.example.com/install.sh --max-redirect=0 -q -O -
 
 RUN wget https://might-redirect.example.com/install.sh -q --max-redirect=0 -O - | sh
 
+# Noncompliant@+1
+RUN wget "https://cmake.org/files/v3.7/cmake-3.7.2-SHA-256.txt.asc"
```
mstachniuk, line 21: I'm a little bit confused 😕 . There is a change in logic for `curl`, but tests are changed for `wget`. Is it expected?

iac-extensions/docker/src/test/java/org/sonar/iac/docker/checks/utils/StringPredicateTest.java
```diff
@@ -22,36 +22,45 @@
...
 class StringPredicateTest {
 
-  @Test
-  void testSingleValuePositive() {
+  @ParameterizedTest
+  @ValueSource(strings = {
+    "value",
+    "\"value\"",
+    "'value'"
+  })
```
mstachniuk, line 40: ❤️

mstachniuk (approved):

iac-extensions/docker/src/test/resources/checks/ClearTextProtocolDowngradeCheck/wget.dockerfile
jonas-wielage-sonarsource: Ah, the changed files don't paint the best picture.
If you look at  `SENSITIVE_HTTPS_URL_BEGINNING`, it's changed to include the `ignoreQuotes` predicate. This is used further down in the `wget` commands, which remain unchanged because the name stayed the same.
The `curl` change didn't change the logic but combined the Predicate creation of `EQUALS_PROTO_FLAG_OPTION = equalsIgnoreQuotes("=https")` as it was used two times.

PR #912, opened 2023-09-12
SONARIAC-720 Rule S6570: Detect missing double quote to prevent globbing and word splitting

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/test/resources/checks/VariableReferenceOutsideOfQuotesCheck/VariableReferenceOutsideOfQuotesCheck.dockerfile
jonas-wielage-sonarsource, line 1: Looks good. Can we add test cases to add a variable as a flag?
`RUN cd $test` for example

jonas-wielage-sonarsource (approved):

PR #913, opened 2023-09-12
SONARIAC-721 Rule S6574: A space before the equal sign in key-value pair may lead to unintended behavior

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/test/resources/checks/SpaceBeforeEqualInKeyValuePairCheck/SpaceBeforeEqualInKeyValuePairCheck.dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+
+# Non compliant use cases
+# Noncompliant@+1 {{Remove space before the equal sign in the key-value pair, as it can lead to unexpected behavior.}}
+LABEL VERSION =42
+#     ^^^^^^^^^^^
+# Noncompliant@+1
+LABEL VERSION = 42
+# Noncompliant@+1
+LABEL VERSION =42 NAME=BOB
```
petertrr, line 15: Will it work the other way around, `LABEL VERSION=42 NAME =Bob`? Or would that be invalid Dockerfile?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpaceBeforeEqualInKeyValuePairCheck.java
```diff
@@ -0,0 +1,64 @@
...
+
+@Rule(key = "S6574")
+public class SpaceBeforeEqualInKeyValuePairCheck implements IacCheck {
+
+  private static final String MESSAGE = "Remove space before the equal sign in the key-value pair, as it can lead to unexpected behavior.";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(LabelInstruction.class, (ctx, label) -> checkKeyValurPairs(ctx, label.labels()));
+    init.register(EnvInstruction.class, (ctx, env) -> checkKeyValurPairs(ctx, env.environmentVariables()));
```
petertrr, line 44: Should there also be subscriber for `ARG` instructions?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpaceBeforeEqualInKeyValuePairCheck.java
```diff
@@ -0,0 +1,64 @@
...
+      ctx.reportIssue(keyValuePairs.get(0), MESSAGE);
+    }
+  }
+
+  private static boolean isSensitiveKeyValuePair(List<KeyValuePair> keyValuePairs) {
+    return keyValuePairs.size() == 1 && keyValuePairs.get(0).equalSign() == null && isStartingWithLiteralStringWithEqual(keyValuePairs.get(0).value());
+  }
+
+  private static boolean isStartingWithLiteralStringWithEqual(@Nullable Argument argument) {
+    return Optional.ofNullable(argument).stream()
```
petertrr, line 58: Is `stream()` needed here or can result be achieved with `map`s and `filter`s on Optional?

petertrr (approved):
LGTM, see a couple of suggestions

iac-extensions/docker/src/test/resources/checks/SpaceBeforeEqualInKeyValuePairCheck/SpaceBeforeEqualInKeyValuePairCheck.dockerfile
rudy-regazzoni-sonarsource: This is indeed an invalid docker syntax and also crash our parser, so that's why I didn't add it :grin:

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpaceBeforeEqualInKeyValuePairCheck.java
rudy-regazzoni-sonarsource: `ARG` instruction is a bit different, we also use `KeyValuePair` for it but it only allows the form with the equals.
Meaning this `ARG key =value` is an invalid syntax.
So it make no sense for this rule to check `ARG` instructions.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpaceBeforeEqualInKeyValuePairCheck.java
rudy-regazzoni-sonarsource: Good catch! No need to stream it, I will just replace the `anyMatch()` with `filter()`+`isPresent()`.

PR #915, opened 2023-09-12
SONARIAC-1089 S6437 Hardcoded secrets in wget, curl, sshpass and htpasswd

mstachniuk:


iac-extensions/docker/src/test/java/org/sonar/iac/docker/checks/utils/command/ArgumentResolutionIncludeUnresolvedPredicateTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.checks.utils.command;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ArgumentResolutionIncludeUnresolvedPredicateTest {
```
jonas-wielage-sonarsource, line 26: I would like some actual testing on the matching, if it's possible. Especially seeing different types in action.

jonas-wielage-sonarsource (changes requested):
While I see that the logic is good and working, I believe we could still improve here. Please have a look at my comments.

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/wget_template.dockerfile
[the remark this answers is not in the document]
mstachniuk: Yes, for all 4 commands long and short forms are checked, where applicable.

iac-extensions/docker/src/test/resources/checks/SecretsGenerationCheck/wget_template.dockerfile
```diff
@@ -0,0 +1,95 @@
+FROM ubuntu:22.04
+
+ARG PASSWORD
+
+# Noncompliant@+1
+RUN wget --user=guest --flag=MySuperPassword https://example.com
```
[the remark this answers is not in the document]
mstachniuk: Yes, or template or copy-paste 🤷‍♂️

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -62,10 +63,68 @@ public class SecretsGenerationCheck implements IacCheck {
     .withAnyOptionExcluding(Collections.emptyList())
     .build();
 
+  private static final CommandDetector WGET_PASSWORD_EQUALS = wgetFlagEquals("--password");
```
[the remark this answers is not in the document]
mstachniuk: Yes, it's hard to choose the best name. 
Let's add 2 small comments on 2 first constant and following rename:
`WGET_PASSWORD_EQUALS` -> `WGET_PASSWORD_FLAG_EQUALS_PWD`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -77,5 +136,67 @@ private static void checkRunInstruction(CheckContext ctx, RunInstruction runInst
...
+      ctx.reportIssue(new CommandDetector.Command(resolvedArgument), MESSAGE);
+    }
+  }
+
+  private static ArgumentResolution getLastArgument(CommandDetector.Command command) {
+    List<ArgumentResolution> arguments = command.getResolvedArguments();
+    return arguments.get(arguments.size() - 1);
+  }
+
+  private static class HtpasswdDetector {
```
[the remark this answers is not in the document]
mstachniuk: This was almost my first implementation, but then I had a problem with cyclomatic complexity.
Other commands (like `git` `ls`) also support the pattern that `-b -n` can be replaced by `-bn`.
So next time it will be easier to extract similar logic.

Another reason for a separate static class is SRP/granularity. `SecretsGenerationCheck` executes all checks using "high-level" `CommandDetector` DLS. The logic in `HtpasswdDetector` is "low-level". It deserves for at least separate static class.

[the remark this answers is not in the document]
jonas-wielage-sonarsource: Fair point. The thing that I don't like about this approach is the creation of a new class object for every command encountered. What about still encapsulating this in a private class, which only has a static boolean method instead of a "public" constructor?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/IncludeUnresolvedPredicate.java
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.docker.checks.utils.command;
+
+import java.util.function.Predicate;
+import org.sonar.iac.docker.symbols.ArgumentResolution;
+
+public class IncludeUnresolvedPredicate extends SingularPredicate {
```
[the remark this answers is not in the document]
mstachniuk: Actually `IncludingUnresolvedArgumentsPredicate` because `SingularPredicate` ignores matching when the Argument is unresolved.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/ArgumentResolutionIncludeUnresolvedPredicate.java
```diff
@@ -0,0 +1,66 @@
...
+    ArgumentResolution resolution = context.getNextArgumentToHandleAndRemoveFromList();
+
+    if (predicate.test(resolution)) {
+      // Add matched argument
+      context.addAsArgumentToReport(resolution);
+    } else {
+      context.setStatus(FOUND_NO_PREDICATE_MATCH);
+      return;
+    }
+    context.setStatus(CONTINUE);
```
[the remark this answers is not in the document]
mstachniuk: I have an idea of how it can be changed, but let's discuss it offline (and for sure not in the scope of this PR).

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -62,20 +66,164 @@ public class SecretsGenerationCheck implements IacCheck {
...
+    if (!dockerImage.isLastDockerImageInFile()) {
+      return;
+    }
+
+    List<RunInstruction> runInstructions = dockerImage.instructions().stream()
+      .filter(RunInstruction.class::isInstance)
+      .map(RunInstruction.class::cast)
+      .collect(Collectors.toList());
+
+    runInstructions.forEach(runInstruction -> checkRunInstruction(ctx, runInstruction));
```
jonas-wielage-sonarsource, line 151: You can use `forEach` directly in line 149 to the extra collection

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/ArgumentResolutionIncludeUnresolvedPredicate.java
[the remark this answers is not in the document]
jonas-wielage-sonarsource: That's true. Let's create a ticket for it, and maybe simply `ABORT` the detection when the type is not implemented yet.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -253,6 +259,21 @@ public CommandDetector.Builder withPredicatesFrom(CommandDetector.Builder otherB
       return this;
     }
 
+    public CommandDetector.Builder withIncludeUnresolved(Predicate<String> predicate) {
+      addIncludeUnresolved(predicate, MATCH);
+      return this;
+    }
+
+    public CommandDetector.Builder withAnyExcludingIncludeUnresolved(Predicate<String> predicate) {
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: The problem in my opinion is the hard brake between `Excluding` and `Include`.
Previously we had the excluding at the end so that it's related to the parameter provided. My suggestion would be:
`withAnyIncludingUnresolvedExcluding` or to be more verbose:
`withAnyIncludingUnresolved(But|And)Excluding`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -62,20 +66,164 @@ public class SecretsGenerationCheck implements IacCheck {
...
     List<ArgumentResolution> resolvedArgument = CheckUtils.resolveInstructionArguments(runInstruction);
 
     DETECTORS.forEach(
       detector -> detector.search(resolvedArgument).forEach(command -> ctx.reportIssue(command, MESSAGE)));
+
+    if (isMountSecret(runInstruction)) {
+      return;
+    }
+
+    DETECTORS_IGNORE_MOUNT_SECRET.forEach(
```
jonas-wielage-sonarsource, line 164: I would not add the `IGNORE_MOUNT_SECRET` to this name, as the detectors don't actively ignore the mount secret. It's only ignored because we are after the if call.
I would add a comment above this line:
`The detectors aren't evaluating a possible mount secret flag, because it is handled in the if condition above`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/DockerImageImpl.java
```diff
@@ -59,6 +60,14 @@ public List<Instruction> instructions() {
...
   }
 
+  @Override
+  public boolean isLastDockerImageInFile() {
+    Body parent = (Body) parent();
+    List<DockerImage> dockerImages = parent.dockerImages();
+    DockerImage last = dockerImages.get(dockerImages.size() - 1);
+    return last == this;
+  }
+
```
jonas-wielage-sonarsource, line 70: Not quite sure if this logic is in the right place here. I think I would have put it in `TreeUtils`. Especially with the call to `parent()`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -253,6 +259,21 @@ public CommandDetector.Builder withPredicatesFrom(CommandDetector.Builder otherB
...
+      addCommandPredicate(new IncludingUnresolvedArgumentsPredicate(predicate, ZERO_OR_MORE));
+      return this;
+    }
+
+    public CommandDetector.Builder withArgumentResolutionIncludeUnresolved(Predicate<ArgumentResolution> predicate) {
```
jonas-wielage-sonarsource, line 272: `withIncludingUnresolved` should be enough, as the parameter already describes the required generic

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -253,6 +259,21 @@ public CommandDetector.Builder withPredicatesFrom(CommandDetector.Builder otherB
       return this;
     }
 
+    public CommandDetector.Builder withIncludeUnresolved(Predicate<String> predicate) {
```
jonas-wielage-sonarsource, line 262: `withIncludingUnresolved`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/IncludingUnresolvedArgumentsArgumentResolutionPredicate.java
```diff
@@ -0,0 +1,64 @@
...
+      if (Type.MATCH.equals(t)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public void match(PredicateContext context) {
+    ArgumentResolution resolution = context.getNextArgumentToHandleAndRemoveFromList();
```
jonas-wielage-sonarsource, line 48: IMO should `ABORT` / fail somehow if the type is not MATCH, as we don't have the implementation right now. Maybe worth to create a refactoring ticket

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -62,20 +66,164 @@ public class SecretsGenerationCheck implements IacCheck {
...
   @Override
   public void initialize(InitContext init) {
-    init.register(RunInstruction.class, SecretsGenerationCheck::checkRunInstruction);
+    init.register(DockerImage.class, SecretsGenerationCheck::checkDockerImage);
+  }
+
+  private static void checkDockerImage(CheckContext ctx, DockerImage dockerImage) {
+    if (!dockerImage.isLastDockerImageInFile()) {
+      return;
+    }
```
jonas-wielage-sonarsource, line 144: I like this solution!

jonas-wielage-sonarsource (approved):
LGTM! Please see my suggestions

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
mstachniuk: > I would not add the IGNORE_MOUNT_SECRET
But then I will have `DETECTORS` list 😢 
Let's rename in that way (to describe what those detectors do):
`DETECTORS` -> `DETECTORS_THAT_STORE_SECRETS`
`DETECTORS_IGNORE_MOUNT_SECRET ` -> `DETECTORS_THAT_HAVE_SECRETS_IN_CMD`

mstachniuk: Also added a different comment with an explanation.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/DockerImageImpl.java
mstachniuk: I disagree. The `*Utils` classes are a sign of a bad design, not Object Oriented Programming.
Our `DockerTree` implementations are in most cases data structures, not rich objects. Our `DockerTree`s look like anemic entities in a typical project with Hibarnate. 

Here is a simple logic, only a real-life instance of `DockerImageImpl` is needed.
It's only valid for `DockerImageImpl`. The `TreeUtils` contains valid functions for all kinds of `Tree` implementations. (BTW IMO they should be moved to the `Tree` interface).

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
mstachniuk: `withArgumentResolutionIncludeUnresolved(Predicate<ArgumentResolution>)` -> `withIncludingUnresolved` and
`withIncludeUnresolved(Predicate<String>)` -> `withIncludingUnresolved`
is not possible because of weak Java typing :-|

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/command/IncludingUnresolvedArgumentsArgumentResolutionPredicate.java
mstachniuk: It's a baby step, let's discuss it offline how to improve `CommandDetector` design

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
jonas-wielage-sonarsource: To bad :/

PR #918, opened 2023-09-13
SONARIAC-729 S6573 shouldn't raise on patterns in Bash case statement

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ShellExpansionsInCommandCheck.java
```diff
@@ -76,10 +79,17 @@ private static boolean contains(List<ArgumentResolution> argumentResolutions, St
     return false;
   }
 
-  private static boolean isCompliantExceptionCommand(List<ArgumentResolution> argumentResolutions) {
-    if (argumentResolutions.size() < 2) {
+  private static boolean isCompliantExceptionCommand(List<ArgumentResolution> argumentResolutionsBeforeWildcard) {
+    // Even exception commands are not allowed to have wildcard as a first argument
+    // So we need to check if one of the compliant commands is somewhere before wildcard but not immediately before
```
mstachniuk, line 84: Maybe it will be good to add some example here?

mstachniuk (approved):

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ShellExpansionsInCommandCheck.java
petertrr: Added and also adjusted logic a bit, as I understood while writing an example that `echo -n *` should also be non-compliant

PR #920, opened 2023-09-13
SONARIAC-730 S6581: Environment variables should not be unset on a different layer than they were set

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/MisusedUnsetCheck.java
```diff
@@ -0,0 +1,66 @@
...
+@Rule(key = "S6581")
+public class MisusedUnsetCheck implements IacCheck {
+  private static final String MESSAGE = "Use the ARG instruction or set & unset environment variable in a single layer.";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(DockerImage.class, (ctx, image) -> image.instructions().stream()
+      .filter(RunInstruction.class::isInstance)
+      .map(RunInstruction.class::cast)
+      .forEach(runInstruction -> checkCommandsIn(runInstruction, ctx, name -> image.scope().getSymbol(name))));
```
jonas-wielage-sonarsource, line 45: I think it will be easier to understand if you pass the `image.scope()` and call the method below.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/MisusedUnsetCheck.java
```diff
@@ -0,0 +1,66 @@
...
+      .withAnyFlagExcept("-f")
+      .withOptionalRepeating(s -> true)
+      .build();
+    unsetDetector.search(argumentResolutions).forEach(c -> c.getResolvedArguments().stream()
+      // first element is `unset` command...
+      .skip(1)
+      // ... and then possibly some flags
+      .dropWhile(s -> s.value().startsWith("-"))
+      .filter(s -> s.argument().expressions().get(0) instanceof Literal)
+      .map(a -> (Literal) a.argument().expressions().get(0))
```
jonas-wielage-sonarsource, line 62: For me it's not clear why we use `get(0)`, maybe we could benefit from a comment.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/MisusedUnsetCheck.java
```diff
@@ -0,0 +1,66 @@
...
+      .filter(l -> lookupSymbol.apply(l.value()) != null)
```
jonas-wielage-sonarsource, line 63: If it's in the scope, does it mean it was set at a previous layer?

iac-extensions/docker/src/test/resources/checks/MisusedUnsetCheck.dockerfile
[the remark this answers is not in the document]
jonas-wielage-sonarsource: There should be compliant examples using `ARG` when it's mentioned in the rspec.
Also what about this example (written here as small as possible):
```
RUN export FOO=bar
RUN unset FOO
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/MisusedUnsetCheck.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static void checkCommandsIn(RunInstruction runInstruction, CheckContext ctx, Function<String, Symbol> lookupSymbol) {
+    List<ArgumentResolution> argumentResolutions = CheckUtils.resolveInstructionArguments(runInstruction);
+    CommandDetector unsetDetector = CommandDetector.builder()
+      .with("unset")
+      // -f unsets function names in Bash, while we only care about variables
+      .withAnyFlagExcept("-f")
+      .withOptionalRepeating(s -> true)
+      .build();
+    unsetDetector.search(argumentResolutions).forEach(c -> c.getResolvedArguments().stream()
```
jonas-wielage-sonarsource, line 56: For me it's a bit too much stream magic. I believe some of the filtering could be put together in an additional method and still be used to filter here.

jonas-wielage-sonarsource (changes requested):
Please look at my suggestions. I believe the code can be made more readable for others.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/MisusedUnsetCheck.java
```diff
@@ -0,0 +1,66 @@
...
+import org.sonar.iac.docker.checks.utils.CommandDetector;
+import org.sonar.iac.docker.symbols.ArgumentResolution;
+import org.sonar.iac.docker.symbols.Symbol;
+import org.sonar.iac.docker.tree.api.DockerImage;
+import org.sonar.iac.docker.tree.api.Literal;
+import org.sonar.iac.docker.tree.api.RunInstruction;
+
+@Rule(key = "S6581")
+public class MisusedUnsetCheck implements IacCheck {
+  private static final String MESSAGE = "Use the ARG instruction or set & unset environment variable in a single layer.";
```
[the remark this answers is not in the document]
petertrr: https://github.com/SonarSource/rspec/pull/3076

iac-extensions/docker/src/test/resources/checks/MisusedUnsetCheck.dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ENV SECRET_VALUE=password SECRET_USER=admin
+
+# Noncompliant@+1 {{Use the ARG instruction or set & unset environment variable in a single layer.}}
+RUN ./script -p $SECRET_VALUE && unset SECRET_VALUE
+#                                      ^^^^^^^^^^^^
+# Noncompliant@+1
+RUN ./script -p $SECRET_VALUE && unset -v SECRET_VALUE
+#                                         ^^^^^^^^^^^^
+# Noncompliant@+2
+# Noncompliant@+1
```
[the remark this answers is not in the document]
petertrr: Well, there are two variables being unset and an issue is raised for each in the current implementation. However, this can be marked as `# Noncompliant@+1 2`

iac-extensions/docker/src/test/resources/checks/MisusedUnsetCheck.dockerfile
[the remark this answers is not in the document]
petertrr: Regarding the second example: like this it won't work because `FOO` won't be persisted between shell invocations; it can be persistent somehow through a file, but not in a standard/straightforward way. So I would say this is compliant with respect to this rule, as bash-managed env variables, even persisted, are different from Docker-managed.

Good catch about `ARG`

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/MisusedUnsetCheck.java
petertrr: It is set either on global scope or in one of the previous instructions in this image scope. In both cases, `unset`-ting it won't work as expected

iac-extensions/docker/src/test/resources/checks/MisusedUnsetCheck.dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN ./script -p $SECRET_VALUE && unset SECRET_VALUE SECRET_USER
+# Noncompliant@+1
+RUN ./script -p $SECRET_VALUE && unset LD_LIBRARY_PATH SECRET_USER
+#                                                      ^^^^^^^^^^^
+
+RUN export PASSWORD=password && ./script -p $PASSWORD && unset PASSWORD
+RUN PASSWORD=password && ./script -p $PASSWORD && unset -v PASSWORD
+RUN unset -f foo
```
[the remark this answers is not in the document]
petertrr: WDYM?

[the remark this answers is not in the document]
jonas-wielage-sonarsource: `RUN someCommand && unset -f foo && someOtherCommand`

iac-extensions/docker/src/test/resources/checks/MisusedUnsetCheck/MisusedUnsetCheck.dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# Noncompliant@+1 {{Use the ARG instruction or set and unset the environment variable in a single layer.}}
+RUN ./script -p $SECRET_VALUE && unset SECRET_VALUE
+#                                      ^^^^^^^^^^^^
+# Noncompliant@+1
+RUN ./script -p $SECRET_VALUE && unset -v SECRET_VALUE
+#                                         ^^^^^^^^^^^^
+# Noncompliant@+1 2
+RUN ./script -p $SECRET_VALUE && unset SECRET_VALUE SECRET_USER
+# Noncompliant@+1
+RUN ./script -p $SECRET_VALUE && unset LD_LIBRARY_PATH SECRET_USER
```
jonas-wielage-sonarsource, line 15: What happens if we write sudo unset? Will this break the implementation with `.get(0)`

jonas-wielage-sonarsource (approved):
LGTM, please see my comment before merging

iac-extensions/docker/src/test/resources/checks/MisusedUnsetCheck/MisusedUnsetCheck.dockerfile
petertrr: No, the `get(0)` refers to `Argument#expressions()`, while `sudo` will be just another of arguments. It won't be captured by `CommandDetector`. But thanks for the remark, I'll add a test case.

PR #922, opened 2023-09-15
SONARIAC-1098 Improve line calculation when mixing HEREDOC and backslash

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
```diff
@@ -76,16 +77,45 @@ public Object createTerminal(Input input, int startIndex, int endIndex, List<Tri
     return new SyntaxTokenImpl(value, range, getCommentsForToken(range));
   }
 
+  /**
+   * Compute the {@link TextRange} from a value at the given {@code startIndex}, regarding the provided {@code input} (source code).
+   * In case it extend to multiple lines, we build a {@link CompoundTextRange} object with a reference to each line {@link TextRange}.
+   * This is required to allow to track back original source code before preprocessing, particularly needed when raising issue
+   * on the preprocessed source code.
+   */
   protected TextRange tokenRange(Input input, int startIndex, String value) {
```
rudy-regazzoni-sonarsource, line 86: This method algorithm should be improved in a second step: it would be better to be able to determine the loop increment value instead of just checking each character one by one until we detect that the lines has changed.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CompoundTextRange.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+import org.sonar.iac.common.api.tree.impl.TextRange;
+import org.sonar.iac.common.api.tree.impl.TextRanges;
+
+public class CompoundTextRange extends TextRange {
+  private final List<TextRange> textRanges;
+
+  public CompoundTextRange(List<TextRange> textRanges) {
+    super(textRanges.get(0).start(), textRanges.get(textRanges.size() - 1).end());
+    this.textRanges = textRanges;
```
mstachniuk, line 31: SC reports java:S2384 here (Store a copy of "textRanges".).
IMO It is "won't fix". This rule only applies to immutable objects, making a copy of the collection each time when passed to an object/method is in our case overkill.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CompoundTextRangeTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.sonar.iac.common.api.tree.impl.TextRange;
+import org.sonar.iac.common.api.tree.impl.TextRanges;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class CompoundTextRangeTest {
+
```
mstachniuk, line 30: Please add tests for `computeTextRangeAtIndex`. It's very complex method and not covered directly by unit test (only by other checks)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CompoundTextRange.java
```diff
@@ -0,0 +1,80 @@
...
+    // compute the final range
+    int startLine = startRange.start().line();
+    int startOffset = startRange.start().lineOffset() + startIndex;
+    int endLine = current.end().line();
+    int endOffset = (startRange.start().line() != current.end().line() ? 0 : startRange.start().lineOffset()) + end;
+    return TextRanges.range(startLine, startOffset, endLine, endOffset);
+  }
+
+  @Override
+  public boolean equals(Object other) {
```
mstachniuk, line 65: I wonder why my IDE generates another code for equals and hashCode (I checked a few templates) 🤔 
It's better to use built-in generators. In case if the class needs to be extended it will be easier to review those methods.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CompoundTextRange.java
```diff
@@ -0,0 +1,80 @@
...
+  public CompoundTextRange(List<TextRange> textRanges) {
+    super(textRanges.get(0).start(), textRanges.get(textRanges.size() - 1).end());
+    this.textRanges = textRanges;
+  }
+
+  /**
+   * Provide the correct {@link TextRange}, which can extend a multiple lines.
+   * It combines the {@code startIndex} and the {@code token} length to find which of the {@link #textRanges} are covered, to compute the correct single/multi line {@link TextRange}.
+   */
+  public TextRange computeTextRangeAtIndex(int startIndex, String token) {
```
mstachniuk, line 38: It's complex logic. After you add tests try to refactor it. Suggested algorithm:
TextRange startRange = findStartRange(...)
// calculate `i` based on `startRange`
TextRange endRange = findEndRange(...)
// calculate the rest

I know come calculation will be done twice, but it should be easier to understand it.

mstachniuk (changes requested):
Not approved because of lack of unit tests for complex code

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CompoundTextRange.java
```diff
@@ -0,0 +1,80 @@
...
+    while (end > current.end().lineOffset() - current.start().lineOffset()) {
+      end -= current.end().lineOffset() - current.start().lineOffset() + 1;
+      i++;
+      current = textRanges.get(i);
+    }
+    // compute the final range
+    int startLine = startRange.start().line();
+    int startOffset = startRange.start().lineOffset() + startIndex;
+    int endLine = current.end().line();
+    int endOffset = (startRange.start().line() != current.end().line() ? 0 : startRange.start().lineOffset()) + end;
```
mstachniuk, line 60: I agree with java:S1774 (Convert this usage of the ternary operator to an "if"/"else" structure.), it's hard to read.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CompoundTextRangeTest.java
```diff
@@ -0,0 +1,51 @@
...
+  @Test
+  void equals() {
```
mstachniuk, line 32: I agreee with java:S3578 (Rename this method name to match the regular expression: '^test[A-Z][a-zA-Z0-9]*$'), maybe the pattern should be `^should[A-Z][a-zA-Z0-9]*$` in our case

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CompoundTextRangeTest.java
```diff
@@ -0,0 +1,51 @@
...
+    TextRange sameRange = TextRanges.range(1, 2, 3, 4);
+    TextRange differentRange = TextRanges.range(5, 6, 7, 8);
+    CompoundTextRange compoundTextRange = new CompoundTextRange(List.of(range));
+    CompoundTextRange compoundTextRangeSame = new CompoundTextRange(List.of(sameRange));
+    CompoundTextRange compoundTextRangeDifferent = new CompoundTextRange(List.of(differentRange));
+    CompoundTextRange compoundTextRangeDifferent2 = new CompoundTextRange(List.of(sameRange, differentRange));
+    CompoundTextRange compoundTextRangeEmpty = new CompoundTextRange(List.of(differentRange));
+
+    assertThat(compoundTextRange)
+      .isEqualTo(compoundTextRange)
```
mstachniuk, line 43: I disagree with java:S2698 (Add a message to this assertion chain before the predicate method.) It makes sense always in `fail()` and sometimes in other cases, but for sure not in every assertion.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerHeredocParser.java
```diff
@@ -45,33 +44,22 @@ public static DockerHeredocParser create() {
     return new DockerHeredocParser(new DockerHeredocNodeBuilder(), DockerLexicalGrammar.HEREDOC_FORM_CONTENT);
   }
 
-  public DockerTree parse(String source, TextPointer offset) {
-    this.heredocNodeBuilder.setOffset(offset);
+  public DockerTree parse(String source, TextRange originalTextRange) {
```
mstachniuk, line 47: Actually, this method can take `SyntaxToken` as an argument

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/CompoundTextRange.java
rudy-regazzoni-sonarsource: I refactored as much as I could.
I get rid of the `i` variable by using an Iterator instead.

I identified the elements that I need to keep in the context: the iterator's current value and the index which changes all along.
For the Iterator's current value, I added a wrapper for it.
For the index value, we keep it through the new method `navigateToRangeAtIndex` return value.

I also added some minor improvements.

mstachniuk (changes requested):
It's better I like the direction of improvement. But still some parts are hard to understand

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
```diff
@@ -76,16 +77,45 @@ public Object createTerminal(Input input, int startIndex, int endIndex, List<Tri
...
+      int[] startLineAndColumn = sourceOffset.sourceLineAndColumnAt(i);
+      // detect line changes
+      if (currentLineAndColumn[0] != startLineAndColumn[0]) {
+        int startColumn = applyByteOrderMark(currentLineAndColumn[1], hasByteOrderMark);
+        int endColum = applyByteOrderMark(previousLineAndColumn[1], hasByteOrderMark);
+        ranges.add(TextRanges.range(currentLineAndColumn[0], startColumn, previousLineAndColumn[0], endColum));
+        currentLineAndColumn = startLineAndColumn;
+      }
+      previousLineAndColumn = startLineAndColumn;
+    }
```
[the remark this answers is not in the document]
rudy-regazzoni-sonarsource: Fair point for the `TextPointer`, but this class stores a `line` and an `offset`, while here it's about `line` and `column`.
Although they are related, they are not the same things (shifted by one), so using `TextPointer` here would bring confusion and be error-prone.
I will try to improve the code, as you suggested!

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
```diff
@@ -76,16 +78,57 @@ public Object createTerminal(Input input, int startIndex, int endIndex, List<Tri
     return new SyntaxTokenImpl(value, range, getCommentsForToken(range));
   }
 
+  /**
+   * Compute the {@link TextRange} from a value at the given {@code startIndex}, regarding the provided {@code input} (source code).
+   * In case it extend to multiple lines, we build a {@link CompoundTextRange} object with a reference to each line {@link TextRange}.
+   * It is used to track back tokens in HereDoc, as currently the full HereDoc content of an instruction is considered as a single {@link SyntaxToken},
+   * before it is being parsed by a specific HereDoc parser which then call this tokenRange method to split into multiple {@link SyntaxToken} with
+   * correct range.
+   * The {@code startIndex} is the position on the instruction line where the provided value begin.
```
mstachniuk, line 87:
```suggestion
   * <p/>   
   * The {@code startIndex} is the position on the instruction line where the provided value begin.
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
```diff
@@ -76,16 +78,57 @@ public Object createTerminal(Input input, int startIndex, int endIndex, List<Tri
     return new SyntaxTokenImpl(value, range, getCommentsForToken(range));
   }
 
+  /**
+   * Compute the {@link TextRange} from a value at the given {@code startIndex}, regarding the provided {@code input} (source code).
+   * In case it extend to multiple lines, we build a {@link CompoundTextRange} object with a reference to each line {@link TextRange}.
+   * It is used to track back tokens in HereDoc, as currently the full HereDoc content of an instruction is considered as a single {@link SyntaxToken},
+   * before it is being parsed by a specific HereDoc parser which then call this tokenRange method to split into multiple {@link SyntaxToken} with
```
mstachniuk, line 85: IMO we should avoid writing how it was done before. We should document how it is working now.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
```diff
@@ -76,16 +78,57 @@ public Object createTerminal(Input input, int startIndex, int endIndex, List<Tri
...
+    int[] currentLineAndColumn = sourceOffset.sourceLineAndColumnAt(startIndex);
+    int[] previousLineAndColumn = currentLineAndColumn;
     char[] fileChars = input.input();
     boolean hasByteOrderMark = fileChars.length > 0 && fileChars[0] == BYTE_ORDER_MARK;
 
-    int startColumn = applyByteOrderMark(startLineAndColumn[1], hasByteOrderMark);
+    int index;
+    for (index = startIndex + 1; index < startIndex + value.length(); index++) {
+      int[] startLineAndColumn = sourceOffset.sourceLineAndColumnAt(index);
+      // detect line changes
```
mstachniuk, line 101: this comment should be before line 107

iac-extensions/docker/src/test/java/org/sonar/iac/docker/tree/impl/CompoundTextRangeTest.java
```diff
@@ -0,0 +1,135 @@
...
+  void shouldReturnCorrectIndexForEmptyValue() {
+    var ranges = ranges(code("this test is good"));
+    CompoundTextRange compoundTextRange = new CompoundTextRange(ranges);
+
+    TextRange range = compoundTextRange.computeTextRangeAtIndex(5, "");
+    assertRange(range, 1, 5, 1, 5);
+  }
+
+  @Test
+  void shouldReturnCorrectIndexForTwoLine() {
```
mstachniuk, line 53: Now tests are much easier to understand ❤️

mstachniuk (approved):
LGTM, and please see my comments. Please also add test cases to the ruling test. Take the inspiration from files we ignored on peach.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
rudy-regazzoni-sonarsource: Did you mean to add a line break?
I'm not sure how valid `<p/>` is, in pure HTML I never saw this usage and from what I read it should be `<br />`.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/DockerNodeBuilder.java
rudy-regazzoni-sonarsource: The formulation is not clear, it will rework it a bit.

PR #925, opened 2023-09-15
SONARIAC-729 S6573 should not raise on patterns in Bash `for` loops

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ShellExpansionsInCommandCheck.java
```diff
@@ -53,9 +59,9 @@ private static void check(CheckContext ctx, CommandInstruction cmd) {
       .with(ShellExpansionsInCommandCheck::isShellExpansion)
       .build();
     for (List<ArgumentResolution> argumentResolutions : splitCommands.elements()) {
-      shellExpansionDetector.search(argumentResolutions).forEach(c -> {
+      shellExpansionDetector.search(argumentResolutions).forEach((CommandDetector.Command c) -> {
```
mstachniuk, line 62:
```suggestion
      shellExpansionDetector.search(argumentResolutions).forEach(c -> {
```

mstachniuk (approved):
LGTM

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/ShellExpansionsInCommandCheck.java
petertrr: This is a code smell detected by the new quality gate :shrug:

mstachniuk: Ahhh ok. It's https://sonarsource.github.io/rspec/#/rspec/S2211 
and the argumentation for me it's ok:

> When the lambda has one or two parameters and does not have a block this rule will not fire up an issue as things are considered more readable in those cases.

Here it is because of code block.

PR #926, opened 2023-09-19
SONARIAC-741 Rule S6589: Dockerfile should only have one ENTRYPOINT and CMD instruction

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/UniqueInstructionPresenceCheck.java
```diff
@@ -0,0 +1,70 @@
...
+      .collect(Collectors.toList());
+
+    for (Instruction instruction : allBodyInstructions) {
+      var kind = instruction.getKind();
+      if (uniqueInstructionKindsWithLabel.containsKey(kind)) {
+        foundInstructionsOfKind.computeIfAbsent(kind, key -> new ArrayList<>()).add(instruction);
+      }
+    }
+
+    for (Map.Entry<DockerTree.Kind, List<Instruction>> entry : foundInstructionsOfKind.entrySet()) {
```
petertrr, line 62: Speaking about `var`s, I think this would be a good example to discuss what is more readable: `Map.Entry<DockerTree.Kind, List<Instruction>> entry` or `var instructionKindToFoundInstructions`. I think the information that would be missing without explicit type can be expressed via more descriptive name.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/UniqueInstructionPresenceCheck.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private static void checkUniqueInstruction(CheckContext ctx, Body body) {
+    Map<DockerTree.Kind, List<Instruction>> foundInstructionsOfKind = new EnumMap<>(DockerTree.Kind.class);
+    List<Instruction> allBodyInstructions = body.dockerImages()
+      .stream()
+      .flatMap(image -> image.instructions().stream())
+      .collect(Collectors.toList());
+
+    for (Instruction instruction : allBodyInstructions) {
```
petertrr, line 55:
```suggestion
      .forEach(instruction -> {
```
As the list is not used anywhere other than in a loop, I think this can be done with streams completely

its/ruling/src/test/resources/expected/docker/docker-S6589.json
```diff
@@ -0,0 +1,70 @@
...
+"docker-project:sources/docker/awesome-compose/sparkjava/sparkjava/Dockerfile": [
+25
+],
+"docker-project:sources/docker/awesome-compose/traefik-golang/backend/Dockerfile": [
+23
+],
+"docker-project:sources/docker/awesome-compose/vuejs/vuejs/Dockerfile": [
+12
+],
+"docker-project:sources/docker/compose-dev-env/backend/Dockerfile": [
```
petertrr, line 67: Are these instructions actually ignored when only a specific build stage is being built (https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage)? For this particular Dockerfile it looks like a developer can `docker build --target development` and have some sort of development environment. The final stage depends only of `build` stage and not the `development` one.

petertrr (commented):
Please see my comments. I wonder if we should allow separate `CMD` and `ENTRYPOINT` instructions for build stages intended to be used separately during development.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/UniqueInstructionPresenceCheck.java
rudy-regazzoni-sonarsource: Ahh that's a good point! And a very good example that var can indeed have some benefits :joy:

its/ruling/src/test/resources/expected/docker/docker-S6589.json
rudy-regazzoni-sonarsource: As discussed, I changed the logic to be only per docker image.

petertrr (approved):
LGTM!

PR #927, opened 2023-09-19
SONARIAC-736 Rule S6587: Cache should be cleaned after package installation

jonas-wielage-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
```diff
@@ -0,0 +1,191 @@
...
+      analyzeCommands(commands, sensitiveAptInstallCommands, APT_INSTALL, REMOVE_APT_CACHE_DETECTOR, APT_CLEAN);
+    }
+
+    sensitiveApkInstallCommands.forEach(command -> ctx.reportIssue(command, MESSAGE));
+    sensitiveAptInstallCommands.forEach(command -> ctx.reportIssue(command, MESSAGE));
+  }
+
+  private static void analyzeCommands(
+    List<ArgumentResolution> commandsToSearchIn, List<CommandDetector.Command> sensitiveInstallCommands,
+    CommandDetector installDetector, CommandDetector removeCacheCommandDetector, CommandDetector cleanCacheCommandDetector) {
```
petertrr, line 121:
```suggestion
    List<ArgumentResolution> commandsToSearchIn,
    List<CommandDetector.Command> sensitiveInstallCommands,
    CommandDetector installDetector,
    CommandDetector removeCacheCommandDetector,
    CommandDetector cleanCacheCommandDetector) {
```

This might be my personal preference, but for me long parameter lists are more readable when every parameter is on its own line

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
```diff
@@ -0,0 +1,191 @@
...
+    init.register(RunInstruction.class, PackageInstallationCacheCheck::checkRunInstruction);
+  }
+
+  private static void checkRunInstruction(CheckContext ctx, RunInstruction runInstruction) {
+    List<ArgumentResolution> resolvedArgument = CheckUtils.resolveInstructionArguments(runInstruction);
+
+    SeparatedList<List<ArgumentResolution>, String> splitCommands = ArgumentResolutionSplitter.splitCommands(resolvedArgument);
+
+    List<CommandDetector.Command> sensitiveApkInstallCommands = new ArrayList<>();
+    List<CommandDetector.Command> sensitiveAptInstallCommands = new ArrayList<>();
```
petertrr, line 108: Are two different lists with two independent `forEach` required here or can all sensitive commands be collected into a single list? I don't think there is any difference in handling them after discovering.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
```diff
@@ -0,0 +1,191 @@
...
+  private static void removeCacheCleanedInstallCommands(List<CommandDetector.Command> installCommands, List<CommandDetector.Command> cacheCleaningCommands) {
+    if (cacheCleaningCommands.isEmpty()) {
+      return;
+    }
+    // only the last one is important
+    var lastCacheCleanCommand = cacheCleaningCommands.get(cacheCleaningCommands.size() - 1);
+    int latestIndexOfRemovableInstall = -1;
+    for (int j = installCommands.size() - 1; j >= 0 && latestIndexOfRemovableInstall == -1; j--) {
+      if (startsBefore(installCommands.get(j), lastCacheCleanCommand)) {
+        latestIndexOfRemovableInstall = j;
```
petertrr, line 154: I suggest `break` here instead of longer loop condition

petertrr (commented):
Nice implementation and a lot of test. Please see my questions.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
```diff
@@ -0,0 +1,191 @@
...
+
+@Rule(key = "S6587")
+public class PackageInstallationCacheCheck implements IacCheck {
+
+  private static final String MESSAGE = "Remove cache after installing packages.";
+
+  private static final Set<String> APK_CACHE_LOCATIONS = Set.of("/etc/apk/cache/*", "/var/cache/apk/*");
+  private static final Set<String> APT_COMMANDS = Set.of("apt", "apt-get", "aptitude");
+  private static final Set<String> APT_CACHE_LOCATION = Set.of("/var/lib/apt/lists/*");
+  private static final Predicate<String> containsROrF = s -> s.toLowerCase(Locale.ROOT).contains("r") || s.contains("f");
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: When using `RUN apk add nginx && rm /var/cache/apk`, it can result in not actually cleaning the cache, if there are directories inside the apk folder. I don't agree to not raise an issue here, as it's actually still sensitive. What would you suggest? Addming the `rm` as a secondary location?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
jonas-wielage-sonarsource: Fine for me either way.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
jonas-wielage-sonarsource: In `removeCacheCleanedInstallCommands` I remove all sensitive commands from these lists if there is a clean cache that is later. When we would collect all installs in a single list, we would lose the information if the install belongs to apk or apt. 
This would complicate some other parts of the code in such a way, that I don't feel it's worth it.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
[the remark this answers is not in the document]
petertrr: Ok, I assume that there may be cases when check for flags will be valuable. Let's leave the logic as is.

petertrr (approved):
LGTM!

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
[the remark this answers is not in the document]
jonas-wielage-sonarsource: TL;DR of the discussion: While it's more a misuse of the `rm` command and a failed attempt at cleaning the cache, we still feel it's valuable to have some the `Remove cache ...` issue displayed in the vicinity. 
As we don't see a lot of misused `rm` commands in dockerfiles, it makes no sense to spend the effort to create secondary locations.

PR #929, opened 2023-09-19
SONARIAC-1103 `ArgumentSplitter` shouldn't split command by separator inside quotes

petertrr:


mstachniuk (commented):
It looks very promising, so let's try this direction. Please check a few suggestions.

mstachniuk:
Please also add some examples to the Ruling test. It's hard to validate this approach in peach.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/symbols/ArgumentResolutionTest.java
```diff
@@ -166,6 +172,65 @@ void shouldNotDeadLoopWhenResolvingSelfAssignedVariable() {
...
+      "RUN [\"echo\", \"$FOO\"]#foo",
+      "RUN [\"echo\", \"'$FOO'\"]#'foo'",
+      "RUN [\"echo\", \"${FOO}\"]#foo",
+      "RUN [\"echo\", \"'${FOO}'\"]#'foo'",
+      // in ExecForm quotes have to be escaped and we are not un-escaping them
+      "RUN [\"echo\", \"\\\"$FOO\\\"\"]#\\\"foo\\\"",
+      "RUN [\"echo\", \"\\\"${FOO}\\\"\"]#\\\"foo\\\"")
+      .map(s -> "FROM scratch\nARG FOO=\"foo\"\n" + s)
+      .map(s -> s.split("#"))
+      .map(Arguments::of);
```
mstachniuk, line 215: Interesting idea how to handle it, but maybe try this:
```
return Stream.of(
      // argument of `echo` is StringLiteral
      Arguments.of("RUN echo foo", "foo"), ....
```
And the logic from line 213 can be moved to the test method. Then it is much clearer what is input and expected result.

iac-extensions/docker/src/test/java/org/sonar/iac/docker/symbols/ArgumentResolutionTest.java
```diff
@@ -166,6 +172,65 @@ void shouldNotDeadLoopWhenResolvingSelfAssignedVariable() {
...
+      "RUN echo ${FOO}#foo",
+      "RUN echo \"${FOO}\"#\"foo\"",
+      "RUN echo '${FOO}'#'${FOO}'",
+      "RUN echo \"\"${FOO}\"\"#\"\"foo\"\"",
+      // in ExecForm all arguments in an array are ExpandableStringLiterals
+      "RUN [\"echo\", \"$FOO\"]#foo",
+      "RUN [\"echo\", \"'$FOO'\"]#'foo'",
+      "RUN [\"echo\", \"${FOO}\"]#foo",
+      "RUN [\"echo\", \"'${FOO}'\"]#'foo'",
+      // in ExecForm quotes have to be escaped and we are not un-escaping them
```
mstachniuk, line 210: Does make it sense to unescape quotes in the case of ExecForm? Then we will not need to care about where the value comes from. 🤔

iac-extensions/docker/src/test/java/org/sonar/iac/docker/symbols/ArgumentResolutionTest.java
```diff
@@ -166,6 +172,65 @@ void shouldNotDeadLoopWhenResolvingSelfAssignedVariable() {
...
+
+    List<ArgumentResolution> argumentResolutions = CheckUtils.resolveInstructionArguments(TreeUtils.firstDescendant(file, RunInstruction.class).get());
+    ArgumentResolution stringArgument = argumentResolutions.get(1);
+
+    assertThat(stringArgument.value()).isEqualTo(expectedSecondArgument);
+  }
+
+  static Stream<Arguments> shouldCorrectlyHandleQuotes() {
+    return Stream.of(
+      // argument of `echo` is StringLiteral
```
mstachniuk, line 188: I like all this cases

mstachniuk (approved):
LGTM, a few nitpicks

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/FlagNoSpaceArgumentPredicate.java
```diff
@@ -30,15 +30,18 @@
  */
 public class FlagNoSpaceArgumentPredicate implements Predicate<ArgumentResolution> {
   private final String flag;
+  private final boolean ignoreQuotes;
 
-  public FlagNoSpaceArgumentPredicate(String flag) {
+  public FlagNoSpaceArgumentPredicate(String flag, boolean ignoreQuotes) {
```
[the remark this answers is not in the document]
petertrr: I added it for consistency, because all other predicates are wrapped into `*IgnoreQuotes`. But I agree, as it is always set to true, I'll remove the flag

iac-extensions/docker/src/test/java/org/sonar/iac/docker/symbols/ArgumentResolutionTest.java
petertrr: I agree that it can be simplified, but I didn't want to repeat `Arguments.of` on every line. And with static import simple `of()` won't be very clear IMO

iac-extensions/docker/src/test/java/org/sonar/iac/docker/symbols/ArgumentResolutionTest.java
petertrr: I think it would make sense to unescape them during argument resolution, because then we will have access to actual string values that will be passed to the launched process. I'll create a ticket for this topic.

PR #932, opened 2023-09-20
SONARIAC-1097 S6597 should not raise an issue when using HEREDOC

jonas-wielage-sonarsource:


iac-extensions/docker/src/test/resources/checks/WorkdirInsteadCdCheck/workdir_mixed.dockerfile
```diff
@@ -0,0 +1,34 @@
...
+#    ^^^^^^^^^^^^
+
+# Noncompliant@+1
+ENTRYPOINT ["/entrypoint.sh", "&&", "cd", "/tmp"]
+#                                   ^^^^^^^^^^^^
+
+# Currently we don't raise issues on any heredoc
+RUN <<"EOT"
+  cd foo/bar
+EOT
```
petertrr, line 28:
```suggestion
EOF
```

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WorkdirInsteadCdCheck.java
```diff
@@ -61,6 +62,13 @@ private static void check(CheckContext ctx, CommandInstruction commandInstructio
     }
   }
 
+  private static void checkRunInstruction(CheckContext ctx, RunInstruction runInstruction) {
```
petertrr, line 65: WDYT about introducing a high-order function in CheckUtils like `ignoringHeredoc(BiConsumer<CheckContext, RunInstruction>)` and using it directly in `init.register`? I think this could unify all the places where we will ignore heredocs and make it easier to discover them once we improve heredoc support.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/api/RunInstruction.java
```diff
@@ -28,5 +28,14 @@
...
+
+  /**
+   * @return the options as a list of {@link Flag}s.
+   */
   List<Flag> options();
+
+  /**
+   * @return true if the arguments of this RunInstruction are in {@link HereDocument} format.
+   */
+  boolean containsHeredoc();
```
petertrr, line 40: Maybe introduce a enum `HasArguments.Kind` and return kind here instead? If we need to distinguish exec form and shell form, we would need to introduce two more boolean methods.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/WorkdirInsteadCdCheck.java
jonas-wielage-sonarsource: Really nice idea!

petertrr (approved):
LGTM!

PR #933, opened 2023-09-20
Fix code smells discovered by SonarQube

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
```diff
@@ -227,42 +228,4 @@ private static ArgumentResolution getLastArgument(CommandDetector.Command comman
     List<ArgumentResolution> arguments = command.getResolvedArguments();
     return arguments.get(arguments.size() - 1);
   }
-
-  private static class HtpasswdDetector {
```
petertrr, line 231: Too long nested class; SQ suggested to extract it

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpecificVersionTagCheck.java
```diff
@@ -71,7 +70,7 @@ private static boolean hasSensitiveVersionTag(String fullImageName) {
     } else if (fullImageName.contains(":")) {
       // raise an issue if the version tag is "latest"
       String[] splitImageName = fullImageName.split(":");
-      return splitImageName.length >= 2 && splitImageName[1].equals("latest");
+      return splitImageName.length > 1 && "latest".equals(splitImageName[1]);
```
petertrr, line 73: SQ won't call `2` a magic number if you substitute it for `1` ¯\\_ツ_/¯

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -31,23 +31,17 @@
...
 import org.sonar.iac.docker.checks.utils.command.SingularPredicate;
 import org.sonar.iac.docker.symbols.ArgumentResolution;
 
 import static org.sonar.iac.docker.checks.utils.ArgumentResolutionSplitter.splitCommands;
-import static org.sonar.iac.docker.checks.utils.command.CommandPredicate.Type.MATCH;
-import static org.sonar.iac.docker.checks.utils.command.CommandPredicate.Type.NO_MATCH;
-import static org.sonar.iac.docker.checks.utils.command.CommandPredicate.Type.OPTIONAL;
-import static org.sonar.iac.docker.checks.utils.command.CommandPredicate.Type.ZERO_OR_MORE;
-import static org.sonar.iac.docker.checks.utils.command.PredicateContext.Status.ABORT;
-import static org.sonar.iac.docker.checks.utils.command.PredicateContext.Status.FOUND_NO_PREDICATE_MATCH;
```
petertrr, line 50: There is a limit on a number of static imports in a file

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
```diff
@@ -163,24 +157,16 @@ private void addCommandPredicate(CommandPredicate commandPredicate) {
       predicates.add(commandPredicate);
     }
 
-    private void addSingularPredicate(Predicate<String> predicate, CommandPredicate.Type type) {
+    private void addSingularPredicate(Predicate<String> predicate, Type type) {
       addCommandPredicate(new SingularPredicate(predicate, type));
     }
 
-    private void addOptionPredicate(SingularPredicate flag, SingularPredicate value) {
```
petertrr, line 170: These removed methods are dead code that was causing low test coverage

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
rudy-regazzoni-sonarsource: While I agree it's not ideal to have too long nested class, I'm not sure if it's a good idea to extract it, because to me this class behavior is really specific to this rule, checking precise values such as `htpasswd`, `-b` and `-n`.
By extracting it outside, we send the implicit message that it can be useful in other checks, which seems unlikely to me.
Also, I didn't think 36 lines to be that long :sweat_smile:
WDYT?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
rudy-regazzoni-sonarsource: @mstachniuk will be happy to see that the `CommandDetector` refactoring has started :smile:

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my few comments, I don't have a strong opinion on them but I wanted to highlight some arguments to ensure we make the right decision on some elements.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetector.java
jonas-wielage-sonarsource: Is this really valuable to remove?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/shell/HtpasswdDetector.java
jonas-wielage-sonarsource, line 1: I would challenge this class as it's really specific to one check

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SecretsGenerationCheck.java
petertrr: On the one hand I agree that it doesn't make sense to make this logic public so to say. On the other hand, that class where it's used is already pretty long and IMO could benefit from being split.

petertrr: But yeah, I've decided to rollback this extraction after all

PR #941, opened 2023-09-27
Prepare for next development iteration

petertrr:


iac-common/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.sonarsource.iac</groupId>
     <artifactId>iac</artifactId>
-    <version>1.21-SNAPSHOT</version>
+    <version>1.22.0-SNAPSHOT</version>
```
mstachniuk, line 9: It should be `1.22-SNAPSHOT` (without `.0`)

mstachniuk (changes requested):

mstachniuk (approved):
LGTM

PR #942, opened 2023-10-09
SONARIAC-1112 Improve CommandDetector, change SingularPredicate const…

mstachniuk:
…ructor


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CommandDetectorBuilder.java
rudy-regazzoni-sonarsource, line 1: Good idea moving this to another file.

rudy-regazzoni-sonarsource (approved):
LGTM!

PR #958, opened 2023-11-15
Remove duplicate plugin exec-maven-plugin

rudy-regazzoni-sonarsource:


iac-extensions/helm/pom.xml
```diff
@@ -88,39 +88,35 @@
             <id>compile-go-calc</id>
             <configuration>
               <workingDirectory>src/main/go/calc</workingDirectory>
+              <executable>go</executable>
```
rudy-regazzoni-sonarsource, line 91: Careful here: I'm using `go` instead of `go1.19`
I can remote it in this PR, unless we all instead use `go`

petertrr: Using just `go` should indeed be preferred; I'll fix it on my system.

petertrr (approved):

PR #965, opened 2023-11-27
SONARIAC-1166 Evaluate basic Helm templates in Go

mstachniuk:


sonar-helm-for-iac/Readme.md
```diff
@@ -44,3 +44,11 @@ go test
 ```shell
 sonar-helm-for-iac.exe
 ```
+
+## Tricks and tips
+
+### missing go.sum entry for module providing package
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Also I don't quite understand the wording `Module providing package`

sonar-helm-for-iac/main.go
```diff
@@ -1,9 +1,92 @@
...
 )
 
 func main() {
 	fmt.Println("Hello World!")
 }
+
+var handles []*template.Template
+
+//export EvaluateTemplate
+func EvaluateTemplate(path string, content string, valuesFileContent string) *C.char {
```
jonas-wielage-sonarsource, line 37: Maybe we should think about how we want go code formatted

petertrr: There is a built-in formatter that can be called with `go fmt`, and we can set up its execution as part of build. However, it might make sense to do it with Gradle already

sonar-helm-for-iac/Readme.md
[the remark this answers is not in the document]
mstachniuk: It's just copy paste of the error produced by `go` command.

sonar-helm-for-iac/main.go
mstachniuk: Unfortunately spotless doesn't support Go code :-( I agree we should think about it.

sonar-helm-for-iac/Readme.md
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Okay, then I would maybe add a `Exception during build: "missing go.sum entry for module providing package"`.

sonar-helm-for-iac/main.go
petertrr: Spotless might be capable of checking license headers, but I didn't find a clear instruction on how to set it up

jonas-wielage-sonarsource: We should create separate tickets that can be tackled after migrating to gradle

mstachniuk: Issue created: https://sonarsource.atlassian.net/browse/SONARIAC-1172 (out of scope of current sprint for now).

PR #971, opened 2023-11-29
SONARIAC-1169 Set up multi-architecture build on Cirrus CI

petertrr:
* Add two additional tasks that build Darwin and Windows artifacts inside `golang-crosscompile` containers and upload them as Cirrus artifacts
* Download these artifacts as part of build task; Linux binary is still being built as part of Maven build
* Change go buildmode to `c-shared`
* In `make.sh`, assume `GOOS` is unset on a developer machine (only set by `go` executable)


mstachniuk (approved):
LGTM

PR #976, opened 2023-11-30
KubernetesParser.java

petertrr:

PR #982, opened 2023-12-07
Revert "SONARIAC-1150 Read the evaluated template and pass it to K8s sensor"

jonas-wielage-sonarsource:
Reverts SonarSource/sonar-iac#976


jonas-wielage-sonarsource:
Closed because the root cause was reverted in this PR: https://github.com/SonarSource/sonar-iac/pull/983

PR #986, opened 2023-12-11
Fix SCA build

mstachniuk:


.cirrus.yml
```diff
@@ -178,7 +178,7 @@ sca_scan_task:
   whitesource_script:
     - source cirrus-env QA
     - source set_maven_build_version $BUILD_NUMBER
-    - mvn clean install -DskipTests
+    - mvn clean install -DskipTests -P-ci,local
```
nils-werner-sonarsource, line 181: Can you give a bit of context what the parameters are about?

mstachniuk: It's about disabling the Maven profile `ci` and enabling the `local` profile. 
The alternative solution is to do the same steps as in `build_task`, i.e., download Go binaries and unzip them.

PR #988, opened 2023-12-12
SONARIAC-1182 Build Go binaries as executables and get data from stdout

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/utils/ExecutableHelper.java
```diff
@@ -0,0 +1,79 @@
...
+    if (!dest.exists()) {
+      return false;
+    }
+    byte[] actualContent = Files.readAllBytes(dest.toPath());
+    return Arrays.equals(actualContent, expectedContent);
+  }
+
+  static byte[] getBytesFromResource(String executable) throws IOException {
+    var out = new ByteArrayOutputStream();
+    try (InputStream in = HelmEvaluator.class.getClassLoader().getResourceAsStream(executable)) {
```
petertrr, line 63: For some reason, both `ExecutableHelper.class` and `Thread.currentThread().getContextClassloader()` failed to locate resource in ITs

sonar-helm-for-iac/main.go
```diff
@@ -18,34 +18,64 @@
...
-	"unsafe"
 )
 
 func main() {
-	fmt.Println("Hello World!")
-}
+	var path string
+	var numTemplateLines int
+	flag.StringVar(&path, "path", "", "Name of the rawTemplate")
+	flag.IntVar(&numTemplateLines, "nl", 0, "Number of lines in the rawTemplate; lines after that are considered rawValues.yaml content")
```
mstachniuk, line 37: Generally speaking good idea of using stdin for passing the content of the files. However in current shape the mechanism is not extendable. What about the following solution:
Running the command without any arguments and passing by stdin information in following format:
```
templates/pod.yaml
42
<raw-data>
values.yaml
32
<raw-data>
END
```

In other words:
* name of the template
* number of bytes to read
* content of the file
* next files
* indicator that everything is passed

Thanks to that it will be easier later to pass other files without changing the interface. WDYT?

sonar-helm-for-iac/make.sh
```diff
@@ -112,16 +112,20 @@ compile_binaries() {
...
         ;;
-      "linux"|"windows")
+      "windows")
         GOARCH="amd64"
-        CGO_ENABLED=1 go build "${GO_FLAGS[@]}" -o target/classes/sonar-helm-for-iac-"$GOOS"-"$GOARCH"
+        CGO_ENABLED=0 go build "${GO_FLAGS[@]}" -o target/classes/sonar-helm-for-iac-"$GOOS"-"$GOARCH"
+        ;;
+      "linux")
+        GOARCH="amd64"
+        CGO_ENABLED=0 CC=musl-gcc go build "${GO_FLAGS[@]}" --ldflags '-linkmode external -extldflags "-s -w -static"' -o target/classes/sonar-helm-for-iac-"$GOOS"-"$GOARCH"
```
mstachniuk, line 128: Is static linking still needed if we build an executable instead of c-shared?

mstachniuk (changes requested):
Please see my few next comments

sonar-helm-for-iac/main.go
petertrr: Implemented in 57bde321148c4cfa186d26498cb346687e346ee4. You can take a look already; I'll address your other comments in a next commit

sonar-helm-for-iac/make.sh
petertrr: Without static linking, we will have a runtime dependency on libc, which may once again cause differences on glibc- vs musl- based systems. Static linking with musl procudes a self-contained executable with little to no overhead in size

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/utils/NativeUtils.java
```diff
@@ -17,38 +17,23 @@
...
-   * Load a native library. This method takes into account that we use OS and architecture as suffixes of the library name.
-   * It also takes care of mapping Go function names to Java method names not to break camelCase convention on Java side.
-   */
-  public <T extends Library> T load(String name, Class<T> libraryClass) {
-    var os = getNormalizedOsName(System.getProperty("os.name"));
-    var arch = getNormalizedArchName(System.getProperty("os.arch"));
-
-    if (!SUPPORTED_PLATFORMS.contains(os + "-" + arch)) {
-      throw new IllegalStateException("Unsupported platform: " + os + "-" + arch);
+  public String getSuffixForCurrentPlatform() {
```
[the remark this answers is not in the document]
petertrr: I'll call it `getCurrentPlatform`, as I reference this combination as platform in some other places.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/utils/NativeUtils.java
```diff
@@ -17,38 +17,23 @@
...
-import com.sun.jna.Library;
-import com.sun.jna.Native;
-import com.sun.jna.NativeLibrary;
-import java.lang.reflect.Method;
 import java.util.Locale;
-import java.util.Map;
 import java.util.Set;
 
-public class Loader {
+public class NativeUtils {
```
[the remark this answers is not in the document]
petertrr: I think the logic is not exactly the part of ExecutableHelper and it's easier to test them separately

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/utils/ExecutableHelper.java
```diff
@@ -0,0 +1,99 @@
...
+
+  public static String extractFromClasspath(File workDir, String executable) throws IOException {
+    byte[] executableData = getBytesFromResource(executable);
+    var dest = new File(workDir, executable);
+    if (!fileMatch(dest, executableData)) {
+      Files.write(dest.toPath(), executableData);
+      if (!System.getProperty("os.name").toLowerCase(Locale.ROOT).contains("windows")) {
+        var perms = Files.getPosixFilePermissions(dest.toPath());
+        perms.add(PosixFilePermission.OWNER_EXECUTE);
+        Files.setPosixFilePermissions(dest.toPath(), perms);
```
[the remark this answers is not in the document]
petertrr: Zip doesn't preserve executable flag, so to execute our binary we need to restore this flag on an unpacked file

sonar-helm-for-iac/main.go
```diff
@@ -18,34 +18,64 @@
...
+	}
+	return rawInput
+}
+
+func bytesToString(input [][]byte) string {
+	result := make([]string, len(input))
+	for i, b := range input {
+		result[i] = string(b)
+	}
+	return strings.Join(result, "\n")
```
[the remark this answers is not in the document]
petertrr: On the Go side, `Scanner.Scan()` returns false once it encounters `\r?\n`. So it should split the lines correctly. I'm not sure how we account for line breaks when calculating text locations. Should line breaks passed from the Java side always be `\n` or platform-dependent? I believe the latter?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/utils/ExecutableHelper.java
jonas-wielage-sonarsource: Yes we have the same issue in SonarText and we just `<Classname>.class.getClassLoader().`

mstachniuk: OK but then the current class `ExecutableHelper.class` should be good

sonar-helm-for-iac/main.go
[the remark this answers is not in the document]
mstachniuk: Actually, they should not be platform-depended but user-file-depended.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/HelmEvaluator.java
```diff
@@ -50,4 +72,43 @@ public TemplateEvaluationResult evaluateTemplate(String path, String content, St
...
+
+  Process startProcess(ProcessBuilder pb, String name, String content, String valuesFileContent) throws IOException {
+    var process = pb.start();
+    try (var os = process.getOutputStream()) {
+      os.write(String.format("%s%n", name).getBytes(StandardCharsets.UTF_8));
+      os.write(String.format("%d%n", content.lines().count()).getBytes(StandardCharsets.UTF_8));
+      if (!content.endsWith("\n")) {
+        content += "\n";
+      }
+      os.write(content.getBytes(StandardCharsets.UTF_8));
```
mstachniuk, line 90: Nitpick. It can be extracted to separate method, the next lines are similar

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/HelmEvaluator.java
```diff
@@ -50,4 +72,43 @@ public TemplateEvaluationResult evaluateTemplate(String path, String content, St
...
+    }
+    return process;
+  }
+
+  private static void monitorProcess(Process process) {
+    try {
+      if (!process.waitFor(PROCESS_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        LOG.debug(HELM_FOR_IAC_EXECUTABLE + " is taking longer than 5 seconds to finish");
+        process.destroy();
+        process.waitFor();
```
mstachniuk, line 107: Is this method needed here? I guess the `process.destroy(); ` is enough

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/HelmEvaluator.java
```diff
@@ -50,4 +72,43 @@ public TemplateEvaluationResult evaluateTemplate(String path, String content, St
...
+      }
+      os.write(valuesFileContent.getBytes(StandardCharsets.UTF_8));
+      os.write(String.format("END%n").getBytes(StandardCharsets.UTF_8));
+    }
+    return process;
+  }
+
+  private static void monitorProcess(Process process) {
+    try {
+      if (!process.waitFor(PROCESS_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
```
mstachniuk, line 104: I ❤️ using timeout here

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/utils/ExecutableHelper.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return dest.getAbsolutePath();
+  }
+
+  @CheckForNull
+  public static byte[] readProcessOutput(Process process) {
+    try (var is = process.getInputStream(); var es = process.getErrorStream()) {
+      var rawEvaluationResult = is.readAllBytes();
+      new BufferedReader(new InputStreamReader(es, StandardCharsets.UTF_8)).lines()
+        .forEach(line -> LOG.debug("[exec] {}", line));
```
mstachniuk, line 65: The prefix in the debug log should be more describable, e.g. `[sonar-helm-for-iac stderr]`

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/HelmEvaluatorTest.java
```diff
@@ -20,48 +20,88 @@
...
+
+  @Test
+  void shouldThrowIfGoBinaryReturnsNonZero() {
     Assertions.assertThatThrownBy(() -> helmEvaluator.evaluateTemplate("/foo/bar/baz.yaml", "", ""))
       .isInstanceOf(IllegalStateException.class)
-      .hasMessage("Empty evaluation result (serialization failed?)");
+      .hasMessage("sonar-helm-for-iac exited with non-zero exit code: 1");
+
+    Assertions.assertThat(logTester.logs(Level.DEBUG))
+      // TODO: better handling of empty input
```
mstachniuk, line 66: What about this TODO?

iac-extensions/kubernetes/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,18 @@
+<configuration>
```
mstachniuk, line 1: Do we need that? We are using `LogTesterJUnit5 `

sonar-helm-for-iac/org.sonarsource.iac.helm/io-handling.go
```diff
@@ -0,0 +1,74 @@
+package org_sonarsource_iac_helm
```
mstachniuk, line 1: IMO We should use this package only for generated files and the rest should be in sonar-helm-for-iac or in other package

mstachniuk (approved):
LGTM I added some small comments, they can be handled in separate PR

iac-extensions/kubernetes/src/test/resources/logback-test.xml
petertrr: I was using it to see logs of the process during debugging, but I guess it can be safely removed now

sonar-helm-for-iac/org.sonarsource.iac.helm/io-handling.go
petertrr: I don't really like the mix of pom.xml, scripts, Go files etc. in the root package; I would prefer to split the Go code more into subpackages. I'll try to rename it then

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/HelmEvaluatorTest.java
petertrr: This was about the line `Expected 2 files, received 3`, when the input is in fact completely empty. But for now I don't want to overcomplicate Go logic to handle such rare cases; I think the messages `Skipping request to read 0 lines` before this line make it a bit more clear

PR #992, opened 2023-12-13
SONARIAC-1184 Support "default" function in Helm template evaluation

jonas-wielage-sonarsource:


sonar-helm-for-iac/main_test.go
```diff
@@ -276,3 +276,43 @@ func Test_to_protobuf_invalid(t *testing.T) {
 	assert.Equal(t, "", templateFromProto.Template)
 	assert.Equal(t, "template: a.yaml:1: unclosed action", templateFromProto.Error)
 }
+
+func Test_evaluate_template_default_function(t *testing.T) {
```
jonas-wielage-sonarsource, line 280: Not sure if we need this test, but at the beginning it was not working for me in java, so I tested it internally in go code first.

mstachniuk: It's ok to have extra test here

mstachniuk (approved):
LGTM

PR #1002, opened 2023-12-18
Wait for external process also after reading from it is finished

petertrr:
Previous logic:
* Start a background thread that waits for 5 seconds for the process to finish and then kills it
* Perform IO exchange with the process

My hypothesis is that test may sometimes finish earlier than background thread kills the process of OS (Windows) releases all the resources after process exits. Hence junit is trying to delete an executable which is still in use by the OS.

New logic:
* Start background thread that waits for 5 seconds for the process to finish and then kills it
* Perform IO exchange with the process
* After it's done, wait for the process to finish from the main thread as well
 * If data is parsed w/o errors, but the process is forcibly killed from the background thread, don't throw an exception


petertrr:
This PR is too outdated at this point, and tests on Windows are no longer flaky, so I'm closing it

PR #1009, opened 2023-12-20
SONARIAC-1217 Change Helm evaluation errors to raise a parsing issue

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
```diff
@@ -51,38 +53,50 @@ public void initialize() {
...
-      if (valuesFile != null) {
-        try {
-          return evaluateHelmTemplate(filename, sourceWithComments, valuesFile.contents());
-        } catch (IOException e) {
-          LOG.debug("Failed to read values file at {}, skipping processing of Helm file '{}'", valuesFile, inputFileContext.inputFile, e);
-        }
-      } else {
-        LOG.debug("Failed to find values file, skipping processing of Helm file '{}'", inputFileContext.inputFile);
-      }
+    if (helmEvaluator == null) {
```
jonas-wielage-sonarsource, line 61:
```suggestion
    if (!isHelmEvaluatorInitialized()) {
```

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
```diff
@@ -51,38 +53,50 @@ public void initialize() {
...
+    if (valuesFile == null) {
+      throw new ParseException("Failed to find values file, skipping processing of Helm file " + inputFile,
+        inputFile.newPointer(0, 0), null);
+    }
+
+    String valuesFileContent;
+    try {
+      valuesFileContent = valuesFile.contents();
+    } catch (IOException e) {
+      throw new ParseException("Failed to read values file at " + valuesFile + " while evaluating Helm file " + inputFile,
```
jonas-wielage-sonarsource, line 82: I would change this message to be in line with the other exception ("skipping processing of Helm file...").
This way you could create a method
```private void throwParseExceptionFor(InputFile inputFile, String customMessagePrefix, @Nullable String cause)```
as you are always raising on the same range and nearly with the same message.
WDYT?

File: iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java [hunk 3]
```diff
@@ -47,21 +47,28 @@ public KubernetesParser(HelmProcessor helmProcessor) {
...
-        }
-      } else {
-        LOG.debug("No InputFileContext provided, skipping processing of Helm file");
-      }
+    if (!hasHelmContent(source)) {
+      return super.parse(source, inputFileContext);
+    } else {
+      return parseHelmFile(source, inputFileContext);
+    }
+  }
```
jonas-wielage-sonarsource, line 55: I like this approach!

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -47,21 +47,28 @@ public KubernetesParser(HelmProcessor helmProcessor) {
...
 
-      return super.parse("{}", inputFileContext, FileTree.Template.HELM);
+  private FileTree parseHelmFile(String source, @Nullable InputFileContext inputFileContext) {
+    if (inputFileContext == null) {
+      LOG.debug("No InputFileContext provided, skipping processing of Helm file");
+      return super.parse("{}", null, FileTree.Template.HELM);
```
jonas-wielage-sonarsource, line 60: Not related to this ticket, but I'm really unsure why we are always null checking here. I checked the whole project, and this may be something we could improve overall... (But this should not be done here)

jonas-wielage-sonarsource (approved):
LGTM (after fixing plugin tests), please see my suggestions

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
petertrr: Yes, I also thought about it. I found that we sometimes pass `null` there in Docker, but not in other extensions

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
petertrr: Good suggestion!

PR #1013, opened 2023-12-21
SONARIAC-1223 Improve error handling in Go exceptions

petertrr:


sonar-helm-for-iac/converters/io-handling.go
```diff
@@ -39,37 +41,56 @@ type InputReader interface {
...
 	}
+
 	for firstLine != "END" {
 		name := firstLine
-		length, _ := strconv.Atoi(s.readInput(scanner, 1))
-		fmt.Fprintf(os.Stderr, "Reading %d lines from stdin\n", length)
-		content := s.readInput(scanner, length)
+
+		var content string
+		contentResult := mo.TupleToResult(s.readInput(scanner, 1)).FlatMap(
```
petertrr, line 64: So the idea here is to read 3 different things from stdin and handle errors. Handling errors will be the same all the time: abort reading, return an error upward. However, with vanilla Go it requires too much boilerplate. Using `mo.Result` allows to declare set of operations, and `err` will be set once first failure occurs; all other operations won't be executed.
The library is not perfect, e.g. doesn't allow mapping between different types, but may come in handy

sonar-helm-for-iac/converters/io-handling.go
```diff
@@ -39,37 +41,56 @@ type InputReader interface {
...
 
 // readInput
 // Reads nLines from the given scanner and returns as a single string.
 // If nLines is negative, reads all lines until EOF.
-func (s StdinReader) readInput(scanner *bufio.Scanner, nLines int) string {
+func (s StdinReader) readInput(scanner *bufio.Scanner, nLines int) (string, error) {
 	if nLines == 0 {
-		fmt.Fprintf(os.Stderr, "Skipping request to read 0 lines\n")
-		return ""
+		return "", errors.New("won't read 0 lines")
```
jonas-wielage-sonarsource, line 93: Maybe we can improve here with the message. Is there a specific cause that can happen why it's reading 0 lines?
`Request to read 0 lines aborted`

sonar-helm-for-iac/converters/io-handling_test.go
```diff
@@ -80,11 +82,59 @@ func Test_read_one_file_with_trailing_newline(t *testing.T) {
...
 
 	assert.Equal(t, 2, len(contents))
 	assert.Equal(t, "file1", contents[0].Name)
 	assert.Equal(t, "line1\nline2", contents[0].Content)
 	assert.Equal(t, "file2", contents[1].Name)
 	assert.Equal(t, "line3", contents[1].Content)
 }
+
+func Test_should_stop_if_zero_length(t *testing.T) {
+	scanner := bufio.NewScanner(strings.NewReader("file1\n0\n"))
```
jonas-wielage-sonarsource, line 95: Clarification question: Is it always a given that we don't end in `\n` in a template content? Otherwise we would fail quite often (what we don't do, so I think that's a given?)

sonar-helm-for-iac/main.go
```diff
@@ -32,19 +33,17 @@ var stdinReader converters.InputReader = converters.StdinReader{}
 var serializer converters.Serializer = converters.ProtobufSerializer{}
 
 func main() {
-	scanner := bufio.NewScanner(os.Stdin)
-	contents := stdinReader.ReadInput(scanner)
-	if code := validateContents(contents); code != 0 {
-		os.Exit(code)
-	}
+	path, rawTemplate, rawValues, processingError := prepareData()
```
jonas-wielage-sonarsource, line 36: Returning tuples in functions would be really nice to have in java...

sonar-helm-for-iac/main.go
```diff
@@ -53,15 +52,29 @@ func main() {
 	os.Stdout.Write(result)
 }
 
-func validateContents(contents []converters.Content) int {
+func prepareData() (string, string, string, error) {
+	scanner := bufio.NewScanner(os.Stdin)
+	contents, err := stdinReader.ReadInput(scanner)
+	if err != nil {
+		return "", "", "", fmt.Errorf("read input: %w", err)
```
jonas-wielage-sonarsource, line 59:
```suggestion
		return "", "", "", fmt.Errorf("Exception during content reading:` %w", err)
```

sonar-helm-for-iac/main.go
```diff
@@ -53,15 +52,29 @@ func main() {
...
+	}
+	if err = validateContents(contents); err != nil {
+		return "", "", "", fmt.Errorf("validate contents: %w", err)
```
jonas-wielage-sonarsource, line 62:
```suggestion
		return "", "", "", fmt.Errorf("Exception during content validation: %w", err)
```

sonar-helm-for-iac/main.go
```diff
@@ -91,7 +104,7 @@ func newHandleID(name string, content string) (int, error) {
 func executeWithValues(templateId int, valuesFileContent string) (string, error) {
 	valuesMap, err := yamlToMap(valuesFileContent)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("error parsing values file: %w", err)
```
jonas-wielage-sonarsource, line 107:
```suggestion
		return "", fmt.Errorf("Error during parsing of values file: %w", err)
```

sonar-helm-for-iac/main.go
```diff
@@ -53,15 +52,29 @@ func main() {
...
+
+func validateContents(contents []converters.Content) error {
 	if len(contents) == 0 {
-		fmt.Fprintf(os.Stderr, "Received empty input, exiting\n")
-		return 1
+		return errors.New("received empty input")
 	} else if len(contents) != 2 {
-		fmt.Fprintf(os.Stderr, "Expected 2 files, received %d (values.yaml missing?)\n", len(contents))
-		return 1
+		return fmt.Errorf("expected 2 files, received %d (values.yaml missing?)", len(contents))
```
jonas-wielage-sonarsource, line 75:
```suggestion
		return fmt.Errorf("Expected 2 files, received %d files, possible missing values file", len(contents))
```

sonar-helm-for-iac/main.go
```diff
@@ -53,15 +52,29 @@ func main() {
...
+	rawTemplate := contents[0].Content
+	rawValues := contents[1].Content
+	return path, rawTemplate, rawValues, nil
+}
+
+func validateContents(contents []converters.Content) error {
 	if len(contents) == 0 {
-		fmt.Fprintf(os.Stderr, "Received empty input, exiting\n")
-		return 1
+		return errors.New("received empty input")
```
jonas-wielage-sonarsource, line 73:
```suggestion
		return errors.New("No input received")
```

sonar-helm-for-iac/converters/io-handling_test.go
jonas-wielage-sonarsource: Ok, looking at HelmEvaluator, we're ending with `END%n`, that's why

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/HelmEvaluatorTest.java
```diff
@@ -66,14 +66,29 @@ static void cleanup() throws IOException {
...
-      .hasMessage("sonar-helm-for-iac exited with non-zero exit code: 1");
+      .hasMessage("Evaluation error in Go library: read input: won't read 0 lines");
 
     Assertions.assertThat(logTester.logs(Level.DEBUG))
-      .contains("[sonar-helm-for-iac] Skipping request to read 0 lines")
-      .contains("[sonar-helm-for-iac] Expected 2 files, received 3 (values.yaml missing?)");
+      .contains("[sonar-helm-for-iac] Reading 0 lines from stdin");
+  }
+
+  @Test
```
jonas-wielage-sonarsource, line 78: Unrelated to this line, but to the HelmEvaluator.
I would change `Empty evaluation result (serialization failed?)` to
`Empty evaluation result received, possible serialization failure in Go library`

jonas-wielage-sonarsource (approved):

sonar-helm-for-iac/main.go
petertrr: The reason to make these messages different from what we're used to in Java, is the way they are later handled. We basically get a long string in a form `message1: message2: ... : messageN`.  And eventually we'll see something like
`DEBUG: Evaluation error in Go library: error parsing values file: error converting YAML to JSON: yaml: line 6: did not find expected key`.
But thinking about it, they can be made more verbose. However, Go plugin suggests to start them with lowercase letter, and IMO it makes sense considering all the information is eventually concatenated

jonas-wielage-sonarsource: I think a bit more verbose would be nice, to better pinpoint the errors (and making it more easily readable).
But I agree, they can get quite long. (But I think that's okay, because it's only debug messages for people that are interested).
Starting lower case is ok

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/HelmEvaluatorTest.java
petertrr: Yes, good suggestion. Actually, now in case serialization fails go binary should exit with code 1, so this branch covers unexpected empty results. I'll adjust the message accordingly

sonar-helm-for-iac/converters/io-handling.go
petertrr: Only if we pass empty text from HelmEvaluator

PR #1017, opened 2023-12-22
SONARIAC-1230 Align AzureResourceManager property keys and deprecate old key format

jonas-wielage-sonarsource:


its/plugin/src/test/java/org/sonarsource/iac/PropertiesTest.java
```diff
@@ -105,6 +105,31 @@ void testArmSuffix(String projectKey, String language, String suffixes, int expe
     executeBuildAndAssertMetric(projectKey, language, "suffixes", suffixes, "ncloc", expectedNcloc);
   }
 
+  // The active property will always win over the deprecated one if set
+  @ParameterizedTest
+  @CsvSource(delimiter = ';', value = {
+    "armDeprecatedPropertyTestBothDisabled; false; false;",
```
petertrr, line 111: I wonder if we need so many integration tests for this. Is it possible to write a test where only configuration is loaded, or are these details (which property takes precedence) not easily accessible from plugin API?

petertrr (approved):
LGTM, but also please see my question

its/plugin/src/test/java/org/sonarsource/iac/PropertiesTest.java
jonas-wielage-sonarsource: I don't think we need that many that are executed everytime. I used this also to make sure that they work as I expect.

jonas-wielage-sonarsource: But they are really cheap to execute as the Orchestrator is already running

jonas-wielage-sonarsource: I will remove some of them

PR #1019, opened 2023-12-27
SONARIAC-1231 Evaluate all files of Chart directory in Go Engine

petertrr:


sonar-helm-for-iac/main.go
```diff
@@ -36,25 +36,41 @@ var serializer converters.Serializer = converters.ProtobufSerializer{}
 type TemplateSources struct {
 	Name        string
 	RawTemplate string
-	RawValues   string
+	sources     map[string]string
```
mstachniuk, line 39: Maybe `files` will be a better name than `sources`? 🤔

mstachniuk (approved):
LGTM

PR #1020, opened 2023-12-28
SONARIAC-1225 Pass all files of the Helm project directory to HelmEvaluator

jonas-wielage-sonarsource:

PR #1024, opened 2023-12-28
SONARIAC-1201 Support built-in Helm objects in template evaluation

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -68,6 +68,7 @@ private FileTree parseHelmFile(String source, @Nullable InputFileContext inputFi
     }
 
     var filename = inputFileContext.inputFile.filename();
+    // TODO: pass "namespaced file path" of the template instead of filename
```
petertrr, line 71: Makes sense to add in/after #1020

sonar-helm-for-iac/converters/data-loading.go
```diff
@@ -0,0 +1,99 @@
...
+
+func PrepareChartValues(templateSources *TemplateSources) (map[string]interface{}, error) {
+	result := map[string]interface{}{}
+
+	values := mo.TupleToResult(LoadValues(templateSources.Values()))
+	chart, err := Map(values, func(values *Values) (*Chart, error) {
+		return Map(mo.TupleToResult(templateSources.SourceFile("Chart.yaml")), func(content string) (*Chart, error) {
+			return LoadChart(content)
+		}).Get()
+	}).Get()
```
jonas-wielage-sonarsource, line 45: Why not evaluate this function prior to the return?

sonar-helm-for-iac/converters/helm_structs.go
```diff
@@ -0,0 +1,45 @@
...
+const k8sVersionMajor = "1"
+const k8sVersionMinor = "20"
+
+var DefaultCapabilities = map[string]interface{}{
+	"KubeVersion": struct {
+		Version string
+		Major   string
+		Minor   string
+	}{
+		fmt.Sprintf("v%s.%s.0", k8sVersionMajor, k8sVersionMinor),
```
jonas-wielage-sonarsource, line 33: Isn't it shifted a place? seems to me like `Version` is initialized by `k8sVersionMajor`, `Major` is initialized by `k8sVersionMinor`, and `Minor` is initialized by 0.

sonar-helm-for-iac/converters/helm_structs.go
```diff
@@ -0,0 +1,45 @@
...
+		k8sVersionMajor,
+		k8sVersionMinor,
+	},
+	// It is possible to directly use `k8s.io/apimachinery/pkg/runtime/scheme.go`, as does Helm. Not sure if we need it.
+	"APIVersions": []string{"v1"},
+	"HelmVersion": "v3",
```
jonas-wielage-sonarsource, line 39: Don't we need the strcut to contain `HelmVersion.Version`, `HelmVersion.GitCommit`, `HelmVersion.GitTreeState` and `HelmVersion.GoVersion`?

sonar-helm-for-iac/converters/data-loading_test.go
```diff
@@ -0,0 +1,58 @@
...
+		"apiVersion: v1",
+		map[string]string{"values.yaml": "foo: bar", "Chart.yaml": "name: test-project"})
+
+	chartData, _ := PrepareChartValues(templateSources)
+
+	assert.Contains(t, chartData, "Values")
+	assert.Contains(t, chartData, "Chart")
+	assert.Contains(t, chartData, "Capabilities")
+	assert.Contains(t, chartData, "Release")
+	assert.Contains(t, chartData, "Template")
```
jonas-wielage-sonarsource, line 40: I feel like we could benefit from an additional test (not necessary in this file), whether the input of the values of the map contain the correct information ( like checking if capabilities is equal to the struct capabilities)

sonar-helm-for-iac/helm-imports.go
```diff
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"github.com/BurntSushi/toml"
 	"github.com/Masterminds/sprig/v3"
+	"github.com/SonarSource/sonar-iac/sonar-helm-for-iac/converters"
```
jonas-wielage-sonarsource, line 27: Why is this `github.com/...` necessary` and we can simply import `converters`? Is it the fqn like in java?

jonas-wielage-sonarsource (approved):
LGTM (almost, but want to unblock a possible merge tomorrow). 
Please see my suggestions and questions.
I would also like to see an additional integration test in ruling, where all the additional structs are used (capabilities, release), to see it not resulting in a parsing error :)

sonar-helm-for-iac/converters/data-loading.go
petertrr: For now this approach with `mo.Result` seems to me a bit better than vanilla error handling in Go, which in this case would require multiple repetitive blocks, because several steps here can return errors preventing the evaluation. So wrapping everything into a chain of `Result`s is a way to handle  errors in a single place.

sonar-helm-for-iac/converters/helm_structs.go
petertrr: No, I copied this from Helm. `Version` is a full version, `Major` and `Minor` are major and minor, and patch is always set to 0

sonar-helm-for-iac/helm-imports.go
petertrr: Yes, like FQN, which is the same both when you include this module in the same project or in another

PR #1027, opened 2024-01-02
SONARIAC-1177 Implement "tpl" function for Helm template evaluation

petertrr:


jonas-wielage-sonarsource (commented):
Just a small comment: A its test-case would be nice to see that we don't raise parsing issue

PR #1033, opened 2024-01-03
SONARIAC-1246 Make Go compilation tasks compatible with Gradle cache

petertrr:


rudy-regazzoni-sonarsource (approved):
LGTM! As discussed I let you check how you want to handle the promote speed-up part, either in this PR or another.

PR #1042, opened 2024-01-03
SONARIAC-1252 Separate CI build from test and analysis

nils-werner-sonarsource:


.cirrus.yml
```diff
@@ -142,10 +142,29 @@ build_task:
     - source cirrus-env BUILD
     - source .cirrus/use-gradle-wrapper.sh
     - go version
-    - regular_gradle_build_deploy_analyze --profile
+    - regular_gradle_build_deploy_analyze --profile -x test -x sonar
```
rudy-regazzoni-sonarsource, line 145: I like the idea of splitting the build and the test, this should speedup the whole process by parallelizing test and ITs.
Also, it reduces a lot of the logs in the build-only  phase (from 9'000+ lines to 3'000).

.cirrus.yml
```diff
@@ -142,10 +142,29 @@ build_task:
...
+  env:
+    <<: *BUILD_SECRETS
+    #allow deployment of pull request artifacts to repox
+    DEPLOY_PULL_REQUEST: true
+    GO_CROSS_COMPILE: '0'
+  build_script:
+    - source cirrus-env BUILD
+    - source .cirrus/use-gradle-wrapper.sh
+    - go version
+    - regular_gradle_build_deploy_analyze -x artifactoryPublish
```
rudy-regazzoni-sonarsource, line 164: By excluding this task, it removes the following tasks:
```
:artifactoryPublish
:sonar-iac-plugin:generatePomFileForMavenJavaPublication
:sonar-iac-plugin:signMavenJavaPublication
:sonar-iac-plugin:artifactoryPublish
:extractModuleInfo
:sonar-iac-plugin:extractModuleInfo
:artifactoryDeploy
```
Out of the 224 tasks, it still executes 218 of them.

As we are building everything here in parallel of the `build` task, it makes no use of the gradle cache, we could remove the SETUP/CLEANUP part.

This is an improvement regarding the CI as it allows to perform the `build` step a bit faster (from 6m to 4m?) so it can start depending steps sooner and improve the global time, but to me this `build_test_analyze_task` is doing too much.
It would be great if it would skip the build part, a quick-win would be to depend on the build step and keep the gradle cache, so it would skip most build part by retrieving previously compiled element from the cache.
WDYT?

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my suggestion, can be tackled in another PR if it makes sense.

.cirrus.yml
nils-werner-sonarsource: The other task already starts, when the build task is done. They are not waiting for `build_test_analyze_task`. I removed the cache setup and cleaning. It will not affect the overall performance, but reduce some workload.

rudy-regazzoni-sonarsource: Yes I understand that.
Let me rephrase is: should `build_test_analyze_task` wait for `build`, so that it can benefit from gradle cache?

It's not going to reduce the CI global time, but it would reduce the number of steps executed by `build_test_analyze_task` (which could then be renaed to `test_analyze_task`), ending up reducing the logs (build logs in build step, and test/sonar logs in test/sonar step only).

rudy-regazzoni-sonarsource: But this would increase the time before we get the result/feedback from the sonar analysis...

I guess I'm just feeling unsatisfied that we build twice the same code :joy:

nils-werner-sonarsource: Fair point. I will give it a try

PR #1043, opened 2024-01-03
SONARIAC-1176 Implement "include" function for Helm template evaluation

mstachniuk:


sonar-helm-for-iac/helm-imports.go
```diff
@@ -210,7 +209,9 @@ var sprigFunctionsWhitelist = []string{
 
 var sprigFunctions = initSprigFunctions()
 
-var generatedNamesCount = 0
+var includedNames = make(map[string]int)
+
+const includedMaxRecursion = 1000
```
nils-werner-sonarsource, line 214: Where did this number come from? Just a gut feeling?

petertrr: Same value, as in Helm

nils-werner-sonarsource (approved):
LGTM!

PR #1046, opened 2024-01-04
SONARIAC-1253 Send Go test coverage data to SQ with Gradle

petertrr:


rudy-regazzoni-sonarsource (approved):
LGTM! Couldn't the sonar `{ properties { ... } }` block be put in the `build.gradle.kts` file of the `sonar-helm-for-iac` module?

petertrr:
> LGTM! Couldn't the sonar `{ properties { ... } }` block be put in the `build.gradle.kts` file of the `sonar-helm-for-iac` module?

This was the approach suggested in sonar-gradle-plugin docs. The plugin is only applied to the root project, so it makes sense to keep all the configuration there. I believe the `sonar {}` extension won't even be available in the subprojects' build scripts

PR #1047, opened 2024-01-04
SONARIAC-1254 Files from `templates/` directory should be prefixed with chart name before evaluation

petertrr:


sonar-helm-for-iac/main_test.go
```diff
@@ -502,7 +535,38 @@ data:
...
+kind: ConfigMap
+`)
+
+	result, err := evaluateTemplate(converters.NewTemplateSources("templates/a.yaml", converters.Files{
+		"templates/a.yaml": template,
+		"values.yaml":      values,
+		"Chart.yaml":       DefaultChartYaml,
+		"templates/b.yaml": bYaml}))
+
+	assert.Equal(t, "\napiVersion: v1\nkind: ConfigMap\nannotations:\n  checksum: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855\n", result)
```
petertrr, line 570: This value is actually incorrect. `.Files` still operate on un-prefixed keys, so actual content of `b.yaml` is not retrieved

sonar-helm-for-iac/converters/data-loading.go
```diff
@@ -54,14 +58,25 @@ func PrepareChartValues(templateSources *TemplateSources) (map[string]interface{
...
-		BasePath: getBasePath(chart),
+
+	// Helm allows referencing files (e.g. in `.Files.*` functions or by name in `template` and `include`) with kind of fully-qualified path,
+	// which starts with chart name. See `pkg/chart/loader/load.go:LoadFiles`.
+	// TODO SONARIAC-1241: handle nested charts, i.e. basePath should incorporate parent path like "parent-chart/charts/nested-chart/templates"
+	chartPathPrefix := (*chart)["name"].(string)
+	for filename, content := range templateSources.files {
+		if strings.HasPrefix(filename, "templates/") {
+			delete(templateSources.files, filename)
+			templateSources.files[chartPathPrefix+"/"+filename] = content
```
petertrr, line 72: I don't like that `tesmplateSources` object is being mutated inside a function, but re-organizing everything would increase size of this PR. I'd rather do it in a separate one.

sonar-helm-for-iac/main_test.go
petertrr: `.Files` can't read templates in Helm; for future `include` this should be acceptable

jonas-wielage-sonarsource (approved):
LGTM

PR #1051, opened 2024-01-05
SONARIAC-1255 Remove gradle cache from step build_test_analyze_task to prevent fail on shadowjar size check

rudy-regazzoni-sonarsource:


nils-werner-sonarsource (commented):
Wouldn't it also be fixed, when `GO_CROSS_COMPILE` is set to `1`? What brings more performance?

nils-werner-sonarsource (approved):
LGTM after reading the explaination on Slack.

PR #1052, opened 2024-01-05
SONARIAC-1256 Fix incorrect transfer of files between Java and Go Code

jonas-wielage-sonarsource:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -77,6 +78,8 @@ private FileTree parseHelmFile(String source, @Nullable InputFileContext inputFi
       fileRelativePath = inputFileContext.inputFile.filename();
     } else {
       fileRelativePath = chartRootDirectory.relativize(filePath).normalize().toString();
+      // transform windows to unix path
+      fileRelativePath = fileRelativePath.replace(File.separatorChar, '/');
```
jonas-wielage-sonarsource, line 82: Currently this produces a lot of issues on master for windows, because the Go-Code expects unix separators. Not sure if this is the best way, but was an easy fix.

its/ruling/src/integrationTest/java/org/sonarsource/iac/IacRulingTest.java
```diff
@@ -129,10 +133,6 @@ private void run_ruling_test(String project, Map<String, String> projectProperti
     orchestrator.getServer().provisionProject(projectKey, projectKey);
     LANGUAGES.forEach(lang -> orchestrator.getServer().associateProjectToQualityProfile(projectKey, lang, "rules"));
 
-    File actualDirectory = FileLocation.of("target/actual/" + project).getFile();
```
jonas-wielage-sonarsource, line 132: We shouldn't use target folder anymore...

its/ruling/src/integrationTest/resources/expected/kubernetes/kubernetes-S2260.json
jonas-wielage-sonarsource, line 1: Yeah, 11 parsing issues removed 🥳

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
petertrr: I think it makes sense to handle all this logic on the Java side so that Go can simply operate with Unix separators and LF line breaks.

petertrr (approved):
LGTM! Still not sure what the problem in that code was, but great investigation and good thing that it works now!

PR #1057, opened 2024-01-16
SONARIAC-1266 Should not throw ClassCastException when Helm template contains multiple documents

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/LineNumberCommentInserter.java
```diff
@@ -41,8 +41,11 @@ public static String addLineComments(String content) {
     var lastIndex = 0;
     while (matcher.find()) {
       lineCounter++;
-      sb.append(matcher.group("lineContent"));
-      sb.append(commentLineNumber(lineCounter));
+      var lineContent = matcher.group("lineContent");
+      sb.append(lineContent);
+      if (!"---".equals(lineContent)) {
```
petertrr, line 46: YAML also has `...` for end of document. I've never seen it being used, but maybe we should also add it here?

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/LineNumberCommentInserterTest.java
```diff
@@ -126,6 +126,17 @@ void shouldAddCommentLineNumberOnAlreadyCommentedLine() {
...
   }
 
+  @Test
+  void shouldNotAddLineCommentWhenSeparateDocument() {
+    checkHelmProcessing(
+      code("line1",
+        "---",
+        "line3"),
+      code("line1 #1",
+        "---",
```
petertrr, line 136: Just curious, does snakeyaml generally not support comments after `---` or only at the very end of the file, when the new document consists only of comment? Because I believe that comment after triple dash is allowed in YAML.

petertrr (approved):
LGTM!

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/LineNumberCommentInserterTest.java
mstachniuk: Yes, the comment is allowed after a triple dash or dot. But snakeyaml-engine doesn't support it :-(

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/LineNumberCommentInserter.java
mstachniuk: Good idea, added

PR #1058, opened 2024-01-16
SONARIAC-1178 Support multiline Helm templates

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/LineNumberCommentInserter.java
```diff
@@ -42,25 +43,67 @@ public static String addLineComments(String content) {
...
+  static int getNumberOfUnmatchedDoubleOpeningBraces(String line) {
+    var count = 0;
+    var index = 0;
+    while (index < line.length() - 1) {
+      if ("{{".equals(line.substring(index, index + DOUBLE_BRACE_SIZE))) {
+        count++;
+        index += DOUBLE_BRACE_SIZE;
+      } else if ("}}".equals(line.substring(index, index + DOUBLE_BRACE_SIZE))) {
+        count--;
+        index += DOUBLE_BRACE_SIZE;
```
mstachniuk, line 94: Why not use `String.indexOf("{{")`?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/CommentLocationVisitor.java
```diff
@@ -44,7 +44,8 @@
 public class CommentLocationVisitor extends TreeVisitor<InputFileContext> {
   private static final Logger LOG = LoggerFactory.getLogger(CommentLocationVisitor.class);
 
-  private static final Pattern CONTAINS_LINE_NUMBER = Pattern.compile("#(?<number>\\d+)(\\s#\\d+)*+$");
+  private static final Pattern CONTAINS_LINE_NUMBER = Pattern.compile("#(?<number>\\d++)(\\s#\\d++:?\\d*+)*+$");
+  private static final Pattern CONTAINS_LINE_NUMBER_RANGE = Pattern.compile("#(?<rangeStart>\\d++):(?<rangeEnd>\\d++)(\\s#\\d++:?\\d*+)*+$", Pattern.UNICODE_CHARACTER_CLASS);
```
mstachniuk, line 48: I guess we can use ` ` (simple space) instead of `\\s` as this is in our control and we always use space. Maybe then we can omit `Pattern.UNICODE_CHARACTER_CLASS`?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
```diff
@@ -22,13 +22,16 @@
...
+import java.util.Objects;
+import javax.annotation.Nullable;
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.iac.common.api.tree.impl.TextPointer;
 import org.sonar.iac.common.api.tree.impl.TextRange;
 import org.sonar.iac.common.extension.visitors.InputFileContext;
 
 /**
- * This class is used to store all lines that has to be shifted.
+ * This class is used to store all lines that has to be shifted. "Original" refers to the line in the processed file, i.e. the one that is being analyzed.
```
mstachniuk, line 33: Please add new line before 2nd sentence

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/CommentLocationVisitor.java
```diff
@@ -91,10 +92,16 @@ public void visitComment(InputFileContext ctx, YamlTree tree) {
 
   private void processComment(InputFileContext ctx, Comment comment) {
     var matcher = CONTAINS_LINE_NUMBER.matcher(comment.value());
+    var rangeMatcher = CONTAINS_LINE_NUMBER_RANGE.matcher(comment.value());
```
mstachniuk, line 95: It's not so good to execute 2 similar regexes on the same text. 
Maybe it will be better to only have second Pattern and make `rangeEnd` part optional?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
```diff
@@ -102,7 +113,9 @@ private LineData getOrCreateLinesData(Integer originalLine) {
   }
 
   static class LineData {
-    private Integer targetLine;
+    private Integer targetStartLine;
+    @Nullable
+    private Integer targetEndLine;
```
mstachniuk, line 118: Maybe it's better to store `targetEndLine = targetStartLine` instead of null Integer? Then line 83 can be simplified.

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/LineNumberCommentInserterTest.java
```diff
@@ -155,6 +156,76 @@ void shouldNotAddLineCommentWhenEndDocument() {
...
+  void shouldAddCommentLineRangesWithLoops() {
+    checkHelmProcessing(
+      code(
+        "{{- range $key, $value := .Values.annotations }}",
+        "  {{ $key }}: {{ $value | quote }} {{- end }}{{- end }}"),
+      code(
+        "{{- range $key, $value := .Values.annotations }} #1",
+        "  {{ $key }}: {{ $value | quote }} {{- end }}{{- end }} #2"));
+  }
+
```
mstachniuk, line 219: Could add a case similar to the following:
```
{{ helm
}} {{ helm }} {{ helm
}}
```

its/ruling/src/integrationTest/resources/sources/kubernetes/helm/corner-cases/templates/multiline-templates.yaml
```diff
@@ -0,0 +1,26 @@
...
+    cfg: |-
+      {{- if or
+        (eq "abc" .Values.foo)
+        (eq "def" .Values.bar)
+      }}
+      if
+      {{- else }}
+      else
+      {{- end }}
+spec:
```
mstachniuk, line 23: Please adapt this example to raise some issue to see that location shifting is working.

mstachniuk (commented):
Good progress, please see my clarification questions and suggestions.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/LineNumberCommentInserter.java
```diff
@@ -42,25 +43,67 @@ public static String addLineComments(String content) {
...
 
     var lastIndex = 0;
+    var goTemplateStartLine = -1;
     while (matcher.find()) {
+      // Iterate over all lines. Possible cases:
+      // - Line contains {{ and corresponding }} on the same line: add line number comment
+      // - Line contains regular text and is _not inside_ a go template: add line number comment
+      // - Line contains regular text and is inside a go template: do not add line number comment
+      // - Line contains {{ but no corresponding }} on the same line: do not add line number comment, store line number
+      // - Line contains }} but no corresponding {{ on the same line: add line number comment, reset line number
```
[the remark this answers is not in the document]
petertrr: I think triple dashes can't occur inside templates, so we shouldn't take them into account when adding comments. If there is a `---` followed by a line comment of a range comment, we could remove it, but as a part of a separate ticket.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/LineNumberCommentInserter.java
petertrr: This would mean multiple calls to get all opening braces, then all closing braces. I don't think that code would benefit much from such refactoring.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/CommentLocationVisitor.java
petertrr: I'm not sure what is the logic behind the rule that suggested adding this. I think `\p*` are the only character classes that were present in the rule's description.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/LineNumberCommentInserter.java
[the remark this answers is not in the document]
mstachniuk: This logic is already here implemented, see line 59. Nevermind, I can add it as I will modify this class in my next ticket.

PR #1071, opened 2024-01-22
SONARIAC-1227 S6870: Storage limits should be enforced

navendu-barua-sonarsource:


iac-extensions/kubernetes/src/test/resources/checks/EphemeralStorageLimitCheck/test_pod_object.yaml
```diff
@@ -0,0 +1,77 @@
...
+      volumeMounts:
+        - name: ephemeral
+          mountPath: "/tmp"
+
+    - name: noncompliant_without_limits # Noncompliant
+    # ^^^^
+      image: nginx
+      resources:
+        requests:
+          cpu: 64m
```
jonas-wielage-sonarsource, line 22: Similar to the last PR, you want to test here the `requests` keyword not the `cpu` keyword. 
The base idea of tests should be to change only one thing out of a compliant test case. 
In this case, if you would change the `requests` to `limits`, you would still have a NonCompliant test case because the `ephemeral-storage` is missing.

iac-extensions/kubernetes/src/test/resources/checks/EphemeralStorageLimitCheck/test_template_object.yaml
```diff
@@ -0,0 +1,81 @@
...
+              cpu: 200m
+          volumeMounts:
+            - name: ephemeral
+              mountPath: "/tmp"
+
+        - name: noncompliant_without_limits # Noncompliant
+        # ^^^^
+          resources:
+            requests:
+              memory: 64Mi
```
jonas-wielage-sonarsource, line 31: See other comment

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/EphemeralStorageLimitCheck.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  String getMessage() {
+    return MESSAGE;
+  }
+
+  @Override
+  void registerObjectCheck() {
+    register(KIND_POD, (BlockObject pod) -> pod.blocks("containers")
+      .filter(container -> !container.blocks("volumeMounts").toList().isEmpty())
```
jonas-wielage-sonarsource, line 47: As the target language is set to Java 11, please do not use later Java language features (`toList()`).
Also you don't need to convert this to a list before checking if it's empty.
You can use something like `.anyMatch(blockObject -> true)` or `.findAny().isEmpty()`
As this method is used on both register, you could even extract the content into a new method.

jonas-wielage-sonarsource (approved):
LGTM, please see my suggestions and add them before merging.

iac-extensions/kubernetes/src/test/resources/checks/EphemeralStorageLimitCheck/helm/test_pod_object_helm.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example
+spec:
+  containers:
+    - name: cpu_limit_helm
+      image: nginx
+      resources:
+        limits: #Kubernetes verifier does not support helm files at the moment.
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Please also update this in the other files you created

PR #1073, opened 2024-01-23
SONARIAC-1274 Improve template processing by adding missing Sprig function

petertrr:


sonar-helm-for-iac/helm-imports.go
```diff
@@ -77,15 +86,20 @@ var sprigFunctionsWhitelist = []string{
 	"first",
 	"float64",
 	"floor",
+	"fromJson",
```
jonas-wielage-sonarsource, line 89: Including this would mean a duplicate of the method, as we already manually implemented this.

sonar-helm-for-iac/helm-imports.go
```diff
@@ -180,15 +209,18 @@ var sprigFunctionsWhitelist = []string{
 	"title",
 	"toDate",
 	"toDecimal",
+	"toJson",
```
jonas-wielage-sonarsource, line 212: Like `fromJson`, we should decide on one of the "implementations"

jonas-wielage-sonarsource (approved):
LGTM, please have a look at my comments

sonar-helm-for-iac/helm-imports.go
petertrr: Good catch! It would be overridden by the latest implementation, but I'll still removed it for consistency.

PR #1076, opened 2024-01-23
SONARIAC-1271 Should not fail parsing unquoted text

mstachniuk:


mstachniuk:
Implemented in https://github.com/SonarSource/sonar-iac/pull/1080

PR #1077, opened 2024-01-24
SONARIAC-1205 S6868: Allowing command execution is security sensitive

jonas-wielage-sonarsource:


iac-extensions/kubernetes/src/test/resources/checks/CpuLimitCheck/test_template_object.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         # ^^^^
           resources:
             requests:
-              memory: 64Mi
+              cpu: 64Mi
```
petertrr, line 28: Why is this changed? Is `Mi` a valid quantifier for CPU?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/CommandExecutionCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private static boolean ruleContainsSensitiveVerb(BlockObject rule) {
+    return containsSensitiveItem(rule, "verbs", "create");
+  }
+
+  private static boolean ruleContainsSensitiveResource(BlockObject rule) {
+    return containsSensitiveItem(rule, "resources", "pods/exec");
+  }
+
+  private static boolean containsSensitiveItem(BlockObject rule, String listKey, String sensitiveItem) {
```
petertrr, line 57:
```suggestion
  private static boolean containsSensitiveItemOrWildcard(BlockObject rule, String listKey, String sensitiveItem) {
```

petertrr (approved):
LGTM

iac-extensions/kubernetes/src/test/resources/checks/CpuLimitCheck/test_template_object.yaml
jonas-wielage-sonarsource: Good catch, the `Mi` should be removed. I changed it here, because I remarked this on another PR and it was missed to change.

PR #1083, opened 2024-01-25
SONARIAC-1203 S6867: Wildcards should not be used to define RBAC permissions

jonas-wielage-sonarsource:


iac-extensions/kubernetes/src/test/resources/checks/RBACWildcardCheck/wildcardCheckTestTemplate.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - apiGroups: [""]
+    resources: ["pods/exec"] # Compliant
+    verbs: ["get"]
+
+  - apiGroups: [""]
+    resources: ["services"] # Compliant
+    verbs: ["create"]
+
+  - apiGroups: [""]
+    verbs: ["create"] # Compliant
```
navendu-barua-sonarsource, line 50: not sure about this, should there be an empty line at the end?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/RBACWildcardCheck.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+      }));
+  }
+
+  private static boolean containsWildCardItem(BlockObject rule, String listKey) {
+    Predicate<YamlTree> wildcardPredicate = TreePredicates.isEqualTo("*");
+    return rule.list(listKey).getItemIf(wildcardPredicate).findAny().isPresent();
+  }
+
+  void reportOnKey(BlockObject rule, String key) {
```
navendu-barua-sonarsource, line 59: may be we should move this method to `AttributeObject` class like others that checks and reports. There is also an existing method `report(textRange, message)` that can be also be used from the new method

navendu-barua-sonarsource (approved):
Lgtm, left couple of minor comments.

iac-extensions/kubernetes/src/test/resources/checks/RBACWildcardCheck/wildcardCheckTestTemplate.yaml
jonas-wielage-sonarsource: Not sure what you mean by this. The test is for when `resources` is missing (should not be the case IRL)

navendu-barua-sonarsource: Some coding guidelines states to keep an empty line at the end of every file. Is that something to be followed ?

PR #1085, opened 2024-01-25
SONARIAC-1277 Cleanup duplicated logic when evaluating Helm templates

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -88,6 +89,13 @@ private FileTree evaluateAndParseHelmFile(String source, InputFileContext inputF
     return super.parse(evaluatedAndCleanedSource, inputFileContext, FileTree.Template.HELM);
   }
 
+  private static void readLinesSizes(String source, InputFileContext ctx, LocationShifter locationShifter) {
```
petertrr, line 92: Maybe it can be a member method of `LocationShifter`? It seems to me, we put a lot of pieces of Helm-related logic into `KubernetesParser`

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -78,6 +78,7 @@ private FileTree parseHelmFile(String source, @Nullable InputFileContext inputFi
   }
 
   private FileTree evaluateAndParseHelmFile(String source, InputFileContext inputFileContext) {
+    readLinesSizes(source, inputFileContext, locationShifter);
     var fileRelativePath = getFileRelativePath(inputFileContext);
     var evaluatedSource = helmProcessor.processHelmTemplate(fileRelativePath, source, inputFileContext);
     var evaluatedAndCleanedSource = cleanSource(evaluatedSource, inputFileContext, locationShifter);
```
petertrr, line 84: Actually, not related to this PR per se, but why not call `cleanSource` inside `HelmProcessor#processHelmTemplate`?

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/plugin/KubernetesParserTest.java
```diff
@@ -272,6 +250,175 @@ void shouldRemoveLineNumberCommentForEndDocumentAfterEvaluation() throws IOExcep
...
       "... #10");
+
+    FileTree file = parseTemplate("{{ dummy helm }}", evaluated);
+
+    assertThat(file.documents().get(0).children()).hasSize(4);
+    assertThat(file.template()).isEqualTo(FileTree.Template.HELM);
+  }
+
+  @Test
+  void shouldFindShiftedLocation() throws IOException, URISyntaxException {
```
petertrr, line 261: I believe simply copying these tests from `CommentLocationVisitorTest` here adds a bit of an overhead, because in these tests we only check pre-processing part (marking lines) and shifting, but execute everything including YAML parsing. WDYT about moving this logic e.g into `HelmProcessor` (maybe as a separate ticket)?

petertrr (approved):
LGTM! Please see my questions

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
mstachniuk: Separation of responsibility. The `HelmProcessor` already does a lot and returns the evaluated template directly from Go executable. It's the exact text we can produce using `helm -s templates/pod.yaml .` I would like to have it for easier debugging purposes. 
But I'm open to discussion and maybe introducing something between `KubernetesParser` and `HelmProcessor` is needed. 🤔 And what is a good name for it?

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/plugin/KubernetesParserTest.java
mstachniuk: I agree it's a bit of overhead, but I didn't find a better place for those tests.
The original `CommentLocationVisitorTest` also executed YAML parsing. In both cases, the Go executable is mocked.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
mstachniuk: Maybe `HelmParser` is a good name?

petertrr: I'd rather call it `HelmPostProcessor` and call inside `HelmProcessor` :D Then existing logic can be encapsulated in `HelmPreProcessor`, so that `HelmProcessor` simply calls `HelmPreprocessor`, `HelmEvaluator` and then `HelmPostprocessor`. But it's not a final thought still.
I think `HelmParser` would be a confusing name as it doesn't parse and is in fact part of the `KubernetesParser`

mstachniuk: Ticket created: https://sonarsource.atlassian.net/browse/SONARIAC-1289

PR #1087, opened 2024-01-25
SONARIAC-1226 S5332: Using clear-text protocols is security-sensitive

petertrr:


iac-extensions/kubernetes/src/test/resources/checks/ClearTextProtocolsCheck/clearTextProtocolsTestTemplate.yaml
```diff
@@ -0,0 +1,38 @@
...
+metadata:
+  name: curl
+spec:
+  template:
+    spec:
+      containers:
+        - name: curl
+          image: curlimages/curl
+          command: ["curl"]
+          args: ["http://example.com/"] # Noncompliant {{Make sure that using clear-text protocols is safe here.}}
```
jonas-wielage-sonarsource, line 12: Please include tests that show case insensitivity and also tests where the URL is in another scalar other than args.
Example:
```
  containers:
    - name: applications
      image: my_image_with_ssh
      ports:
        - hostIP: "ftp://example.com/"
```
Doesn't make the most sense in the file itself (maybe you find a better example directly from the already raised its) but it shows, that these can be found at other places and not only in lists like `[" "]`.

jonas-wielage-sonarsource: Also this test could also be nice: `args: ["-v", "http://example.com/"]`

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/ClearTextProtocolsCheckTest.java
```diff
@@ -0,0 +1,40 @@
...
+class ClearTextProtocolsCheckTest {
+
+  IacCheck check = new ClearTextProtocolsCheck();
+
+  // static Stream<String> insecureProtocols() {
+  // return Stream.of("http", "ftp");
+  // }
+
+  // @MethodSource("insecureProtocols")
+  // @ParameterizedTest(name = "[{index}] should detect insecure protocol: \"{0}\"")
```
jonas-wielage-sonarsource, line 34: Commented out code.

jonas-wielage-sonarsource (approved):
LGTM, please have a look at my suggestions about adding tests

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/ClearTextProtocolsCheck.java
```diff
@@ -0,0 +1,58 @@
...
+@Rule(key = "S5332")
+public class ClearTextProtocolsCheck implements IacCheck {
+  private static final String MESSAGE = "Make sure that using clear-text protocols is safe here.";
+  private static final List<String> INSECURE_PROTOCOLS = List.of("http", "ftp");
+  private static final Pattern LOOPBACK_PATTERN = Pattern.compile("localhost|127(?:\\.\\d+){0,2}\\.\\d+$|^(?:0*:)*+:?0*1", Pattern.CASE_INSENSITIVE);
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(ScalarTree.class, ClearTextProtocolsCheck::checkUrlsInScalar);
+  }
```
jonas-wielage-sonarsource, line 41: And btw, nice idea to directly go to the scalar tree here.

PR #1090, opened 2024-01-26
SONARIAC-1288 Log how many pure Kubernetes and Helm files were analyzed by K8s sensor

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -49,18 +49,26 @@ public class KubernetesParser extends YamlParser {
...
     this.locationShifter = locationShifter;
+    this.kubernetesParserStatistics = kubernetesParserStatistics;
   }
 
   @Override
   public FileTree parse(String source, @Nullable InputFileContext inputFileContext) {
     if (!hasHelmContent(source)) {
-      return super.parse(source, inputFileContext);
+      kubernetesParserStatistics.incrementPureKubernetesFileCount();
+      var result = super.parse(source, inputFileContext);
```
petertrr, line 64: Maybe instead methods like `recordPureKubernetesFile` in `KubernetesParserStatistics`:
```java
public <T> T recordPureKubernetesFile(Supplier<T> callable) {
  incrementPureKubernetesFileCount();
  var result = callable();
  incrementPureKubernetesParsedFileCount();
  return result;
}
```
and use it here like
```suggestion
return kubernetesParserStatistics.recordPureKubernetesFile(() -> super.parse(source, inputFileContext))`
```
?

petertrr (approved):
LGTM, just one suggestion

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
mstachniuk: Good point

PR #1098, opened 2024-02-05
Use a more powerful machine for Windows builds

nils-werner-sonarsource:
Reduces pure Windows build time from 8:20 to 4:30. That reduces the CI time from ~15 min to 11 min


petertrr (approved):
LGTM

PR #1110, opened 2024-02-08
Improve ci pipeline

nils-werner-sonarsource:


.cirrus.yml
```diff
@@ -94,41 +95,46 @@ win_vm_definition: &WINDOWS_VM_DEFINITION
...
+# ----------------------------------------------
+# -----------------TEMPLATES--------------------
+# ----------------------------------------------
 
 gradle_its_template: &GRADLE_ITS_TEMPLATE
   <<: *SETUP_GRADLE_CACHE
   env:
     # to be able to reuse cached results of previous tasks
     GO_CROSS_COMPILE: '1'
+    KEEP_ORCHESTRATOR_RUNNING: 'true'
```
petertrr, line 137: How is the orchestrator stopped then? Is it by `OrchestratorExtension#afterAll` or is it killed with the Cirrus container?

petertrr (approved):
LGTM, please see my question

.cirrus.yml
nils-werner-sonarsource: It's killed when the container is killed. I had to put such a condition to allow the local execution without the need to kill the process manually. Not perfect but simple.

PR #1115, opened 2024-02-09
[DO NOT MERGE] Create orchestrator log setup

jonas-wielage-sonarsource:


jonas-wielage-sonarsource:
Problem was resolved some time ago by updating library

PR #1118, opened 2024-02-12
SONARIAC-1328 Add end offset to Go AST nodes

petertrr:


sonar-helm-for-iac/converters/tree_converter.go
```diff
@@ -79,208 +103,234 @@ func convert(node parse.Node) proto.Message {
...
 	return &pbstructs.BreakNode{
 		NodeType: pbstructs.NodeType_NodeBreak,
-		Pos:      int64(node.Pos),
+		Pos:      int64(node.Position()),
+		Length:   int64(len(KwBreak)),
 	}
 }
 
-func convertChainNode(node parse.ChainNode) proto.Message {
-	return &pbstructs.ChainNode{
```
petertrr, line 124: I still can't understand if these nodes can actually occur or are just legacy. Because of that, I can't properly test it, so I decided to remove it.

sonar-helm-for-iac/ast.proto
```diff
@@ -65,13 +66,15 @@ message ActionNode {
   int64 pos = 2;
   // The pipeline in the action.
   optional PipeNode pipe = 3;
+  int64 length = 4;
```
mstachniuk, line 69: It will be good to put `length` next to `pos` here. They have the meaning together and are used in all nodes. As this is in "development" cycle we don't have to care about backward compatibility. Later it can be harder to change.
We should have the same order of fields in Go and Java side.
This apply to next Nodes too.

sonar-helm-for-iac/converters/tree_converter_test.go
```diff
@@ -34,44 +50,124 @@ func Test_convert_simple(t *testing.T) {
...
+`
+	tpl, _ := template.New("test-define-and-template").Parse(code)
+
+	tree := defaultTestConverter.ConvertTree(code, tpl.Tree)
+
+	basicTreeAsserts(t, tree, "test-define-and-template", 21, 29, 3)
+
+	n, _ := anypb.UnmarshalNew(tree.Root.Nodes[0], proto.UnmarshalOptions{})
+	textNode := n.(*pbstructs.TextNode)
+	// Note: comment and `define` nodes are not present in the AST
```
mstachniuk, line 123: Is there an option to include `define` nodes to AST? This will be needed for raising issues in custom templates.

sonar-helm-for-iac/converters/tree_converter_test.go
```diff
@@ -11,9 +11,25 @@ import (
 	"text/template/parse"
 )
 
+type TestConverter struct {
+	DefaultConverter
+}
+
+var defaultTestConverter = &TestConverter{}
```
mstachniuk, line 18: Maybe it can be named just `converter` instead of `defaultTestConverter`?

sonar-helm-for-iac/main.go
```diff
@@ -121,7 +124,8 @@ func evaluateTemplateWithReferences(templateName string, templateContent string,
 		return resultWithError(err)
 	}
 
-	return executePreparedTemplate(tmpl, templateName, values)
+	result := executePreparedTemplate(tmpl, templateName, values)
```
mstachniuk, line 127: this can be inline

mstachniuk (changes requested):
Good direction, please see my comments

sonar-helm-for-iac/converters/tree_converter_test.go
petertrr: Not easily. It is parsed and added into `Tree.treeSet`, so that it can be later used during execution. It can be extracted from there and placed into the tree (however, there is no node type for `define`). But maybe it's another point in favor of implementing a our own parser.

sonar-helm-for-iac/converters/tree_converter.go
petertrr: Nope, it appears in things like `(.Values).foo`. Added a test case and restored code

mstachniuk (approved):
LGTM

PR #1120, opened 2024-02-14
SONARIAC-1158 Resolve simple value location

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -0,0 +1,61 @@
...
+      .map(FieldNode::identifiers)
+      .collect(Collectors.toList());
+  }
+
+  private static Predicate<Node> hasOverlayingLocation(Integer position) {
+    // TODO adjust after SONARIAC-1328
+    return node -> node.position() > position;
+  }
+
+  public static void addChildrenIfPresent(List<Node> children, @Nullable Node tree) {
```
petertrr, line 56: WDYT about moving this method into `AbstractNode` and constructing `children` as a class field in constructor?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -0,0 +1,61 @@
...
+  private GoTemplateAstHelper() {
+    // utility class
+  }
+
+  public static List<List<String>> findNodes(GoTemplateTree tree, TextRange range, String text) {
+    var positionAndLength = TextRanges.toPositionAndLength(range, text);
+    var position = positionAndLength.first();
+    return tree.root().children().stream()
+      .filter(hasOverlayingLocation(position))
+      .map(Node::children)
```
petertrr, line 43: It should go deeper than just second level of children? Maybe add a TODO comment here?

sonar-iac-plugin/src/test/java/org/sonar/iac/ArchUnitTest.java
```diff
@@ -61,4 +65,37 @@ public void check(JavaClass javaClass, ConditionEvents events) {
    */
   @ArchTest
   static final ArchRule shouldNotCallPathToRealPath = noClasses().should().callMethod(Path.class, "toRealPath", LinkOption[].class);
+
+  @ArchTest
+  static final ArchRule shouldOverrideChildrenMethodIfContainsNodes = classes().that().implement(Node.class)
```
petertrr, line 70: Nice! I also thought about adding an arch unit test to enforce `fromPb` method on all nodes, but didn't do it

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -0,0 +1,61 @@
...
+import org.sonar.iac.helm.tree.api.GoTemplateTree;
+import org.sonar.iac.helm.tree.api.Node;
+
+public class GoTemplateAstHelper {
+
+  private GoTemplateAstHelper() {
+    // utility class
+  }
+
+  public static List<List<String>> findNodes(GoTemplateTree tree, TextRange range, String text) {
```
petertrr, line 38: I could suggest something like `ValueIdentifiers` instead of `ListString`, to resemble `FieldNode.identifiers`, where we get these names from. OTOH, `ValuesPath` is similar to `YamlPath`, so I'm also fine with that.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
mstachniuk: This will impact memory consumption - most nodes must allocate space for the list. This list is unnecessary in each case, so it's better to create it lazily. In the same way, it's solved in Docker AST.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
mstachniuk: It's need to be solved now

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
mstachniuk: Let's agree on `ValuePath`

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -0,0 +1,79 @@
...
+import org.sonar.iac.helm.tree.api.Node;
+import org.sonar.iac.helm.tree.impl.LocationImpl;
+
+public class GoTemplateAstHelper {
+
+  private GoTemplateAstHelper() {
+    // utility class
+  }
+
+  public static List<ValuePath> findNodes(GoTemplateTree tree, TextRange range, String text) {
```
petertrr, line 40: Maybe rename `text` to `sourceText` or `inputFileContent`?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static Predicate<Node> hasOverlayingLocation(Location location) {
+    return node -> {
+      var position = location.position();
+      var length = location.length();
+      var nodePosition = node.location().position();
+      var nodeLength = node.location().length();
+      return (nodePosition >= position && nodePosition <= position + length) ||
+        (nodePosition + nodeLength >= position && nodePosition + nodeLength <= position + length) ||
+        (nodePosition < position && nodePosition + nodeLength > position + length);
```
petertrr, line 62: Can this be simplified as `!(nodePosition > position + length || nodePositon + nodeLength < position)`?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -0,0 +1,79 @@
...
+    };
+  }
+
+  private static List<Node> allChildren(List<Node> nodes) {
+    List<Node> allNodes = new ArrayList<>(nodes);
+    for (int i = 0; i < allNodes.size(); i++) {
+      allNodes.addAll(allNodes.get(i).children());
```
petertrr, line 69: Shouldn't it be `allNodes.addAll(allNodes.get(i).children()); allNodes.addAll(allChildren(allNodes.get(i).children()))`?

mstachniuk: No, this will produce duplicates on the list

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/api/Node.java
```diff
@@ -36,4 +38,11 @@ public interface Node {
    * @return the location of the node in the file
    */
   Location location();
```
petertrr, line 40: I just thought that maybe we should've converted `Location`s to `TextRange`s first, and then operate only with them. But maybe it's something to consider when reworking the Go AST

petertrr (approved):
LGTM!

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/api/Node.java
mstachniuk: Yes, let's think about it later.

PR #1123, opened 2024-02-14
SONARIAC-1334 Implement ability to raise secondary locations on other files

jonas-wielage-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/extension/visitors/InputFileContext.java
```diff
@@ -56,18 +58,19 @@ public void reportIssue(RuleKey ruleKey, @Nullable TextRange textRange, String m
...
 
       if (textRange != null && TextRanges.isValidAndNotEmpty(textRange)) {
-        issueLocation.at(toInputFileRange(textRange));
+        issueLocation.at(toInputFileRange(inputFile, textRange));
       }
 
       issue.forRule(ruleKey).at(issueLocation);
 
       secondaryLocations.stream()
         .filter(location -> location != null && TextRanges.isValidAndNotEmpty(location.textRange))
```
petertrr, line 67: Nitpick: both lambdas operate with same objects, but arguments are called differently (`location` and `secondary`)

iac-common/src/main/java/org/sonar/iac/common/api/checks/SecondaryLocation.java
```diff
@@ -20,31 +20,35 @@
...
-import org.sonar.iac.common.api.tree.impl.TextRanges;
 
 public class SecondaryLocation {
 
   public final TextRange textRange;
 
   public final String message;
 
+  @Nullable
+  public final String filePath;
```
petertrr, line 34: Maybe add a description, that `filePath` doesn't have to be strictly absolute or relative, but has to satisfy `FilePredicates.is`, and that if it's `null`, then the issue will be raised on the same file?

petertrr (approved):
LGTM, just a couple of minor suggestions

iac-common/src/test/java/org/sonar/iac/common/extension/visitors/InputFileContextTest.java
```diff
@@ -88,10 +90,45 @@ void secondary_location_is_not_added_when_its_range_is_invalid_or_empty() {
...
+    assertThat(issues).hasSize(1);
+    assertThat(issues.get(0).flows().get(0).locations()).hasSize(1);
+
+    IssueLocation secondaryIssueLocation = issues.get(0).flows().get(0).locations().get(0);
+    assertThat(secondaryIssueLocation.inputComponent()).isEqualTo(secondaryFile);
+    assertThat(secondaryIssueLocation.message()).isEqualTo("messageSecondary");
+    assertThat(secondaryIssueLocation.textRange().start().line()).isEqualTo(VALID_RANGE.start().line());
+    assertThat(secondaryIssueLocation.textRange().start().lineOffset()).isEqualTo(VALID_RANGE.start().lineOffset());
+    assertThat(secondaryIssueLocation.textRange().end().line()).isEqualTo(VALID_RANGE.end().line());
+    assertThat(secondaryIssueLocation.textRange().end().lineOffset()).isEqualTo(VALID_RANGE.end().lineOffset());
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Nice, that would make it cleaner. Used it this way because we have two different kinds of textRanges.

PR #1124, opened 2024-02-14
SONARIAC-1172 Add code formatter for Go code as part of the build

petertrr:


build-logic/src/main/kotlin/org.sonarsource.iac.java-conventions.gradle.kts
```diff
@@ -25,8 +27,7 @@ tasks.withType<Test> {
     useJUnitPlatform()
 
     testLogging {
-        exceptionFormat =
-            org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL // log the full stack trace (default is the 1st line of the stack trace)
+        exceptionFormat = TestExceptionFormat.FULL // log the full stack trace (default is the 1st line of the stack trace)
```
petertrr, line 30: The new version of spotless detects this as a line longer than allowed

sonar-helm-for-iac/build.gradle.kts
```diff
@@ -77,6 +78,19 @@ if (isCi) {
...
+        go {
+            val goVersion = providers.environmentVariable("GO_VERSION").getOrElse("1.21.1")
+            gofmt("go$goVersion")
+            target("**/*.go")
+            targetExclude("**/*.pb.go")
+        }
+    }
+    tasks.named("check") {
+        dependsOn("spotlessCheck")
+    }
```
jonas-wielage-sonarsource, line 93: I would suggest adding something like this (maybe changing the message). This way we could direct the user to a tool which he should run locally.

```suggestion
    // spotless is enabled only for CI, because spotless relies on Go installation being available on the machine and not in a container
    spotless {
        go {
            val goVersion = providers.environmentVariable("GO_VERSION").getOrElse("1.21.1")
            gofmt("go$goVersion")
            target("**/*.go")
            targetExclude("**/*.pb.go")
        }
    }
    tasks.named("check") {
        dependsOn("spotlessCheck")
    }
    tasks.named("spotlessGoCheck") {
        finalizedBy("spotlessGoCheckFailure")
    }

    tasks.register("spotlessGoCheckFailure") {
        doLast {
            println("Go code is not formatted correctly. Please format your code locally, for example with the IntelliJ Go plugin, which contains \"gofmt\".")
        }
        onlyIf {
            tasks.getByPath("spotlessGoCheck").state.failure != null
        }
    }
```

jonas-wielage-sonarsource: Or if you don't like the solution, add a bit more commentary, that gofmt is integrated into the IntelliJ Go plugin

jonas-wielage-sonarsource (approved):

PR #1125, opened 2024-02-15
Fix ITs configuration: fix `keepSonarqubeRunning` property

petertrr:


its/ruling/src/integrationTest/java/org/sonarsource/iac/IacRulingTest.java
```diff
@@ -163,10 +162,10 @@ private void run_ruling_test(String project, Map<String, String> projectProperti
   }
 
   @AfterAll
-  public static void after() {
+  public static void after() throws InterruptedException {
     if (keepSonarqubeRunning) {
       // keep server running, use CTRL-C to stop it
-      new Scanner(System.in).next();
```
petertrr, line 169: I couldn't find a way to allow a test run with Gradle to read `System.in`; it leads to exception otherwise.

gradle.properties
```diff
@@ -7,3 +7,6 @@ org.gradle.parallel=false
 org.gradle.caching=true
 
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError
+
+# Disable deprecated behavior that will be removed in sonar plugin 5.x
+systemProp.sonar.skipCompile=true
```
petertrr, line 12: Removes a warning from logs

its/ruling/src/integrationTest/java/org/sonarsource/iac/IacRulingTest.java
nils-werner-sonarsource: This property is quite old. I wonder why it is affecting you. I introduced it to keep the Orchestrator running when testing locally.

build-logic/src/main/kotlin/org.sonarsource.iac.integration-test.gradle.kts
```diff
@@ -16,6 +16,9 @@ val integrationTestTask =
     tasks.register<Test>("integrationTest") {
         description = "Runs integration tests."
         group = "verification"
+        inputs.dir("$rootDir/its/sources")
+        inputs.property("SQ version", System.getProperty("sonar.runtimeVersion", "LATEST_RELEASE"))
```
nils-werner-sonarsource, line 20: Funny, I was trying to get the property also for the Plugin ITs, but failed so I used env instead. So I learned something new. 😆

nils-werner-sonarsource (approved):
LGTM!

PR #1128, opened 2024-02-16
SONARIAC-1331 Extend Verifier to be able to verify secondary locations on secondary files

jonas-wielage-sonarsource:


iac-common/src/testFixtures/java/org/sonar/iac/common/testing/Verifier.java
```diff
@@ -244,22 +250,44 @@ public void reportIssue(HasTextRange toHighlight, String message, List<Secondary
...
+
+    private static void addSecondaryOnDifferentFile(MultiFileVerifier.Issue reportedIssue, SecondaryLocation secondary) {
+      reportedIssue.addSecondary(
+        Path.of(secondary.filePath),
+        secondary.textRange.start().line(),
+        secondary.textRange.start().lineOffset() + 1,
+        secondary.textRange.end().line(),
+        secondary.textRange.end().lineOffset(),
+        secondary.message);
+    }
```
jonas-wielage-sonarsource, line 290: This approach is necessary / chossen because of the way the `InternalIssueVerifier` API is implemented.
Basically `InternalIssueVerifier` implements `MultiFileVerifier` and `SingleFileVerifie`r, so it supports both.
It uses `InternalIssue` that implements `MultiFileVerifier.Issue` and `SingleFileVerifier.Issue`, but we can't use InternalIssue because it's not public...

jonas-wielage-sonarsource: In theory, we could still be using the `SingleFileVerifier` instead of the `InternalIssueVerifier`, as it's basically the same.

iac-common/src/test/java/org/sonar/iac/common/checks/ToDoCommentCheckTest.java
```diff
@@ -19,57 +19,41 @@
...
 
   @Test
   void shouldReportIssueForToDoComment() {
     TextTree tree = createTreeWithComment("TODO fix me");
-    initContext();
-
-    context.scan(tree);
 
-    verify(verifier).reportIssue("Complete the task associated to this \"TODO\" comment.");
-    verify(issueBuilder).onRange(1, 1, 1, 13);
```
petertrr, line 50: That was a weird test...

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/KubernetesVerifier.java
```diff
@@ -90,32 +91,66 @@ public static void verify(String templateFileName, IacCheck check) {
...
+      InputFileContext inputFileContext = prepareHelmContext(templateFileName);
+      InternalIssueVerifier verifier = verifyHelm(check, inputFileContext, content);
+      verifier.assertNoIssues();
+    } else {
+      verifyNoIssue(PARSER, BASE_DIR.resolve(templateFileName), check);
+    }
   }
 
-  public static void verifyHelmFile(IacCheck check, InputFileContext inputFileContext, String content) {
+  protected static InternalIssueVerifier verifyHelm(IacCheck check, InputFileContext inputFileContext, String content) {
```
petertrr, line 131: I suggest to rename this method, because now it has the same method as `verifyHelm` that actually compares issues. This one can be `prepareHelmVerifier` or something like that

petertrr (approved):
LGTM

iac-common/src/testFixtures/java/org/sonar/iac/common/testing/Verifier.java
```diff
@@ -244,22 +250,44 @@ public void reportIssue(HasTextRange toHighlight, String message, List<Secondary
...
 
     protected void reportIssue(TextRange textRange, String message, List<SecondaryLocation> secondaryLocations) {
-      Issue issue = new Issue(textRange, message, secondaryLocations);
+      var issue = new Issue(textRange, message, secondaryLocations);
       if (!raisedIssues.contains(issue)) {
         TextPointer start = textRange.start();
         TextPointer end = textRange.end();
-        SingleFileVerifier.Issue reportedIssue = verifier
+
+        var reportedIssue = ((SingleFileVerifier) verifier)
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Yes, because otherwise we would get `InternalIssue` back, which we can't use.
This way we get `SingleFileVerifier.IssueBuilder` back, which we are allowed to use.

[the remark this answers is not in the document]
jonas-wielage-sonarsource: I decided to change InternalIssueVerifier to MultiFileVerifier, as it's more readable and understandable.
Also I decided to add some commentary to explain this.

PR #1129, opened 2024-02-19
SONARIAC-1343 Publish values file to SQ

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -70,9 +70,9 @@ public FileTree parse(String source, @Nullable InputFileContext inputFileContext
   }
 
   private FileTree parseHelmFile(String source, @Nullable HelmInputFileContext inputFileContext) {
-    if (inputFileContext == null) {
-      LOG.debug("No InputFileContext provided, skipping processing of Helm file");
-      return super.parse("{}", null, FileTree.Template.HELM);
+    var fileTree = validateInputFileContext(inputFileContext);
```
petertrr, line 73: Maybe call it `fallbackFileTree` or `defaultFileTree`?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -102,6 +102,21 @@ private FileTree parseHelmFile(String source, @Nullable HelmInputFileContext inp
...
   }
 
+  private Optional<FileTree> validateInputFileContext(@Nullable HelmInputFileContext inputFileContext) {
+    if (inputFileContext == null) {
+      LOG.debug("No InputFileContext provided, skipping processing of Helm file");
+      return Optional.ofNullable(super.parse("{}", null, FileTree.Template.HELM));
+    }
+
+    if (("values.yaml".equals(inputFileContext.inputFile.filename()) || "values.yml".equals(inputFileContext.inputFile.filename())) &&
+      new KubernetesSensor.HelmProjectMemberPredicate(inputFileContext.sensorContext).apply(inputFileContext.inputFile)) {
```
petertrr, line 112: Do we really need this check here? We only get into this method if a file satisfies Sensor's predicates and has Helm content. I'd rather suggest to check that this file is in the root of chart directory, not to skip things like `templates/values.yaml`

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/plugin/KubernetesParserTest.java
```diff
@@ -512,4 +513,31 @@ void shouldRethrowParseExceptionsWithDifferentDetails(@Nullable String details)
...
   }
+
+  @Test
+  void shouldParseValuesYamlFileAsSimpleKubernetesFile() {
+    when(inputFile.toString()).thenReturn("chart/values.yaml");
+    when(inputFile.filename()).thenReturn("values.yaml");
+
+    var actual = parser.parse("foo: bar", inputFileContext);
+
+    assertThat(actual.template()).isEqualTo(FileTree.Template.NONE);
```
petertrr, line 524: ~I don't follow; shouldn't `KubernetesParser#validateInputFileContext` return an empty tree marked as Helm template for this case?~
OK, it's clear after I looked at the second test. Then I suggest to rename this one to `shouldParseValuesYamlFileWithoutHelmContentAsSimpleKubernetesFile`

petertrr (approved):
LGTM, please see some small suggestions

PR #1133, opened 2024-02-20
Increase scanner memory for plugin ITs

nils-werner-sonarsource:


mstachniuk (approved):
Looks good. There is still a buffer looking at the memory usage of plugin and ruling tests. Maybe 4GB will be even better? 🤔

PR #1134, opened 2024-02-20
Improve verifier messages for secondary locations; enable Jacoco html reports locally

petertrr:


build-logic/src/main/kotlin/org.sonarsource.iac.java-conventions.gradle.kts
```diff
@@ -41,7 +41,7 @@ tasks.jacocoTestReport {
     reports {
         xml.required.set(true)
         csv.required.set(false)
-        html.required.set(false)
+        html.required.set(providers.environmentVariable("CI").map(String::toBoolean).orElse(true))
```
jonas-wielage-sonarsource, line 44: So basically locally its always false?

jonas-wielage-sonarsource, line 44: Shouldn't it be false when we're in ci?
`html.required.set(!providers.environmentVariable("CI").isPresent)`

petertrr: Yes, good catch, would be a silly mistake :sweat_smile:

jonas-wielage-sonarsource (approved):
LGTM

PR #1142, opened 2024-02-22
SONARIAC-1319 Should not evaluate templates when Chart.yaml is missing

mstachniuk:


petertrr (approved):
LGTM, but maybe rename the ticket and the PR? E.g. `Should throw parse exception if essential Helm files are missing`

PR #1180, opened 2024-02-28
SONARIAC-1330 Should ignore file when filename contains a new line

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
```diff
@@ -86,6 +87,12 @@ String processHelmTemplate(String path, String source, HelmInputFileContext inpu
   }
 
   static Map<String, String> validateAndReadFiles(HelmInputFileContext inputFileContext) {
+    var lineSeparators = List.of("\n", "\r\n");
```
mstachniuk, line 90: I will add here also `\r`, \u2028, and \u2029 to be on safe side. Some filesystems potentially allows such names: https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
```diff
@@ -86,6 +87,12 @@ String processHelmTemplate(String path, String source, HelmInputFileContext inpu
...
+    if (lineSeparators.stream().anyMatch(sep -> inputFileContext.inputFile.filename().contains(sep) ||
+      inputFileContext.getAdditionalFiles().keySet().stream().anyMatch(f -> f.contains(sep)))) {
```
mstachniuk, line 92: For me if `inputFileContext.inputFile` contains new line we should throw exception (as it is here)
and if Additional File contains new line we should exclude it and try evaluating. Of course there should also debug log.

What if a directory name contains a new line? We should also cover it!

mstachniuk (changes requested):
Please see my comments

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
petertrr: > What if a directory name contains a new line?

We only pass file names to the Go executable, so this should be more or less fine

mstachniuk (approved):
LGTM

PR #1181, opened 2024-02-28
SONARIAC-1357 Raise Kubernetes issues on yaml values instead of key-value pairs

mstachniuk:


iac-common/src/main/java/org/sonar/iac/common/yaml/object/ListObject.java
```diff
@@ -62,13 +65,17 @@ public Stream<YamlTree> getItemIf(Predicate<YamlTree> predicate) {
   }
 
   public ListObject reportIfAnyItem(Predicate<YamlTree> predicate, String message) {
-    getItemIf(predicate).findFirst().ifPresent(item -> report(message));
+    getItemIf(predicate).findFirst().ifPresent(item -> reportOnItems(message));
```
petertrr, line 68: Shouldn't it highlight only on matching items then?

iac-extensions/kubernetes/src/test/resources/checks/CapabilitiesCheck/CapabilitiesChart/templates/capabilities-pod.yaml
```diff
@@ -1,33 +1,4 @@
-apiVersion: v1
```
petertrr, line 1: Why did you remove these tests?

iac-extensions/kubernetes/src/test/resources/checks/CapabilitiesCheck/CapabilitiesChart/templates/capabilities-pod.yaml
```diff
@@ -41,9 +12,8 @@ spec:
       {{- if .Values.capabilities }}
         capabilities:
           add:
-# Noncompliant@+1
+# Noncompliant@+2
           {{- range .Values.capabilities }}
-#                          ^^^^^^^^^^^^^^^^
             - {{ . | quote }}
```
petertrr, line 17:
```suggestion
            - {{ . | quote }}
            #^[sc=0;ec=29]
```
So now we will highlight this part, right?

petertrr (approved):
LGTM. That's a lot of updated tests, good job!

iac-extensions/kubernetes/src/test/resources/checks/CapabilitiesCheck/CapabilitiesChart/templates/capabilities-pod.yaml
mstachniuk: Good catch. Removed because I debugged the last case and forgot to restore it. Now it's resorted.

iac-extensions/kubernetes/src/test/resources/checks/CapabilitiesCheck/CapabilitiesChart/templates/capabilities-pod.yaml
mstachniuk: Right. But I'm not a fan of this syntax and would have a clean loop without comments in between.
Unfortunately, this change introduced regression in such Helm loops, because now the issue is raised on `{{ . | quote }}` what doesn't contain any value path and the trailing comment with line numbers doesn't include the above line.
We have to think about how to solve such cases (detecting loops, looking before, etc.) but it's a topic for the next iteration.

iac-common/src/main/java/org/sonar/iac/common/yaml/object/ListObject.java
mstachniuk: The current logic was that it matches all items, e.g. all Capabilities are sensitive. I guess such a method may be needed in the future.

PR #1182, opened 2024-02-29
Update renovate.json: group dependencies, don't pin GHactions, schedule once per week

petertrr:
…le once per week


nils-werner-sonarsource (approved):
LGTM!

PR #1184, opened 2024-03-01
SONARIAC-1353 Highlight precise simple not-evaluated value in Helm file in primary location

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
```diff
@@ -77,6 +77,31 @@ void addLineSize(InputFileContext ctx, int originalLine, int size) {
...
+   *   <li>if the location from 1st and 2nd step is the same then first line location is taken and line offsets of original TextRange</li>
+   * </ul>
+   */
+  public TextRange shiftLocation(InputFileContext currentCtx, TextRange textRange) {
+    var shiftedToLine = computeShiftedLocation(currentCtx, textRange);
+    var shiftedTextRange = shiftedToLine;
+    if (currentCtx instanceof HelmInputFileContext helmContext) {
+      shiftedTextRange = computeHelmValuePathTextRange(helmContext, shiftedToLine);
+    }
+    if (shiftedTextRange.equals(shiftedToLine)) {
```
petertrr, line 94: I wonder if there can be a case like
```yaml
{{ print "foo: "
-}} bar
```
After evaluation we get
```yaml
foo: bar
#    ^^^
```
but after going back to original offsets we will get
```yaml
{{ print "foo: "
-}} bar
#    ^^^
```
because part of the line is occupied by template and thus will shift.

petertrr (approved):
Nice improvements in precision! Please see my question.

iac-common/src/testFixtures/java/org/sonar/iac/common/testing/TextRangePluginAssert.java
```diff
@@ -0,0 +1,103 @@
...
+import org.assertj.core.api.AbstractAssert;
+import org.assertj.core.api.Assertions;
+import org.assertj.core.api.SoftAssertions;
+import org.sonar.api.batch.fs.TextRange;
+
+/**
+ * This Assert class is for {@link org.sonar.api.batch.fs.TextRange}.
+ * For {@link org.sonar.iac.common.api.tree.impl.TextRange} see {@link TextRangeAssert}.
+ */
+public class TextRangePluginAssert extends AbstractAssert<TextRangePluginAssert, TextRange> {
```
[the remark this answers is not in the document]
mstachniuk: rather `TextRangePluginApiAssert` to have `TextRange` at the beginning - useful by searching for the class

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
mstachniuk: There are a lot of edge cases. Let's verify it later.

PR #1197, opened 2024-03-20
Add `github-notify-promotion` step in `promote` task

petertrr:


.cirrus.yml
```diff
@@ -169,6 +169,7 @@ build_template: &BUILD_TEMPLATE
     - source cirrus-env BUILD
     - source .cirrus/use-gradle-wrapper.sh
     - regular_gradle_build_deploy_analyze ${BUILD_ARGUMENTS}
+    - time source set_gradle_build_version $BUILD_NUMBER
```
petertrr, line 172: If it is executed as a separate script, or in `promote` task directly, it takes about a minute. Here, it takes <10 seconds.

mstachniuk: 1. The `project_version_cache` can be also added to `sca_scan_task` (and a call of `gradlew` in `sca_scan_task` can be removed - please verify)
2. `time` here can be removed

mstachniuk (approved):

PR #1198, opened 2024-03-21
SONARIAC-1378 Transform`KubernetesParser.validateInputFileContext()` to a verifier method

nils-werner-sonarsource:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
```diff
@@ -69,15 +69,19 @@ public FileTree parse(String source, @Nullable InputFileContext inputFileContext
   }
 
   private FileTree parseHelmFile(String source, @Nullable HelmInputFileContext inputFileContext) {
-    var defaultFileTree = validateInputFileContext(inputFileContext);
-    if (defaultFileTree.isPresent()) {
-      return defaultFileTree.get();
+    if (inputFileContext == null) {
+      LOG.debug("No InputFileContext provided, skipping processing of Helm file");
+      return buildEmptyTree();
+    }
```
mstachniuk, line 75: Maybe this can be moved to `isInvalidHelmInputFile`? Then we will mitigate `java:S1142`

mstachniuk (approved):
LGTM

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesParser.java
nils-werner-sonarsource: I thought about the same; unfortunately, we would lose the null check and get warnings about the use of the context at a later point in the method.

PR #1199, opened 2024-03-22
SONARIAC-1340 Decouple Helm evaluation from `KubernetesParser`

nils-werner-sonarsource:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmProcessor.java
```diff
@@ -56,7 +58,7 @@ public static boolean isHelmEvaluatorExecutableAvailable() {
     return OperatingSystemUtils.getCurrentPlatformIfSupported().isPresent();
   }
 
-  private void initialize() {
+  protected void initialize() {
     try {
```
mstachniuk, line 62: Actually, this code block (lines 62-67) should be moved to `KubernetesSensor.initContext()`. Then the method `initialize()` can be removed which will simplify the overall design.

mstachniuk: It can be done in next PR

mstachniuk (approved):

PR #1200, opened 2024-03-22
Disable pinning version of GH actions

petertrr:


.github/workflows/releasability-check.yml
```diff
@@ -13,6 +13,6 @@ jobs:
     permissions:
       id-token: write
       contents: write
-    uses: SonarSource/gh-action_release/.github/workflows/releasability-check.yaml@5.3.2
+    uses: SonarSource/gh-action_release/.github/workflows/releasability-check.yaml@5
```
jonas-wielage-sonarsource, line 16: Should be v5

.github/workflows/dogfood.yml
```diff
@@ -36,14 +36,14 @@ jobs:
...
-        uses: Ilshidur/action-slack@689ad44a9c9092315abd286d0e3a9a74d31ab78a # 2.1.0
+        uses: Ilshidur/action-slack@2
         env:
           SLACK_WEBHOOK: ${{ fromJSON(steps.secrets.outputs.vault).SLACK_WEBHOOK }}
           SLACK_OVERRIDE_MESSAGE: 'Dogfood build for `${{ steps.dogfood.outputs.sha1 }}`: *successful*'
         with:
           args: 'Succeed to build dogfood branch'
       - name: Notify failures on Slack
-        uses: Ilshidur/action-slack@689ad44a9c9092315abd286d0e3a9a74d31ab78a # 2.1.0
+        uses: Ilshidur/action-slack@2
```
jonas-wielage-sonarsource, line 46: should be V2 I believe

.github/workflows/dogfood.yml
```diff
@@ -36,14 +36,14 @@ jobs:
         run: echo "The dogfood branch was `${{ steps.dogfood.outputs.dogfood-branch }}` and its HEAD SHA1 was `${{ steps.dogfood.outputs.sha1 }}`"
       #slack notifications
       - name: Notify success on Slack
-        uses: Ilshidur/action-slack@689ad44a9c9092315abd286d0e3a9a74d31ab78a # 2.1.0
+        uses: Ilshidur/action-slack@2
```
jonas-wielage-sonarsource, line 39: V2

.github/workflows/dogfood.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     steps:
       - name: get secrets
         id: secrets
-        uses: SonarSource/vault-action-wrapper@2.8.0
+        uses: SonarSource/vault-action-wrapper@2
```
jonas-wielage-sonarsource, line 22: V2

jonas-wielage-sonarsource (approved):
Mostly lgtm, please see my comments

.github/workflows/releasability-check.yml
petertrr: Git tag is [not](https://github.com/SonarSource/gh-action_release/releases) prefixed with `v` for this action. I'm not sure if Github can deduce it automatically, but it definitely can match tag without a `v`

.github/workflows/dogfood.yml
petertrr: Same here, this action doesn't prefix tags

PR #1205, opened 2024-03-25
SONARIAC-1355 Enable parsing of comment nodes in the Go template AST

petertrr:
Related diff: https://github.com/SonarSource/go/compare/src/v1.21.8-3...src/v1.21.8.4


mstachniuk (approved):
LGTM
Please also update comments in `GoTemplateTree.mode()` and `CommentNode` in ast.proto

PR #1208, opened 2024-03-26
SONARIAC-1356 Provide precise node lengths in the Go template AST

petertrr:
Relevant diff: https://github.com/SonarSource/go/compare/src/v1.21.8-4...src/v1.21.8-7

### DIFF
--- a/iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
+++ b/iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
@@ -19,14 +19,11 @@
  */
 package org.sonar.iac.kubernetes.visitors;
 
-import java.io.IOException;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.snakeyaml.engine.v2.exceptions.Mark;
 import org.snakeyaml.engine.v2.exceptions.MarkedYamlEngineException;
 import org.sonar.api.batch.fs.InputFile;
@@ -35,8 +32,6 @@ import org.sonar.iac.common.api.tree.impl.TextRange;
 import org.sonar.iac.common.api.tree.impl.TextRanges;
 import org.sonar.iac.helm.ShiftedMarkedYamlEngineException;
 import org.sonar.iac.helm.tree.api.FieldNode;
-import org.sonar.iac.helm.tree.api.Location;
-import org.sonar.iac.helm.tree.impl.LocationImpl;
 import org.sonar.iac.helm.tree.utils.GoTemplateAstHelper;
 
 import static org.sonar.iac.common.yaml.YamlFileUtils.splitLines;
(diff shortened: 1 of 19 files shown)


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/LocationShifter.java
```diff
@@ -168,36 +166,18 @@ public static TextRange computeHelmValuePathTextRange(HelmInputFileContext helmC
     var goTemplateTree = helmContext.getGoTemplateTree();
     var sourceWithComments = helmContext.getSourceWithComments();
     if (goTemplateTree != null && sourceWithComments != null) {
-      try {
-        var contents = helmContext.inputFile.contents();
-        // The go template tree contains locations aligned to source code with additional trailing line numbers comments
-        var textRanges = GoTemplateAstHelper.findValuePathNodes(goTemplateTree, textRange, sourceWithComments)
-          .map(FieldNode::location)
-          .map(LocationShifter::fixLocation)
```
petertrr, line 176: Since locations are now more precise and shouldn't extend outside the text, I removed two workarounds we had: `fixLocation` and `trimEndToText`. Unit tests and ITs are passing, so it doesn't seem to have immediate drawbacks.

its/ruling/src/integrationTest/resources/expected/kubernetes/kubernetes-S6865.json
```diff
@@ -1,429 +1,429 @@
...
-  "kubernetes-project:sources/kubernetes/helm-chart-sonarqube/charts/sonarqube-dce/templates/tests/sonarqube-test.yaml": [
-    12
-  ],
-  "kubernetes-project:sources/kubernetes/helm-chart-sonarqube/charts/sonarqube/templates/sonarqube-sts.yaml": [
-    50
-  ],
-  "kubernetes-project:sources/kubernetes/helm-chart-sonarqube/charts/sonarqube/templates/tests/sonarqube-test.yaml": [
-    12
-  ]
+"kubernetes-project:ruling/src/integrationTest/resources/sources/kubernetes/helm/S5849/templates/deployment.yaml": [
```
mstachniuk, line 2: We should introduce spotless configuration for JSON files too

sonar-helm-for-iac/src/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.3
 	github.com/gobwas/glob v0.2.3
 	github.com/samber/mo v1.11.0
-	github.com/sonarsource/go/src v1.21.8-4
+	github.com/sonarsource/go/src v1.21.8-7
```
mstachniuk, line 10: I wonder why `go.work.sum` is not updated and still contains `github.com/sonarsource/go/src v1.21.8-3 h1.....` 🤔

mstachniuk (approved):
LGTM

sonar-helm-for-iac/src/go.mod
petertrr: We probably don't even need it anymore, since we went with keeping forked dependency in a separate repo. I'll see (as a separate ticket) if it can/should be removed or how to maintain it properly if it's needed.

PR #1214, opened 2024-03-27
SONARIAC-1351 Precise primary location for "include" function in Helm expression

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -38,25 +41,43 @@ private GoTemplateAstHelper() {
...
       .map(FieldNode::identifiers)
       .map(ValuePath::new)
       .toList();
   }
 
+  private static Stream<Node> findIncludeFunctionsFirstArg(GoTemplateTree tree, Location location) {
+    return collectNodesByType(tree.root(), NodeType.NODE_COMMAND, CommandNode.class)
+      .filter(hasOverlayingLocation(location))
+      .filter(cmd -> !cmd.arguments().isEmpty())
+      .filter(cmd -> cmd.arguments().get(0) instanceof IdentifierNode identifierNode && "include".equals(identifierNode.identifier()))
```
nils-werner-sonarsource, line 77: For better readability please move this filter to a dedicated method.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelper.java
```diff
@@ -38,25 +41,43 @@ private GoTemplateAstHelper() {
...
+    return findValuePathNodes(tree, location)
       .map(FieldNode::identifiers)
       .map(ValuePath::new)
       .toList();
   }
 
+  private static Stream<Node> findIncludeFunctionsFirstArg(GoTemplateTree tree, Location location) {
+    return collectNodesByType(tree.root(), NodeType.NODE_COMMAND, CommandNode.class)
+      .filter(hasOverlayingLocation(location))
+      .filter(cmd -> !cmd.arguments().isEmpty())
```
nils-werner-sonarsource, line 76: This is not a good filter when accessing more than one element of the list. `cmd.arguments().size() > 1` should be more safe, even when it is slower. Currently, you assume, that when there is at least one element, there should be the 2nd as well. This may be the case right now, but maybe not in the future.

nils-werner-sonarsource (approved):
LGTM! Please see comments about the filters.

PR #1215, opened 2024-03-27
SONARIAC-1360 Shouldn't throw Exceptions when highlighting issue location

mstachniuk:


its/ruling/src/integrationTest/resources/expected/kubernetes/kubernetes-S6428.json
```diff
@@ -5,6 +5,22 @@
...
+14
+],
+"kubernetes-project:ruling/src/integrationTest/resources/sources/kubernetes/helm/corner-cases/templates/new-line-u2029.yaml": [
+9,
+14
+],
+"kubernetes-project:ruling/src/integrationTest/resources/sources/kubernetes/helm/corner-cases/templates/new-line-windows.yaml": [
+9,
+14
+],
```
nils-werner-sonarsource, line 23: Why do we raise more issues due to the fix?

sonar-helm-for-iac/src/converters/io-handling.go
```diff
@@ -53,14 +54,17 @@ func (s StdinReader) ReadInput(scanner *bufio.Scanner) (string, Files, error) {
 		return "", contents, nil
 	}
 
-	templateName := string(firstLine)
+	templateName := strings.TrimSuffix(string(firstLine), "\n")
 	for !bytes.Equal(firstLine, END_TOKEN) {
-		name := string(firstLine)
+		name := strings.TrimSuffix(string(firstLine), "\n")
+
+		fmt.Fprintf(os.Stderr, "XXXX readed '%s'\n", name)
```
petertrr, line 61: What's the purpose of this log? Was it for debugging only?

sonar-helm-for-iac/src/converters/io-handling.go
```diff
@@ -98,17 +104,41 @@ func (s StdinReader) readInput(scanner *bufio.Scanner, nLines int) ([]byte, erro
...
-	// the last delimiter added needs to be removed again
-	if len(rawInput) > 0 {
-		rawInput = rawInput[:len(rawInput)-1]
-	}
 
 	// if scanner has encountered an error, scanner.Err will return it here
 	return rawInput, scanner.Err()
 }
+
+// By default bufio.Scanner scans input line by line using \n (LR) as separator, trims \r (CR)
```
petertrr, line 117: Isn't it `LF` and not `LR`?

sonar-helm-for-iac/src/converters/io-handling.go
```diff
@@ -98,17 +104,41 @@ func (s StdinReader) readInput(scanner *bufio.Scanner, nLines int) ([]byte, erro
...
+	if atEOF && len(data) == 0 {
+		return 0, nil, nil
+	}
+
+	indexCRLF := bytes.Index(data, []byte("\r\n"))
+	indexNewLine := bytes.IndexAny(data, "\r\n\u2028\u2029")
+
+	if indexCRLF >= 0 && indexCRLF <= indexNewLine {
+		// line ends with CRLR
+		return indexCRLF + 2, data[0 : indexCRLF+2], nil
```
petertrr, line 130: I don't quite understand. If your idea is to keep all separators in the resulting string, why are you only trimming `\n` in the `io-hanling.go`? Or is the idea that since we only append LF on the Java side, all other separators are valid tokens in the input text and because of that should be kept?

sonar-helm-for-iac/src/main.go
```diff
@@ -63,6 +63,7 @@ func main() {
 
 func readAndValidateSources() (*converters.TemplateSources, error) {
 	scanner := bufio.NewScanner(os.Stdin)
+	scanner.Split(converters.ScanLinesIncludeNewLine)
```
petertrr, line 66: Maybe these two lines can be extracted into a function, not to repeat it in all of the tests.

petertrr (commented):
Great finding! Please see my questions.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/HelmEvaluator.java
```diff
@@ -99,13 +99,16 @@ void writeTemplateAndDependencies(Process process, String name, String content,
...
+  // Here \n (instead of %n) needs to be used as it is expected on Go site
+  @SuppressWarnings("java:S3457")
   private static void writeFileToProcess(OutputStream out, String fileName, String content) throws IOException {
-    writeStringAsBytes(out, "%s%n".formatted(fileName));
-    writeStringAsBytes(out, "%d%n".formatted(content.lines().count()));
+    writeStringAsBytes(out, "%s\n".formatted(fileName));
+    writeStringAsBytes(out, "%d\n".formatted(content.lines().count()));
     if (!content.endsWith("\n")) {
       content += "\n";
     }
```
petertrr, line 114: Do we need to do also anything about `writeStringAsBytes(out, content)`? E.g. make sure all lines end with `\n`? Can it make things easier on the Go side?

mstachniuk: I considered replacing all line separators by `\n`, but it was a dirty solution for me because the evaluated output will be not the same as produced by `helm` command. It will also increase the location shifting mechanism complexity.

sonar-helm-for-iac/src/converters/io-handling.go
petertrr: Ok, and `template` can manage to parse the input with any separators, so afterward it works fine? And by keeping the original separators we get correct offsets with respect to the original input?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/HelmEvaluator.java
petertrr: Ok, probaly I caught the idea: https://github.com/SonarSource/sonar-iac/pull/1215#discussion_r1542557046

petertrr: But please add comments in HelmEvaluator and in io-handling about why we only "normalize" file name and lines count and not the content.

petertrr (approved):
LGTM!

sonar-helm-for-iac/src/converters/io-handling.go
mstachniuk: Yes & yes

PR #1231, opened 2024-04-08
SONARIAC-1380 S6954: Elements should not be empty or null

rudy-regazzoni-sonarsource:
I put the check on the following elements:
- direct and indirect properties of resources
- variable
- output (+ for output)

I didn't put the check on parameters.


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/EmptyOrNullValueCheck.java
```diff
@@ -0,0 +1,143 @@
...
+    for (Property property : resource.resourceProperties()) {
+      checkExpression(ctx, property, property.value(), "property");
+    }
+  }
+
+  private static void checkVariable(CheckContext ctx, VariableDeclaration variable) {
+    checkExpression(ctx, variable, variable.value(), "variable");
+  }
+
+  private static void checkOutput(CheckContext ctx, OutputDeclaration output) {
```
mstachniuk, line 95: Please move those 3 methods after `initialize()` for easier top bottom reading

iac-extensions/arm/src/test/resources/checks/EmptyOrNullValueCheckTest/emptyOrNullValue.bicep
```diff
@@ -0,0 +1,112 @@
...
+
+param par1 object = null
+param par2 string = ''
+param par3 object = {}
+param par4 array = []
+param myList array = [
+  'val1'
+  'val2'
+  'val3'
+]
```
mstachniuk, line 86: 1. Those params are between `output`s, please move before or after
2. Maybe we should also raise issues for empty/null parameters? 🤔  The value is optional, the default value needs to be verified for different types. WDYT?

iac-extensions/arm/src/test/resources/checks/EmptyOrNullValueCheckTest/emptyOrNullValue.json
```diff
@@ -0,0 +1,197 @@
...
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
+      "apiVersion": "2023-01-01",
+      "name": "test",
+      "other": null,
+      "kind": "",
+      "tags": {},
+      "properties": {
```
mstachniuk, line 59: The case for empty array in resource object is missing

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/EmptyOrNullValueCheckTest.java
```diff
@@ -0,0 +1,73 @@
...
+      issue(92, 10, 92, 25),
+      issue(93, 10, 93, 25),
+      issue(94, 10, 94, 25)
+    // TODO SONARIAC-1403 ARM Template parser should produce the same AST as Bicep for output with FOR loop
+    // issue(159, 10, 159, 24),
+    // issue(160, 10, 160, 22),
+    // issue(161, 10, 161, 22),
+    // issue(162, 10, 162, 22),
+    // issue(173, 12, 173, 24),
+    // issue(181, 12, 181, 24)
```
mstachniuk, line 65: Those issue locations don't match empty or null values in JSON file.

sonar-helm-for-iac/org.sonar.iac.helm/template-evaluation.pb.go
```diff
@@ -0,0 +1,190 @@
+// SonarQube IaC Plugin
```
mstachniuk, line 1: Why this file is included here? I guess it should be generated in "src/org.sonar.iac.helm/" - maybe there is some configuration issue?

mstachniuk: Or the local `protoc` or `protoc-gen-go` version is not as expected?

mstachniuk (changes requested):
Good direction and please see my comments

iac-extensions/arm/src/test/resources/checks/EmptyOrNullValueCheckTest/emptyOrNullValue.bicep
rudy-regazzoni-sonarsource: 1. 👍 
2. That's a good question., initially I was thinking that it it was not needed for param, but I don't see a valid use case with null/empty value. I will give a try and see if there is a lot of issues raised in ITs.

rudy-regazzoni-sonarsource: I checked. This is not a good idea as there can be a valid use case: an optional parameter with an empty value by default.
If no default value is provided, then providing a value on deployment is mandatory; otherwise, it will fail.

mstachniuk (approved):
LGTM

PR #1232, opened 2024-04-08
SONARIAC-1038 ARM JSON: parse expression in string and build the AST to be same as Bicep equivalent

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/SyntaxTokenImpl.java
```diff
@@ -57,4 +57,8 @@ public List<Tree> children() {
   public Kind getKind() {
     throw new UnsupportedOperationException("No kind for SyntaxToken");
   }
+
+  public void setTextRange(TextRange textRange) {
```
petertrr, line 61: In the current implementation, it is used to shift ranges of nodes of Bicep expression according to the location, where it is present in JSON.

nils-werner-sonarsource: I am not a fan of changing TextRanges afterward. Typically we create a new SyntaxToken to be able to process the original one.

petertrr:
> I am not a fan of changing TextRanges afterward. Typically we create a new SyntaxToken to be able to process the original one.

Yes, I agree. However, we don't have a straightforward way to modify children of an existing node as well. The idea here is that we build a Bicep tree from a string, and the nodes will start at `1:0`. Then we need to shift their start offset to point to the beginning of the expression string. This would mean traversing the Bicep tree and replacing all `SyntaxToken`s with shifted copies , but for that we don't have a method in `Tree`.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/SyntaxTokenImpl.java
petertrr: Yes, I agree. However, we don't have a straightforward way to modify children of an existing node as well. The idea here is that we build a Bicep tree from a string, and the nodes will start at `1:0`. Then we need to shift their start offset to point to the beginning of the expression string. This would mean traversing the Bicep tree and replacing all `SyntaxToken`s with shifted copies , but for that we don't have a method in `Tree`.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmJsonBaseConverter.java
```diff
@@ -185,6 +202,48 @@ public Expression toExpression(YamlTree tree) {
...
+  }
+
+  private Expression toExpressionFromString(ScalarTree scalar) {
+    // Remove enclosing square brackets
+    var expressionString = scalar.value().substring(1, scalar.value().length() - 1);
+    var textRangeStart = scalar.metadata().textRange().start();
+
+    // Replacing line breaks, because the original string in JSON is definitely one-liner, but can contain line breaks which we should
+    // treat as escaped symbols.
+    var expression = (Expression) BICEP_EXPRESSION_PARSER.parse(expressionString.replace("\n", "\\n"));
```
petertrr, line 216: This feels like a workaround but I couldn't come up with anything better

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/SyntaxTokenImpl.java
petertrr: And actually a similar problem arises in the next ticket when I need to substitute `FunctionCall` nodes with `Identifier`s for parameters and variables. Both this and shifting of SyntaxTokens can be solved by e.g. introducing a `replace(Tree old, Tree new)` in the `ArmTree`. However, I'm not sure if it's the best approach.

nils-werner-sonarsource (commented):
I like the idea of utilizing the bicep parser, but I don't like to change the trees. Please have a look at my suggestions to adapt the Bicep parsing.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmJsonBaseConverter.java
```diff
@@ -185,6 +198,48 @@ public Expression toExpression(YamlTree tree) {
...
+      var shiftOffset = shiftStart.lineOffset() + 1;
+      syntaxToken.setTextRange(TextRanges.range(shiftStart.line(),
+        textRange.start().lineOffset() + shiftOffset,
+        shiftStart.line(),
+        textRange.end().lineOffset() + shiftOffset));
+    }
+    for (Tree child : expression.children()) {
+      shiftTextRangeRecursively(child, shiftStart);
+    }
+  }
```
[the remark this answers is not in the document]
petertrr: Addressed in c5ad38d7c23e20c681c65a3e0870379a7bbee03d

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/SyntaxTokenImpl.java
nils-werner-sonarsource: I think we can now get rid of this function.

nils-werner-sonarsource (approved):
LGTM!

PR #1233, opened 2024-04-09
SONARIAC-1400 Logic for Tracking Variable Usage in Azure Resource Manager Templates and Bicep Files

jonas-wielage-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/symbols/Symbol.java
```diff
@@ -45,6 +44,8 @@ public void addUsage(Scope scope, DockerTree tree, Usage.Kind kind) {
     var usage = new Usage(scope, tree, kind);
     usages.add(usage);
     if (tree instanceof HasSymbol treeWithSymbol) {
+      // Each symbol usage only references previous usages
+      // This is done to easily to compute findLastAccessibleAssignedValue in ArgumentResolution
       treeWithSymbol.setSymbol(new Symbol(this));
```
jonas-wielage-sonarsource, line 49: I was confused during my implementation, so I looked why this was added to the Docker Symbol.
I found the discussion in https://github.com/SonarSource/sonar-iac/pull/586#discussion_r1129239346 and added the information as a comment.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private boolean shouldRegisterAsUsage(Identifier identifier) {
+    ArmTree identifierParent = identifier.parent();
+    if (identifierParent == null) {
+      return true;
+    }
+
+    try {
+      identifierParent.getKind();
+    } catch (UnsupportedOperationException e) {
```
jonas-wielage-sonarsource, line 87: I don't like that I need to catch UnsupportedOperationException for `Kinds`, maybe we should change something here...
Happens for example in `InterpolatedStringRightPieceImpl`

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+    if (symbol != null) {
+      symbol.addUsage(symbolTable, identifier, Usage.Kind.ACCESS);
+
+    }
+  }
+
+  private boolean shouldRegisterAsUsage(Identifier identifier) {
+    ArmTree identifierParent = identifier.parent();
+    if (identifierParent == null) {
+      return true;
```
jonas-wielage-sonarsource, line 82: Could also register as false, basically irrelevant as we will never have an `Identifier` at the root...

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
rudy-regazzoni-sonarsource: From what I saw we only have 4 elements without kinds: `SyntaxToken` and the three Interpolated string parts.
Doesn't feel right to have those special use case, maybe we should create a `Kind` for them also.
This can probably be tackled in a separate ticket.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -0,0 +1,90 @@
...
+ * Class used to visit a ArmTree and build symbols and their usages for variables.
+ * Those Symbol/Usage can later be used in checks to resolve Argument(really?) or to report issue variable flow.
+ */
+
+public class ArmSymbolVisitor extends TreeVisitor<InputFileContext> {
+  private static final EnumSet<ArmTree.Kind> PARENT_KIND_INDICATING_NO_USAGE = EnumSet.of(
+    ArmTree.Kind.VARIABLE_DECLARATION,
+    ArmTree.Kind.OUTPUT_DECLARATION,
+    ArmTree.Kind.PROPERTY);
+  private final SymbolTable symbolTable = new SymbolTable();
```
rudy-regazzoni-sonarsource, line 42: From my understanding, this will give all analyzed files the same `SymbolTable` instance.
I don't think this is expected, I would move this new `SymbolTable()` into `visitFile()`.
Also, I would add a test to show that two files analyzed get different `SymbolTable` and that those symbols does not reference each other.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/AbstractArmSymbolVisitorTest.java
rudy-regazzoni-sonarsource, line 1: I understand the purpose of this class, and I believe it's a good thing, as it validates that both the Bicep and JSON ARM templates are validated the same way.
The only issue I have is that the code tested is in a separate file from the assertions/test themselves, which make it harder to understand/read the tests.
Do you think there could be an elegant solution where they could stay together? For example with `@ParameterizedTest` and a `@MethodSource` that would provide the parsed Bicep/Json AST?
This would also benefit us by putting the Bicep and JSON ARM close to each other so that we can easily see and check that they are properly equivalent.

rudy-regazzoni-sonarsource, line 1: The tests are missing assertions about the `symbol()` method of the AST nodes.
In your test, you retrieve symbols only from the `SymbolTable`, but I believe one of the other important features you are introducing is the ability to access symbols from the AST nodes.
Also, I believe an interesting assertion you can add concerns showing that we have different symbol instances between each reference.

rudy-regazzoni-sonarsource (dismissed):
Nice work! A lot of good things, I added comments about a few things that may be improved, let me know if you want to discuss some of them.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -0,0 +1,90 @@
...
+      symbol.addUsage(symbolTable, identifier, Usage.Kind.ACCESS);
+    }
+  }
+
+  private static boolean shouldRegisterAsUsage(Identifier identifier) {
+    var identifierParent = identifier.parent();
+    if (identifierParent == null) {
+      // Identifier without parent are invalid, so should not register as usage
+      return false;
+    }
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Yes, they shouldn't normally appear, but in case of a malformed AST this is mostly to prevent NPE's

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
jonas-wielage-sonarsource: Good catch!

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/AbstractArmSymbolVisitorTest.java
jonas-wielage-sonarsource: I agree, it's not optimal. I previously experimented with `@ParameterizedTest` and `@MethodSource`, but did come to a road block with this approach. I try to give it another go.

jonas-wielage-sonarsource: There are already tests in place, for `Identifier` it's in `SymbolTest.usageInIdentifierShouldBehaveAsExpected` (this was easier than searching for the `Identifier` node in the full AST, as the trees are completely different in Bicep and Json and the Identifier node is a lot of layers down.

jonas-wielage-sonarsource: There are already tests in place, for `Identifier` it's in `SymbolTest.usageInIdentifierShouldBehaveAsExpected` (this was easier than searching for the `Identifier` node in the full AST, as the trees are completely different in Bicep and Json and the Identifier node is a lot of layers down.

jonas-wielage-sonarsource: For `VariableDeclaration` this is asserted in `AbstractArmSymbolVisitorTest.assertSingleSymbolFromVariableDeclaration`
(Yes, it's difficult to find, I try to make the tests better)

jonas-wielage-sonarsource: > Also, I believe an interesting assertion you can add concerns showing that we have different symbol instances between each reference.

We don't have different symbol instances between each reference. Each variable will have the same symbol for every usage. 
In Docker it's as you described. There I added the comments to explain the behavior creating a new Symbol on every Access.

rudy-regazzoni-sonarsource: Thanks for the clarifications!

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/json/IdentifierImpl.java
```diff
@@ -55,4 +58,18 @@ public List<Tree> children() {
...
+    return symbol;
+  }
+
+  @Override
+  public void setSymbol(Symbol symbol) {
+    if (this.symbol != null) {
+      throw new IllegalArgumentException("A symbol is already set");
+    }
+    this.symbol = symbol;
+  }
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: I would like to keep it as is, to have the same approach as in Docker.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
```diff
@@ -0,0 +1,459 @@
...
+        if (!code.isEmpty()) {
+          code += "\n";
+        }
+        code += outputValue;
+      }
+      return code;
+    }
+  }
+
+  private static String buildJsonFile(List<String> variables, String outputValue) {
```
rudy-regazzoni-sonarsource, line 430:
```suggestion
  private static String buildJsonFile(@Nullable List<String> variables, @Nullable String outputValue) {
```

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
```diff
@@ -0,0 +1,459 @@
...
+    } else {
+      return ArmTestUtils.parseBicep(code);
+    }
+  }
+
+  private static String fileWithDefaultVariableDeclaration(String language) {
+    return buildSourceCode(language, List.of(VARIABLE_DECLARATION.get(language)), null);
+  }
+
+  private static String buildSourceCode(String language, List<String> variableDeclarations, String outputValue) {
```
rudy-regazzoni-sonarsource, line 412:
```suggestion
  private static String buildSourceCode(String language, @Nullable List<String> variableDeclarations, @Nullable String outputValue) {
```

rudy-regazzoni-sonarsource (commented):

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
```diff
@@ -0,0 +1,459 @@
...
+    visitor.register(File.class, (ctx, tree) -> visited.add("file_visit"));
+    visitor.register(VariableDeclaration.class, (ctx, tree) -> visited.add("variable_declaration_visit"));
+    visitor.register(Identifier.class, (ctx, tree) -> visited.add("identifier_visit"));
+    visitor.scan(inputFileContext, file);
+
+    assertThat(visited).containsExactly(
+      "file_visit",
+      "variable_declaration_visit",
+      "identifier_visit");
+  }
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: I'm not exactly sure what you mean with this comment.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
```diff
@@ -0,0 +1,459 @@
...
+      sb.append("      \"value\": \"");
+      sb.append(outputValue);
+      sb.append("\",\n");
+      sb.append("    }\n");
+      sb.append("  },");
+    }
+
+    sb.append("\n}");
+    return sb.toString();
+  }
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: That looks nice. Thanks for showing me this, I forgot we have the new syntax available.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
```diff
@@ -0,0 +1,459 @@
...
+    "bicep;VAR;" + "var bar =  '${foo}ConcatToVariable'",
+    // access in function in variableDeclaration should create usage
+    "bicep;VAR;" + "var bar =  '${toLower(foo)}ConcatToVariable'",
+
+    // output
+    "bicep;OUT;" + "output foo string =  foo",
+
+    "json;VAR;" + "\"bar\": \"[variables('foo')]\"",
+    "json;VAR;" + "\"bar\": \"[concat(variables('foo'), '-addToVar')]\"",
+    "json;VAR;" + "\"bar\": \"[concat(toLower(variables('foo')), '-addToVar')]\""
```
mstachniuk, line 129: I guess the string concatenation here is used for better formatting the code and parameters visibility?
Maybe it is better to use `@MethodSource` and JUnit5 Arguments? Then the parameters are separated naturally by language syntax

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
```diff
@@ -0,0 +1,459 @@
...
+      .hasSize(1);
+  }
+
+  @ParameterizedTest
+  @MethodSource("languagesToTest")
+  void shouldOnlyCreateOneAccessUsageWhenRegisteringIdentifierMultipleTimes(String language) {
+    String code = ArmSourceCodeBuilder.create(language)
+      .addVariableDeclaration(VARIABLE_DECLARATION.get(language))
+      .addVariableDeclaration(VARIABLE_DECLARATION_WITH_USAGE.get(language))
+      .build();
```
mstachniuk, line 223: I have mixed feelings about the `ArmSourceCodeBuilder`. From one side it is a nice generic API, from the other side it is a test code and it is harder to imagine what will be the content of `code` after executing these few lines. 
It's ok to leave it as it is but maybe next time it is better to choose simpler approach.

mstachniuk (approved):
LGTM and please see my nitpicked comments

iac-extensions/arm/src/test/java/org/sonar/iac/arm/visitors/ArmSymbolVisitorTest.java
jonas-wielage-sonarsource: Yes I agree. I already started this in my next PR for Parameters, but I will change it here as well!

PR #1234, opened 2024-04-09
SONARIAC-1382 S4507: Delivering code in production with debug features activated is security-sensitive

rudy-regazzoni-sonarsource:


sonar-helm-for-iac/org.sonar.iac.helm/template-evaluation.pb.go
```diff
@@ -0,0 +1,190 @@
+// SonarQube IaC Plugin
```
petertrr, line 1: I believe this file was regenerated in an incorrect directory by accident. It already exists in `src`.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/DebugSettingCheck.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private static Consumer<ContextualResource> checkDebugSetting() {
+    return (ContextualResource resource) -> {
+      if (resource.object("debugSetting").property("detailLevel").is(sensitiveDetailLevel())) {
+        resource.property("debugSetting").report(MESSAGE);
+      }
+    };
+  }
+
+  private static Predicate<Expression> sensitiveDetailLevel() {
```
petertrr, line 47: I think the predicate can be a static class field, as it doesn't depend on the input.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/DebugSettingCheck.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  protected void registerResourceConsumer() {
+    register("Microsoft.Resources/deployments", checkDebugSetting());
+  }
+
+  private static Consumer<ContextualResource> checkDebugSetting() {
+    return (ContextualResource resource) -> {
+      if (resource.object("debugSetting").property("detailLevel").is(sensitiveDetailLevel())) {
+        resource.property("debugSetting").report(MESSAGE);
```
petertrr, line 42: Should it also be `resource.object()`, like in the line above?

petertrr (approved):

sonar-helm-for-iac/org.sonar.iac.helm/template-evaluation.pb.go
rudy-regazzoni-sonarsource: Yeah dunno why It gen regenerated everytime on my machine :sweat_smile:

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/DebugSettingCheck.java
rudy-regazzoni-sonarsource: I did this initially, but it has a slight impact on the JSON ARM code: it highlight only the object `{...}` without the `debugSetting:` part.
Using `property()` instead of `object()` solve this.
(strangely this doesn't happen in Bicep)

PR #1238, opened 2024-04-10
Variable in Bicep

nils-werner-sonarsource:


petertrr:
This can be closed I believe, as it was only an example?

PR #1240, opened 2024-04-11
SONARIAC-1371 S6874: Use a hard-coded value for the apiVersion

petertrr:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/json/ResourceDeclarationImplTest.java
```diff
@@ -90,7 +90,6 @@ void shouldParseResource() {
   @ParameterizedTest
   @ValueSource(strings = {
     "\"type\": 5,                            \"apiVersion\": \"2022-12-29\", \"name\": \"myResource\"",
-    "\"type\": \"Microsoft.Kusto/clusters\", \"apiVersion\": 5,              \"name\": \"myResource\"",
```
petertrr, line 93: This is an invalid value before parsing, from the perspective of JSON. But after parsing and building the tree we can have an arbitrary Expression here

iac-extensions/arm/src/test/resources/checks/HardcodeApiVersionCheck/resources.json
```diff
@@ -0,0 +1,31 @@
...
+    "customApiVersion": {
+      "type": "string"
+    }
+  },
+  "variables": {
+    "customApiVersion": "[first(providers(‘Microsoft.Compute’,’virtualMachines’).apiVersions)]"
+  },
+  "resources": [
+    {
+      "apiVersion": "[parameters('customApiVersion')]",
```
jonas-wielage-sonarsource, line 13: I would add the test case where the apiVersion is `[first(providers(‘Microsoft.Compute’,’virtualMachines’).apiVersions)]`.
This should correctly raise an issue on your side.

iac-extensions/arm/src/test/resources/checks/HardcodeApiVersionCheck/resources.json
```diff
@@ -0,0 +1,31 @@
...
+      "location": "[resourceGroup().location]"
+    },
+    {
+      "apiVersion": "[variables('customApiVersion')]",
+      "type": "Microsoft.Compute/virtualMachines",
+      "name": "nonCompliantResource2",
+      "location": "[resourceGroup().location]"
+    },
+    {
+      "apiVersion": "2023-09-01",
```
jonas-wielage-sonarsource, line 25: Maybe we should add additional test cases to document the behavior?
What if we encounter `[concat('2023','-09-01')]`. ( We should raise, but I believe it's interesting to document in test)

jonas-wielage-sonarsource (approved):
LGTM, please have a look at my suggestions

PR #1244, opened 2024-04-11
SONARIAC-1372 S6949: Don't hardcode resource locations

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/ArmTreeUtils.java
```diff
@@ -112,4 +116,10 @@ public static Predicate<Expression> containsParameterReference(Collection<String
       hasIdentifier.identifier() instanceof Identifier identifier &&
       parameterNames.contains(identifier.value());
   }
+
+  public static Optional<Property> getResourceProperty(ResourceDeclaration resourceDeclaration, String key) {
```
mstachniuk, line 120: Maybe it can be moved to `ResourceDeclaration`? This will be a more Object Oriented design

iac-extensions/arm/src/test/resources/checks/HardcodedResourceLocationRule/resources.bicep
```diff
@@ -0,0 +1,29 @@
...
+param location string = resourceGroup().location
+
+resource compliant1 'Microsoft.Storage/storageAccounts@2022-09-01' = {
+  name: storageAccountName
+  location: location
+  sku: {
+    name: 'Standard_LRS'
+  }
+  kind: 'StorageV2'
+}
```
mstachniuk, line 29: please add a testcase where we are unable to resolve variable/parameter. The same for JSON

mstachniuk (approved):

PR #1247, opened 2024-04-12
SONARIAC-1408 Resource elements order for JSON

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ElementsOrderCheck.java
```diff
@@ -30,6 +30,7 @@ public class ElementsOrderCheck implements IacCheck {
   public void initialize(InitContext init) {
     new ElementsOrderTopLevelJson().initialize(init);
     new ElementsOrderTopLevelBicep().initialize(init);
+    new ElementsOrderResourceJson().initialize(init);
```
jonas-wielage-sonarsource, line 33: I would suggest adding the sub classes to a sub package in `checks`

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ElementsOrderResourceJson.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(ResourceDeclarationImpl.class, ElementsOrderResourceJson::checkResource);
+  }
+
+  private static void checkResource(CheckContext checkContext, ResourceDeclarationImpl resourceDeclaration) {
+    var prevIndex = 0;
+    for (Property property : resourceDeclaration.resourceProperties()) {
+      var index = elementsOrder.getOrDefault(property.key().value(), DEFAULT_ORDER_FOR_UNKNOWN_PROPERTY);
```
jonas-wielage-sonarsource, line 66:
```suggestion
      var index = elementsOrder.getOrDefault(property.key().value().toLowerCase(Locale.ROOT), DEFAULT_ORDER_FOR_UNKNOWN_PROPERTY);
```

jonas-wielage-sonarsource: I believe we're currently having the assumption, that key names are case insensitive. I don't find it mentioned in the ARM docu, so difficult to say. (Maybe worth to start a discussion?)
I only found case sensitive names for symbolic resource names, which we don't support yet.
[Resource Declaration - Use symbolic name](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/resource-declaration#use-symbolic-name)

With the assumption of case sensitive key's my suggestion should be applied in this check and the other already merged checks. See my other comment for an example of breaking test.

iac-extensions/arm/src/test/resources/checks/ElementsOrderCheckTest/resourceExpected.json
```diff
@@ -0,0 +1,25 @@
+{
+  "resources": [
+    {
+      "comments": "if any",
+      "condition": true,
+      "scope": "% parent scope %",
+      "type": "Microsoft.Compute/virtualMachines",
+      "apiVersion": "2023-09-01",
```
jonas-wielage-sonarsource, line 8: With the assumption of case insensitive keys, changing `apiVersion` to `apiversion` would break the test.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ElementsOrderResourceJson.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(ResourceDeclarationImpl.class, ElementsOrderResourceJson::checkResource);
+  }
+
+  private static void checkResource(CheckContext checkContext, ResourceDeclarationImpl resourceDeclaration) {
+    var prevIndex = 0;
+    for (Property property : resourceDeclaration.resourceProperties()) {
```
jonas-wielage-sonarsource, line 65: With the planned changes to my PR, this may not be sufficient, as `name`, `type`, `version` and `resources` will not be contained here anymore... :/

jonas-wielage-sonarsource (approved):
LGTM, please have a look at my comment about case sensitivity,
and my concern about changes in my PR

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/ElementsOrderResourceJson.java
jonas-wielage-sonarsource: Changed in my PR to support your use case.

mstachniuk: After rebase on master with your changes is looks ok

PR #1249, opened 2024-04-15
SONARIAC-1419 JSON filenames containing compile_commands should be excluded

guillem-bartina-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM

PR #1251, opened 2024-04-15
SONARIAC-1374 S1481: Unused local variables should be removed

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AbstractUnusedSymbolCheck.java
```diff
@@ -0,0 +1,73 @@
...
+        if (declarationClass().isInstance(declaration) && isUnused(symbol)) {
+          reportOnDeclaration(checkContext, declaration);
+        }
+      });
+    }
+  }
+
+  private static boolean isUnused(Symbol symbol) {
+    return symbol.usages().stream().noneMatch(usage -> Usage.Kind.ACCESS == usage.kind());
+  }
```
mstachniuk, line 57: Please move the method to `Symbol` class

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/UnusedVariablesCheckTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Test
+  void shouldRaiseOnUnusedVariablesBicep() {
+    BicepVerifier.verify("UnusedVariablesCheckTest/unusedVariable.bicep", CHECK);
+  }
+
+  @Test
+  void shouldNotRaiseAnythingWhenNoAssignmentDetected() {
+    BicepParser parser = BicepParser.create();
+    Path path = Paths.get("src", "test", "resources", "checks").resolve("UnusedVariablesCheckTest/unusedVariable.bicep");
```
mstachniuk, line 66: I don't like inconsistency between how the path to the file is passed here and in other checks.
I would suggest creating a small method in `BicepVerifier` (or a separate class `BicepSymbolVerifier`) to get the symbol table and hide such details on how to create the symbol table.

iac-extensions/arm/src/test/resources/checks/UnusedVariablesCheckTest/containsUnresolvableVariable.json
```diff
@@ -0,0 +1,29 @@
+{
```
mstachniuk, line 1: Is is possible to provide an equivalent for Bicep (with an unresolvable variable)?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AbstractUnusedSymbolCheck.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @CheckForNull
+  private static Declaration retrieveDeclaration(Symbol symbol) {
```
mstachniuk, line 60: Please move the method to `Symbol` class. Maybe then `findAssignmentDeclaration` will be a better name?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AbstractUnusedSymbolCheck.java
```diff
@@ -0,0 +1,73 @@
...
+
+abstract class AbstractUnusedSymbolCheck implements IacCheck {
+  private static final String MESSAGE = "Remove the unused %s \"%s\".";
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(File.class, this::checkSymbol);
+  }
+
+  abstract Class<? extends Declaration> declarationClass();
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: I didn't do it, because it would require an additional null check, I will try and see if I can still cover branches

iac-extensions/arm/src/test/resources/checks/UnusedVariablesCheckTest/containsUnresolvableVariable.json
jonas-wielage-sonarsource: No, it can only happen in JSON

PR #1253, opened 2024-04-16
SONARIAC-882 ARM JSON Support Template expressions

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmJsonBaseConverter.java
```diff
@@ -254,14 +254,17 @@ public Expression toLiteralExpression(ScalarTree tree) {
   }
 
   protected List<Property> toProperties(Tree tree) {
+    List<Property> properties = new ArrayList<>();
+    if (tree instanceof ScalarTree) {
+      return properties;
+    }
```
jonas-wielage-sonarsource, line 260: Nitpick: I would suggest to return `Collections.emptyList()`

jonas-wielage-sonarsource (approved):
LGTM, please look at my suggestion.

PR #1254, opened 2024-04-17
S1481: variable should be detected when used in assigned identity

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -96,12 +106,71 @@ void visitAccessUsage(HasIdentifier tree) {
...
+      return;
+    }
+    String variableName = matchesAssignedIdentitiesPattern(identifier.value());
+    if (variableName != null) {
+      addAccessUsageIfSymbolExists(identifier, variableName);
+    }
+  }
+
+  /**
+   * Checks whether the identifier is contained in the following tree structure:
```
mstachniuk, line 133: The comment explains WHAT the method is doing (what is ok in this case), but I need to understand WHY.
It looks like some special case.
Please clarify here in a comment (or in `visitIdentifierJson()`) why we should not count it as usage. 
Also, the ticket & RSPEC say nothing about this case, please update at least the ticket.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -45,6 +52,8 @@
  * Those Symbol/Usage can later be used in checks to  report issues in the variable flow.
  */
 public class ArmSymbolVisitor extends TreeVisitor<InputFileContext> {
+  private static final Pattern ASSIGNED_IDENTITIES_PATTERN = Pattern.compile("(?U)^\\[resourceID\\('Microsoft\\" +
+    ".ManagedIdentity/userAssignedIdentities/',variables\\('(?<variableName>[a-zA-Z_]\\w*)'\\)\\)]$");
```
mstachniuk, line 56: Nice regex, and:
1. It doesn't work as expected if the Identifier contains a space before `variables(...)` (or in other valid places)
2. Is it not better to try parse it using `ArmTemplateExpressionParser` and validate the tree? Or relax Pattern to `variables\\('(?<variableName>[a-zA-Z_]\\w*)'\\) ` and simplify the logic in `containedInUserAssignedIdentitiesTreeStructure`? I guess it can be more such cases with different resource IDs or other expressions.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
mstachniuk: Ok it is FP you found today, but still the WHY should be documented

mstachniuk (commented):
Please see my clarification questions

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
jonas-wielage-sonarsource: I will add a comment. Yes it's the FP, I don't think we need to update the rspec for it.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
jonas-wielage-sonarsource: I will take the second approach as it doesn't add even more complexity. Also it would maybe catch some more FP's that we don't catch now.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/visitors/ArmSymbolVisitor.java
```diff
@@ -45,6 +49,7 @@
  * Those Symbol/Usage can later be used in checks to  report issues in the variable flow.
  */
 public class ArmSymbolVisitor extends TreeVisitor<InputFileContext> {
+  private static final Pattern ASSIGNED_IDENTITIES_PATTERN = Pattern.compile("(?U)variables\\('(?<variableName>[a-zA-Z_]\\w*)'\\)");
```
mstachniuk, line 52: Nitpick. The `(?U)` doesn't work with [a-zA-Z], maybe it should be replaced by:

```suggestion
  private static final Pattern ASSIGNED_IDENTITIES_PATTERN = Pattern.compile("(?U)variables\\('(?<variableName>[\p{L}_]\\w*)'\\)");
```

if you like to fully support Unicode

mstachniuk (approved):
LGTM

PR #1255, opened 2024-04-17
SONARIAC-1376 S6953: Don't use "allowedValues" for a location parameter

jonas-wielage-sonarsource:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/LocationParameterAllowedValuesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+        .map(Symbol::findAssignmentDeclaration)
+        .filter(declaration -> declaration.is(ArmTree.Kind.PARAMETER_DECLARATION))
+        .map(ParameterDeclaration.class::cast)
+        .filter(parameterDeclaration -> !parameterDeclaration.allowedValues().isEmpty())
+        .ifPresent(declaration -> reportOnDeclaration(checkContext, declaration));
+    }
+  }
+
+  private void reportOnDeclaration(CheckContext checkContext, ParameterDeclaration declaration) {
+    ArmTree treeToRaiseOn = declaration.allowedValues().get(0).parent();
```
jonas-wielage-sonarsource, line 67: We only report when we know that `allowedValues` is not empty, so this won't crash

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/LocationParameterAllowedValuesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+    boolean isBicep = declaration instanceof ParameterDeclarationImpl;
+    if (isBicep) {
```
nils-werner-sonarsource, line 69: Something we should improve is to add the info to the tree or context if it's in a Bicep or ARM Template file.

jonas-wielage-sonarsource: Yes, would be great to have it: https://sonarsource.atlassian.net/browse/SONARIAC-1428

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/LocationParameterAllowedValuesCheck.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public void initialize(InitContext init) {
+    init.register(ResourceDeclaration.class, this::checkResourceLocation);
+  }
+
+  void checkResourceLocation(CheckContext checkContext, ResourceDeclaration resourceDeclaration) {
+    resourceDeclaration.resourceProperties().stream()
+      .filter(property -> "location".equals(property.key().value().toLowerCase(Locale.ROOT)))
+      .map(Property::value)
+      .forEach(expression -> reportIfParameterWithAllowedValuesIsUsed(checkContext, expression));
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: I tried to do it, but it doesn't work, as the `ContextualResource` only provides `properties` from the `"property"` key. We need to query `resourceProperties()` here, which our current `ContextualResource` doesn't support.

PR #1256, opened 2024-04-17
SONARIAC-1405 Fix Syntax highlighting for ARM IllegalArgumentException

mstachniuk:


iac-common/src/testFixtures/java/org/sonar/iac/common/testing/IacTestUtils.java
```diff
@@ -92,4 +92,14 @@ public static InputFileContext createInputFileContextMock(String filename, Strin
...
+  public static InputFileContext createInputFileContextMockFromContent(String content, String filename, String languageKey) {
+    var inputFileContext = createInputFileContextMock(filename, languageKey);
+    try {
+      when(inputFileContext.inputFile.contents()).thenReturn(content);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    return inputFileContext;
+  }
 }
```
nils-werner-sonarsource, line 105: Please provide a good description to this yet another mocking method. I feel we actually loose control over all the test helper method to mock something.

nils-werner-sonarsource (changes requested):
Why do we need to have a shift method inside an immutable class? The location should already be created with the correct position and length. We also should wait to add another common entity that handles a location in the file. Why does TextRange not suit you anymore? Your PR increases the technical debt more. We should avoid this—an evident WHY should be defined before generalizing or introducing a new common concept/data class. TBH, I can't see the why in this case. Please clarify first. I am happy to discuss this.

iac-common/src/main/java/org/sonar/iac/common/api/tree/impl/LocationImpl.java
```diff
@@ -133,6 +131,11 @@ public TextRange toTextRange(String sourceCode) {
     return new TextRange(start, end);
   }
 
+  @Override
+  public Location shift(int positionShift, int lengthShift) {
+    return new LocationImpl(position + positionShift, length + lengthShift);
+  }
+
```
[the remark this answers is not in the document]
mstachniuk: Why? The class is still immutable. It is a common pattern in functional programming and even in Java, see:
`String.formatted()`, `String.substring()`, `LocalDate.atTime()`.
All those methods operate on immutable object and return new immutable object based on orginal one. The original is unchanged.

iac-common/src/testFixtures/java/org/sonar/iac/common/testing/IacTestUtils.java
mstachniuk: With my change the `InputFileContex` passed to `TreeParser.parse(String source, @Nullable InputFileContext inputFileContext)` has to be fully mocked (i.e. `InputFileContex.inputFile.content()` returns a real value).
Before, we didn't need that because `ArmTemplateExpressionParser` didn't need access to the real course code. 

The alternative solution is to store file content in the `ArmJsonBaseConverter` field (and pass it to many places that will generate more changes.

mstachniuk:
> Why does TextRange not suit you anymore?

The `Location` and `LocationImpl` is not a new concept it was introduced for Helm parser (the location returned by Helm Go binary is returned in that format). Here I only moved it from Kubernetes package to common.

The problem with `TextRange` is that it is a good representation for humans but very bad for computations. Shifting locations of parsed template expression according to `TextRange.start()` is hard, it is not just simple addition. And try to extract text based on `TextRange` from String - we don't have a method for that.
Such operations are super simple if we convert `TextRange` to `Location`. The logic for that already existed in our codebase I just reused it.

iac-common/src/testFixtures/java/org/sonar/iac/common/testing/IacTestUtils.java
mstachniuk: But it's a valid point - another method should be replaced by this one.

PR #1257, opened 2024-04-18
SONARIAC-1425 Split S6956 implementation into 2 rules

mstachniuk:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/PropertiesOrderCheck.java
```diff
@@ -0,0 +1,36 @@
...
+package org.sonar.iac.arm.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.iac.arm.checks.elementsorder.PropertiesOrderBicep;
+import org.sonar.iac.arm.checks.elementsorder.PropertiesOrderJson;
+import org.sonar.iac.common.api.checks.IacCheck;
+import org.sonar.iac.common.api.checks.InitContext;
+
+@Rule(key = "S6956")
+public class PropertiesOrderCheck implements IacCheck {
```
petertrr, line 29: Maybe call it `TopLevelPropertiesOrderCheck`? I get confused, because for me it seems "properties" describe resource's `properties` block.

petertrr (approved):
LGTM, please see my concern about naming

PR #1258, opened 2024-04-18
SONARIAC-1373 S6952: Redundant explicit dependencies between resources should be removed

petertrr:


iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.json
```diff
@@ -0,0 +1,31 @@
...
+  "resources": [
+    {
+      "apiVersion": "2019-04-01",
+      "type": "Microsoft.Network/networkInterfaces",
+      "name": "noncompliant1",
+      "location": "[resourceGroup().location]",
+      "dependsOn": [
+        "[resourceId('Microsoft.Network/virtualNetworks', 'exampleVNet')]",
+        "exampleDependency",
+        "[concat('Microsoft.DBforPostgreSQL/servers/', variables('serverName'))]"
```
mstachniuk, line 13: 1. The variable `serverName` should be defined to have a complete example
2. Please also add a negative case here (only explicit dependency without any implicit reference)

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
```diff
@@ -0,0 +1,93 @@
...
+//                                                  ^^^^^^^^^^^^^^^
+  ]
+  properties: {
+    ipConfigurations: [
+      {
+        name: 'ipconfig1'
+        properties: {
+          subnet: {
+            id: exampleSubnet.id
+//              ^^^^^^^^^^^^^<
```
mstachniuk, line 47:
```suggestion
//              ^^^^^^^^^^^^^
```

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
```diff
@@ -0,0 +1,93 @@
...
+          }
+        }
+      }
+      {
+        name: 'ipconfig2'
+        properties: {
+          subnet: {
+            id: reference('exampleSubnet').id
+//                        ^^^^^^^^^^^^^^^<
```
mstachniuk, line 56:
```suggestion
//                        ^^^^^^^^^^^^^^^
```

iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/RedundantResourceDependenciesCheckTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.sonar.iac.common.testing.Verifier;
+
+class RedundantResourceDependenciesCheckTest {
+  @Test
+  void shouldRaiseIssuesInJson() {
+    // TODO SONARIAC-1426: S6952: Cover `*ResourceId` functions in the `dependsOn` block
+    // should also raise on line 11 (`dependsOn` with `resourceId`). On line 13 there is `dependsOn` with more complex function,
+    // which we can't resolve, because it's dynamic, so it's a known FN.
+    ArmVerifier.verify("RedundantResourceDependenciesCheck/test.json", new RedundantResourceDependenciesCheck(),
+      Verifier.issue(12, 8, 12, 27, "Remove this explicit dependency as it is already defined implicitly."));
```
mstachniuk, line 32: 1. Please also add a secondary location here for verification
2. static import

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
```diff
@@ -0,0 +1,93 @@
...
+        }
+      }
+    ]
+  }
+}
+
+resource publicIPAddresses 'Microsoft.Network/publicIPAddresses@2021-05-01' = [for i in range(0, 3): {
+  name: '${publicIPAddressName}${i}'
+  location: location
+}]
```
mstachniuk, line 67: Why this case is here?

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
```diff
@@ -0,0 +1,93 @@
...
+resource nonCompliant3 'Microsoft.Network/networkInterfaces@2020-06-01' = {
+  dependsOn: [
+    publicIPAddresses[0] // Noncompliant
+  ]
+  properties: {
+    name: publicIPAddresses[0].name
+  }
+}
+
+resource compliant 'Microsoft.Network/networkInterfaces@2020-06-01' = {
```
mstachniuk, line 78: please also add compliant example with explicit `dependsOn`.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/RedundantResourceDependenciesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+      }
+
+      var symbolicReferences = new ArrayList<TextTree>();
+      for (var referencedSymbol : referencedResourcesSymbolic) {
+        var referencedName = referencedSymbol.value();
+        if (dependencyName.equals(referencedName)) {
+          symbolicReferences.add(referencedSymbol);
+        }
+      }
```
mstachniuk, line 93: Those 2 loops can be rewritten using stream API - as the stream API is used in this check already a lot, it will be good to have consistency in the scope of one file (until there is a better reason not to use stream API, e.g., performance, but IMO it is not a case here).

mstachniuk (approved):
Good progress on the ticket. Indeed it is a complex check and has a lot of cases. It was a good idea to split it. I like your comments with doc references as it is not always clear how to do things. 
Please also check (it might be a note in new ticket) how dependencies work with nessted resources.
Please see my questions, mostly nitpicks and clarification questions.

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
petertrr: It's a secondary location

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
petertrr: It's a secondary location

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/RedundantResourceDependenciesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return null;
+  }
+
+  private static List<TextTree> collectReferences(ResourceDeclaration resourceDeclaration) {
+    // Traverse all nodes of all property values and find expressions that can be references to other resources. This includes:
+    // * function calls to `reference[s]?` function
+    // * `Identifier`s, because they can be usages of symbolic names
+
+    return resourceDeclaration.properties().stream().flatMap(property -> getAllPropertyValues(property.value()))
```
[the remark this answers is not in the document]
petertrr: According to [docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#valid-uses-1), the `reference` function is somewhat limited, but can be applied outside of resource properties.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/RedundantResourceDependenciesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+      .flatMap(expression -> ArmTreeUtils.findAllNodes(expression, FunctionCall.class))
+      .filter(IS_REFERENCE_FUNCTION)
+      .map(functionCall -> functionCall.argumentList().elements().get(0))
+      .filter(StringLiteral.class::isInstance)
+      .map(TextTree.class::cast)
```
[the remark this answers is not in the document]
petertrr: Doesn't really make much difference; I thought the result will be broader originally. I'll change it to `StringLiteral` now.

iac-extensions/arm/src/test/resources/checks/RedundantResourceDependenciesCheck/test.bicep
petertrr: This is used in `dependsOn` of a next resource. I included it, because it enables `dependsOn` with an element of an array

petertrr:
> Please also check how dependencies work with nested resources.

AFAIU ([1](https://stackoverflow.com/questions/53068960/nested-arm-template-and-dependson), [2](https://stackoverflow.com/questions/44741476/azure-arm-dependencies-between-nested-templates?rq=3)), dependencies cannot be across parent-nested border, so if we have a valid template, we shouldn't check if we are inside nested deployment or not.

PR #1259, opened 2024-04-19
SONARIAC-1405 Fix ARM template expression conversion of multiline strings

nils-werner-sonarsource:


iac-common/src/main/java/org/sonar/iac/common/yaml/tree/YamlTreeMetadata.java
```diff
@@ -29,69 +33,134 @@
...
 
-    // endMark is not present. This happens for example when we have a file with only a comment.
-    // in that case, the root node will be an empty MappingNode with only a startMark to which the comment is attached
-    return endMark.map(mark -> TextRanges.range(startLine, startColumn, mark.getLine() + 1, mark.getColumn()))
-            .orElseGet(() -> TextRanges.range(startLine, startColumn, startLine, startColumn));
-  }
+    private static int pointer(@Nullable Mark mark) {
+      if (mark == null) {
+        throw new ParseException("Nodes are expected to have a start mark during conversion", null, null);
+      }
```
nils-werner-sonarsource, line 118: To cover this, I would have to write an artificial test case. There is no case that this will be null. But it's the result from SnakeYaml. So better safe than sorry.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmJsonBaseConverterTest.java
```diff
@@ -165,4 +170,51 @@ void shouldShiftTextRangesOfExpressionNodes() {
...
+    var expression = converter.toExpression(functionScalar);
+
+    assertThat(expression.getKind()).isEqualTo(ArmTree.Kind.FUNCTION_CALL);
+
+    var functionCall = ((FunctionCall) expression);
+    assertThat(functionCall.textRange()).hasRange(2, 14, 4, 5);
+
+    var identifier = functionCall.name();
+    assertThat(identifier.value()).isEqualTo("function_call");
+    assertThat(identifier.textRange()).hasRange(2, 15, 2, 28);
```
mstachniuk, line 210: IMO it should be hasRange(2, 16, 2, 29); if I'm not mistaken

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmJsonBaseConverterTest.java
```diff
@@ -165,4 +170,51 @@ void shouldShiftTextRangesOfExpressionNodes() {
...
+
+    var arguments = functionCall.argumentList().elements();
+    assertThat(arguments).hasSize(1);
+
+    var argument = arguments.get(0);
+    assertThat(argument.getKind()).isEqualTo(ArmTree.Kind.STRING_LITERAL);
+    assertThat(argument).hasRange(3, 19, 3, 27);
```
mstachniuk, line 217: And here for sure the values are wrong as line 3  has only 12 characters. IMO the shifting without converting first to Location (or pointer offset from the beginning of the file) doesn't work. I had the same problem in my PR and it is a why I did this conversion to Location.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmJsonBaseConverterTest.java
```diff
@@ -165,4 +170,51 @@ void shouldShiftTextRangesOfExpressionNodes() {
...
+    assertThat(((StringLiteral) argument).value()).isEqualTo("foobar");
+  }
```
mstachniuk, line 219: Please also add the case when string contains "\\n"

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmJsonParserTest.java
```diff
@@ -0,0 +1,44 @@
...
+    var parser = new ArmJsonParser();
+
+    var file = parser.parseJson("""
+      {
+        "string": "line1
+          line2"
+      }
+      """);
+    ScalarTree scalar = (ScalarTree) ((MappingTree) file.documents().get(0)).elements().get(0).value();
+    assertThat(scalar.value()).isEqualTo("line1\n    line2");
```
mstachniuk, line 41: please add assertions for TextRanges

mstachniuk (changes requested):
I request changes as TextRanges of parsed ARM expression inside a scallar contins wrong locations what can cause more IllegalArgumentExceptions in syntax highlighter (and checks)

iac-extensions/arm/src/test/java/org/sonar/iac/arm/parser/ArmJsonBaseConverterTest.java
nils-werner-sonarsource: The String already contains a line break `"function": "[function_call(\n          'foobar'\n        )]"`. What do you expect to see when `foobar` is split?

mstachniuk: This is an example what I had in mind: https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/ASimFileEvent/ARM/vimFileEventVMwareCarbonBlackCloud/vimFileEventVMwareCarbonBlackCloud.json#L38

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/ArmJsonParser.java
```diff
@@ -73,4 +81,26 @@ private static void setParents(ArmTree tree) {
       setParents(child);
     }
   }
+
+  class ArmJsonConverter extends YamlConverter {
+
+    @Override
+    protected YamlTree convertScalar(ScalarNode node) {
+      var metadata = YamlTreeMetadata.fromNode(node);
```
mstachniuk, line 89: Please use in new code not a deprecated method

mstachniuk (approved):
LGTM

PR #1267, opened 2024-04-24
SONARIAC-1427 Test the bulk migration using the starlark CLI

olivier-schmitt-sonarsource:
This PR tests the automated migration tool: https://github.com/SonarSource/languages-experimental-tooling/pull/211

The 3 commits reflect the process:
- adjust the original YAML file to make life easier for the tool
- use the tool to generate the Starlark file
- adjust the Starlark file to re-order keys

The Starlark file:
- contains env definition
- contains tasks definitions
- does not not contain any function mapped to YAML anchors

This way it's easier to spot common constructs and to come with an optimal Starlak implementation.


.cirrus.star
```diff
@@ -1,4 +1,430 @@
+# Check Starlark specs: https://github.com/bazelbuild/starlark/blob/master/spec.md
+# Cirrus
+load("cirrus", "yaml", "fs")
```
mstachniuk, line 3: Why is it needed? If I see correctly none of yaml or fs functions are used.

.cirrus.star
```diff
@@ -1,4 +1,430 @@
...
+                "cirrus_jfrog_promote multi",
+                "source ${PROJECT_VERSION_CACHE_DIR}/evaluated_project_version.txt",
+                "github-notify-promotion",
+                "burgr-notify-promotion"
+            ]
+        }
+    }
+
+
 def main(ctx):
```
mstachniuk, line 419: It will be good to have this function at the top of the file if possible. It is good to see the list of tasks at the beginning.

.cirrus.star
```diff
@@ -1,4 +1,430 @@
...
-    return load_features(ctx)
+    conf = dict()
+    merge_dict(conf, load_features(ctx))
+    merge_dict(conf, env())
+    merge_dict(conf, build_task())
```
mstachniuk, line 423: I like the fact that functions definitions are located in the file in the same order as here (and in the order as they are executed).

.cirrus.star
```diff
@@ -1,4 +1,430 @@
...
+                "memory": "6G",
+                "cpu": 10
+            },
+            "project_version_cache": {
+                "folder": "$PROJECT_VERSION_CACHE_DIR",
+                "fingerprint_script": "echo $BUILD_NUMBER",
+                "populate_script": "mkdir -p \"$PROJECT_VERSION_CACHE_DIR\"\ntouch \"$PROJECT_VERSION_CACHE_DIR\"/evaluated_project_version.txt\n",
+                "reupload_on_changes": "true"
+            },
+            "create_gradle_directory_script": [
```
mstachniuk, line 76: This block is used 6 times in the whole file. Is it expected such duplication? I thought the template would have more building blocks and the duplication would be avoided.

mstachniuk (commented):
I have some clarification questions

mstachniuk:
General observation. The change looks like just another format. Instead of YAML we put everything into dictionaries. Is it planned to use more some imperative approach?

.cirrus.star
olivier-schmitt-sonarsource: Part of the template in the migration tool but indeed those helper functions are not required in this pipeine.

.cirrus.star
olivier-schmitt-sonarsource: Done in the next iteration, the bulk migration is a temporary state toward the final pipeline definition.

.cirrus.star
olivier-schmitt-sonarsource: This is fixed in the next iteration.
The duplication helps to expose the common building blocks in an explicitly way, then the DRY phase can start.

olivier-schmitt-sonarsource:
> General observation. The change looks like just another format. Instead of YAML we put everything into dictionaries. Is it planned to use more some imperative approach?

Yes, done, in the next iteration here: 
https://github.com/SonarSource/sonar-iac/pull/1268

mstachniuk (approved):
Thanks for answering my questions, LGTM

nils-werner-sonarsource (approved):
LGTM!

PR #1269, opened 2024-04-25
Add GH workflow to update rule metadata

petertrr:


.github/workflows/rule-metadata-update.yml
```diff
@@ -0,0 +1,34 @@
...
+          ./gradlew :ruleApiUpdate -PartifactoryUsername=${ARTIFACTORY_ROLE} -PartifactoryPassword=${ARTIFACTORY_ACCESS_TOKEN}
+        env:
+          ARTIFACTORY_ROLE: vault-${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ROLE }}
+          ARTIFACTORY_ACCESS_TOKEN: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+      - uses: peter-evans/create-pull-request@v6
+        with:
+          author: ${{ github.actor }} <${{ github.actor }}>
+          commit-message: Update rule metadata
+          title: Update rule metadata
+          base: master
```
jonas-wielage-sonarsource, line 32: Am I correct that this can only be executed on master?

jonas-wielage-sonarsource (approved):
LGTM

.github/workflows/rule-metadata-update.yml
petertrr: It will create a PR targeting master, yes. I didn't think of the use-case where we release from another branch...

I can add a parameter for the workflow, which will default to `master`

petertrr: As discussed, we probably will always target `master` and cherry-pick to feature branches if required. And a PR can always be edited after it's created. So I removed the configurable property

PR #1273, opened 2024-04-26
SONARIAC-1435 Generate Parser and Visitor with ANTLR for properties file

mstachniuk:


iac-extensions/spring-config/build.gradle.kts
```diff
@@ -1,11 +1,14 @@
 plugins {
+    antlr
     id("org.sonarsource.iac.code-style-convention")
     id("org.sonarsource.iac.java-conventions")
 }
 
 description = "SonarSource IaC Analyzer :: Extensions :: Spring Config"
 
 dependencies {
+    antlr("org.antlr:antlr4:4.13.1")
```
petertrr, line 10: Please add this to `libs.versions.toml`

iac-extensions/spring-config/build.gradle.kts
```diff
@@ -15,3 +18,17 @@ dependencies {
     testImplementation(libs.sonar.analyzer.test.commons)
     testImplementation(testFixtures(project(":iac-common")))
 }
+
+tasks.generateGrammarSource {
+    maxHeapSize = "64m"
+    arguments = arguments + listOf("-visitor", "-long-messages", "-no-listener", "-package", "org.sonar.iac.springconfig.parser.properties")
+    // Due to bug in ANTLR Gradle plugin https://stackoverflow.com/a/49388412
```
petertrr, line 25: I don't get what the bug is. Is it that ANTLR source is not in a package but we want the generated file to be? Then I wouldn't call it a bug and rephrase the comment like `antrl plugin requires grammar file to be placed without package in path, but we want the generated file to follow standard Java convention`

iac-extensions/spring-config/build.gradle.kts
```diff
@@ -15,3 +18,17 @@ dependencies {
...
+    outputDirectory =
+        project.layout.buildDirectory.dir(
+            "generated-src/antlr/main/org/sonar/iac/springconfig/parser/properties/"
```
petertrr, line 28: Can it be simple `generated-src/main` and not `generated-src/antlr/main`?

build-logic/src/main/kotlin/org.sonarsource.iac.code-style-convention.gradle.kts
```diff
@@ -14,7 +14,7 @@ spotless {
                 )
             )
         licenseHeaderFile(rootProject.file("LICENSE_HEADER")).updateYearWithLatest(true)
-        targetExclude("*/generated-sources/**")
+        targetExclude("*/generated-sources/**", "*/generated-src/**")
```
petertrr, line 17: Why not configure it to output to `generated-sources` as well?

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
```diff
@@ -0,0 +1,211 @@
...
+      "visitRow mail.defaultRecipients[1]=owner@mail.com<EOF>",
+      "visitLine mail.defaultRecipients[1]=owner@mail.com<EOF>",
+      "visitKey mail.defaultRecipients[1]",
+      "visitKey owner@mail.com",
+      "visitEol <EOF>");
+  }
+
+  @Test
+  void shouldParseDifferentNewLines() {
+    var code = "foo1=bar1\nfoo2=bar2\rfoo3=bar3\r\nfoo4=4\n";
```
petertrr, line 131: I would also suggest adding tests with empty lines between properties to see how many trailing newlines are included in the text. And generally, I would suggest adding more examples based on the repo with the grammar (multiline values, etc.)

petertrr (commented):
Good progress, please see my questions

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
jonas-wielage-sonarsource: Specifically this file contains a lot of cases: 
https://github.com/antlr/grammars-v4/blob/master/properties/examples/example1.txt

build-logic/src/main/kotlin/org.sonarsource.iac.code-style-convention.gradle.kts
mstachniuk: The `generated-src` was a default. We have to anyway configure it, so I'm ok to change it.

iac-extensions/spring-config/build.gradle.kts
mstachniuk: It has to start with `generated-src/antlr/main` as this prefix is used as output to the task. Not following this convention requires additional configuration of this task or `sourcesJar`.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
```diff
@@ -0,0 +1,792 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.antlr.v4.runtime.CharStreams;
+import org.antlr.v4.runtime.CommonTokenStream;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class PropertiesParserBaseVisitorTest {
```
petertrr, line 31: Could you maybe also add one test for text ranges to demonstrate that what is provided is reasonable? The visitor can collect this data in another list, and in a test also string matching can be done.

petertrr (approved):
I don't have any major questions, so it LGTM

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
```diff
@@ -0,0 +1,792 @@
...
+      "visitKey bar3",
+      "visitEol \\u2029");
+  }
+
+  @Test
+  void shouldAcceptAllCharactersInComments() {
+    var code = """
+      #comment 1 ! # : =
+      !comment 2 # aa
+      foo2=bar2""";
```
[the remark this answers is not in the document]
mstachniuk: Sure I will add it. Trailing comments are not allowed - they will be a part of the value.

PR #1275, opened 2024-04-26
SONARIAC-1438 Implement "SpringConfigTree"

jonas-wielage-sonarsource:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/ProfileImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+import org.sonar.iac.common.api.tree.Comment;
+import org.sonar.iac.common.api.tree.Tree;
+import org.sonar.iac.springconfig.tree.api.Profile;
+import org.sonar.iac.springconfig.tree.api.Tuple;
+
+public class ProfileImpl extends AbstractSpringConfigImpl implements Profile {
+  private final List<Tuple> properties;
+
+  public ProfileImpl(List<Tuple> properties) {
```
petertrr, line 32: Should other fields (`comments`, `name`, `isActive`) also be set in the constructor?

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/ScalarImpl.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public SyntaxToken value() {
+    return value;
+  }
+
+  @Override
+  public List<Tree> children() {
+    return Arrays.asList(value);
```
petertrr, line 42: `List.of`?

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/TupleImpl.java
```diff
@@ -0,0 +1,51 @@
...
+  private final Scalar value;
+
+  public TupleImpl(Scalar key, Scalar value) {
+    this.key = key;
+    this.value = value;
+  }
+
+  @Override
+  public List<Tree> children() {
+    return Arrays.asList(key, value);
```
petertrr, line 39: `List.of` to have an immutable collection?

petertrr (approved):
LGTM, but please see my questions

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/ScalarImpl.java
jonas-wielage-sonarsource: I used `Arrays.asList()` to be able to use a null value and not resulting in an NPE.
I'm not sure we ever run into this problem though.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/TupleImpl.java
petertrr: OK, that may make sense :thinking: 
However, then we need to annotate potentially nullable parameters. I believe in a valid properties file `key` will never be null, but the `value` may. OTOH, we can create a tuple with an artificial empty SyntaxToken in such case.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/ScalarImpl.java
jonas-wielage-sonarsource: I changed to `List.of`, as we don't expect null values

PR #1279, opened 2024-04-29
Update sonar-formatter to version 23

renaud-tognelli-sonarsource:
Update of sonar-formatter to version 23 so that records format are specified.

As discussed, `Blowdryer.immutableUrl` should point to an immutable file, as it will never be checked for updates.


petertrr (approved):
LGTM

PR #1280, opened 2024-04-30
SONARIAC-1437 Implement metrics and highlighting visitors for `.properties` files

petertrr:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
```diff
@@ -95,6 +95,6 @@ private static TextRange textRange(ParserRuleContext ctx) {
       ctx.start.getLine(),
       ctx.start.getCharPositionInLine(),
       ctx.stop.getLine(),
-      ctx.stop.getCharPositionInLine());
+      ctx.stop.getCharPositionInLine() + 1);
```
petertrr, line 98: Looking at the YAML tree, the end offset of tokens corresponds to the index of the next character after the last. And highlighting verifier also expects them like that. W/o this adjustment registered highlighting is by one character not enough.

mstachniuk: I guess it should be fixed in `PropertiesParseTreeVisitor.textRange` as we will face the same problem on issue highlighting.

sonar-helm-for-iac/src/org.sonar.iac.helm/template-evaluation.pb.go
```diff
@@ -18,7 +18,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
+// 	protoc-gen-go v1.33.0
```
mstachniuk, line 21: Why is this committed? Locally I have the same change.
I guess some software update updated also `protoc-gen-go` on our machines. 
Maybe it should be updated in separate PR + CI configuration?

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
mstachniuk: Good catch. Please add a comment on why this adjustment is needed. It means that ANTLR treats the end as the position of the last character.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigMetricsVisitor.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.sonar.iac.springconfig.plugin.SpringConfigSensor.isPropertiesFile;
+
+public class SpringConfigMetricsVisitor extends YamlMetricsVisitor {
+  public SpringConfigMetricsVisitor(FileLinesContextFactory fileLinesContextFactory, NoSonarFilter noSonarFilter) {
+    super(fileLinesContextFactory, noSonarFilter);
+  }
+
+  @Override
+  protected boolean acceptFileForLoc(InputFileContext inputFileContext) {
+    // YAML files should be processed line-by-line; properties files will be handled by a visitor
```
mstachniuk, line 38: 1. I guess it will be handled by `YamlMetricsVisitor`? The comment is not clear enought.
2. If yes, it will not be correctly, because of `YamlMetricsVisitor:61`: `        if (!(line.isBlank() || line.startsWith("#"))) {` - in properties file comments also starts from `!`. Please adjust that and add test case.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/SyntaxTokenImpl.java
```diff
@@ -38,6 +39,11 @@ public String value() {
     return value;
   }
 
+  @Override
+  public List<Comment> comments() {
```
mstachniuk, line 43: Why this is needed? We don't attach comments to the SyntaxTokens, AFAIK.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigMetricsVisitorTest.java
```diff
@@ -0,0 +1,97 @@
...
+    scan(
+      // language=properties
+      """
+        # Comment 1
+        sonar.host.url=http://localhost:9000
+
+
+        sonar.projectKey=sonar-iac
+        sonar.projectName=sonar-iac
+        # Comment 2
```
mstachniuk, line 63: please add a comment that starts from `!`

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigMetricsVisitorTest.java
```diff
@@ -0,0 +1,97 @@
...
+        spring:
+        # datasource
+          datasource:
+            url: jdbc:mysql://127.0.0.1:3306/example-database
+            driver-class-name: com.mysql.jdbc.Driver
+            username: root
+            password:
+        # cache
+          cache:
+            type: ehcache
```
mstachniuk, line 89: Please add empty line to validate how it is calculated

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigHighlightingVisitorTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+class SpringConfigHighlightingVisitorTest extends AbstractHighlightingTest {
+  protected SpringConfigHighlightingVisitorTest() {
+    super(new SpringConfigHighlightingVisitor(), new SpringConfigParser());
+  }
+
+  @Test
+  void shouldHighlightProperties() {
+    highlight(
+      // language=properties
```
mstachniuk, line 41: Why such comment?

mstachniuk (commented):
Good progress on it, please see my comments. Handling comments that start with `!` must be fixed before accepting the PR.

sonar-helm-for-iac/src/org.sonar.iac.helm/template-evaluation.pb.go
petertrr: Yes, `protoc-gen-go` was updated by Renovate (with manual additions by me), but I must've not regenerated the code. Yes, I can look at it as a separate thing and rollback here. Probably, something like a simple check in the build task to ensure that no changes are left uncommitted in the Go code should do.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigMetricsVisitor.java
petertrr: My idea is the following. We lose a lot of information when flattening the YAML tree, so we shouldn't directly register metrics visitor on SyntaxToken nodes for YAML files. For them, the implementation from the base class will be called to calculate metrics line-by-line. For properties files, we can use the tree. I updated the comment.
I don't quite get your question. Implementation of the YamlMetricsVisitor is not called for properties, so there shouldn't be a problem with comments. Or did you mean something else?

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/tree/impl/SyntaxTokenImpl.java
petertrr: Yes, it is needed for the `defaultMetricsVisitor`

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigHighlightingVisitorTest.java
petertrr: It enables syntax highlighting for the string content in IDEA. Not sure if it is useful enough to keep, but it can help to double-check that an example has valid syntax.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigMetricsVisitorTest.java
petertrr: Good catch. Currently it would be filteterd out by `MetricsVisitor:81`, where it would match `isBlank()`. I'll try to do something with this.

sonar-helm-for-iac/src/org.sonar.iac.helm/template-evaluation.pb.go
petertrr: https://github.com/SonarSource/sonar-iac/pull/1289

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/plugin/visitors/SpringConfigMetricsVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Override
+  protected void languageSpecificMetrics() {
+    register(SyntaxToken.class, defaultMetricsVisitor());
+    register(Profile.class, (ctx, tree) -> addCommentLines(tree.comments()));
+  }
+
+  @Override
+  protected boolean isBlank(String line) {
```
mstachniuk, line 50: Nitpick. Now, the method name is confusing. Please find a better name.

mstachniuk (approved):
LGTM and please see my new comment.

PR #1281, opened 2024-04-30
SONARIAC-1433 Implement custom visitor to transform the generated AST to "SpringConfigTree"

mstachniuk:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
```diff
@@ -0,0 +1,84 @@
...
+        properties.add(new TupleImpl(keyScalar, valueScalar));
+      }
+      if (row.comment() != null) {
+        var commentContext = row.comment().commentStartAndText();
+        var value = commentContext.getText();
+        var contentText = commentContext.commentText().getText();
+        comments.add(new CommentImpl(value, contentText, textRange(commentContext)));
+      }
+    }
+    return new FileImpl(List.of(new ProfileImpl(properties, comments, "", true)));
```
petertrr, line 58: I believe the default value should be `default`: https://docs.spring.io/spring-boot/reference/features/profiles.html#:~:text=The%20name%20of%20the%20default,Properties

mstachniuk: IMO, in case the profile is not defined, it should be empty. Thanks to that we can distinguish the following situation:

```
foo=bar
#---
spring.profiles.active=default
foo=baz
```

jonas-wielage-sonarsource: I'm not sure, if they wouldn't just be the same profile then, and one foo overrides the other...

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+    for (PropertiesParser.RowContext row : ctx.row()) {
+      if (row.line() != null) {
+        var keyScalar = createKeyScalar(row);
+        var valueScalar = createValueScalar(row);
+        properties.add(new TupleImpl(keyScalar, valueScalar));
+      }
+      if (row.comment() != null) {
+        var commentContext = row.comment().commentStartAndText();
+        var value = commentContext.getText();
+        var contentText = commentContext.commentText().getText();
```
jonas-wielage-sonarsource, line 59: Nitpick: Could also be extracted into a separate method
`private Comment retrieveCommentFromRow(Row row) {...}`

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/SpringConfigAssertions.java
```diff
@@ -0,0 +1,41 @@
...
+    return TextRangeAssert.assertThat(actual);
+  }
+
+  public static TextRangePluginApiAssert assertThat(@Nullable org.sonar.api.batch.fs.TextRange actual) {
+    return TextRangePluginApiAssert.assertThat(actual);
+  }
+
+  public static ExternalIssueAssert assertThat(@Nullable ExternalIssue issue) {
+    return ExternalIssueAssert.assertThat(issue);
+  }
```
jonas-wielage-sonarsource, line 40: What is the benefit of wrapping the calls again? For me it creates yet another layer to use / forget.

jonas-wielage-sonarsource: I feel like, most of the time only the TextRangeAssert is needed.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+import org.sonar.iac.springconfig.tree.api.Tuple;
+import org.sonar.iac.springconfig.tree.impl.FileImpl;
+import org.sonar.iac.springconfig.tree.impl.ProfileImpl;
+import org.sonar.iac.springconfig.tree.impl.ScalarImpl;
+import org.sonar.iac.springconfig.tree.impl.SyntaxTokenImpl;
+import org.sonar.iac.springconfig.tree.impl.TupleImpl;
+
+public class PropertiesParseTreeVisitor extends PropertiesParserBaseVisitor<SpringConfig> {
+
+  private static final List<String> PROFILE_NAME_PROPERTIES = List.of("spring.profile", "spring.config.active.on-profile");
```
jonas-wielage-sonarsource, line 42: Should be changed to `"spring.profiles.active", "spring.config.activate.on-profile"`.
I updated the ticket description for this.
Additionally there can be the spring.config.activate.on-cloud-platform which we may want to support.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitorTest.java
```diff
@@ -0,0 +1,249 @@
...
+    assertThat(databaseKeyValue.key().value().value()).isEqualTo("foo");
+    assertThat(databaseKeyValue.key().textRange()).hasRange(1, 0, 1, 2);
+    assertThat(databaseKeyValue.key().value().textRange()).hasRange(1, 0, 1, 2);
+
+    assertThat(databaseKeyValue.value()).isNull();
+  }
+
+  @Test
+  void shouldReadSimpleProperties() {
+    var code = """
```
jonas-wielage-sonarsource, line 69: Do we also want to test the colon as a delimiter?
Or a key with an array inside?

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+    return valueScalar;
+  }
+
+  private static String profileName(List<Tuple> properties) {
+    return properties.stream()
+      .filter(tuple -> PROFILE_NAME_PROPERTIES.contains(tuple.key().value().value()))
+      .filter(tuple -> tuple.value() != null)
+      .map(tuple -> tuple.value().value().value())
+      .collect(Collectors.joining(" "));
+  }
```
jonas-wielage-sonarsource, line 96: Not sure about the concatenation, see other discussion.
Nevertheless, this method can be extracted into a separate class, because it makes sense that I use it too and don't implement the same logic again)
(Or I will extract it later)

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/SpringConfigAssertions.java
mstachniuk: Yes, mostly `TextRangeAssert` is needed because we are creating another assertions when we see the benefit from it. See `DockerAssertions` it contains more useful assertions.
The befit is to have all assertions for the module in one place. It is recommended by AssertJ: https://assertj.github.io/doc/#assertj-core-custom-assertions-entry-point

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
mstachniuk: Changed to `spring.profiles.active`
The `spring.config.activate.on-cloud-platform` I will not add here to not mix profiles and cloud platforms.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitorTest.java
mstachniuk: It is already tested in `PropertiesParserBaseVisitorTest` but I'm ok to add it here as well as it is a test on different abstraction level.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/PropertiesParseTreeVisitor.java
mstachniuk: I'm also not sure and we should validate it later and adapt.

jonas-wielage-sonarsource (approved):
LGTM, maybe we should create a ticket about verifying profile logic

PR #1284, opened 2024-05-02
SONARIAC-1443 Implement ruling tests for `spring-config` extension

petertrr:


its/ruling/src/integrationTest/java/org/sonarsource/iac/IacRulingTest.java
```diff
@@ -137,6 +137,14 @@ void testArm() throws IOException {
     runRulingTest("azureresourcemanager", properties);
   }
 
+  @Test
+  void testSpringConfig() throws IOException {
+    var properties = Map.of(
+      "sonar.inclusions", "sources/spring-config/**/*.properties,sources/spring-config/**/*.yml,sources/spring-config/**/*.yaml",
```
petertrr, line 143: I thought about narrowing it to `application*` files to avoid the noise from other issues discovered in YAMLs, however I believe we can also test how the spring-config sensor selects Spring-related files if we don't artificially filter others out.

petertrr:
> Why https://github.com/in28minutes/spring-boot-examples was not added?

I looked at the number of files in this repo and thought it would be enough for the integration test. Do you think the other repo is also needed?

mstachniuk:
It is ok to merge as it is too.

PR #1286, opened 2024-05-02
SONARIAC-1434 Implement "SpringConfigPropertiesParser" to manage ANTLR parsing and transformation of properties files

mstachniuk:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/ErrorListener.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Nullable
+  private InputFileContext inputFileContext;
+
+  public ErrorListener(@Nullable InputFileContext inputFileContext) {
+    this.inputFileContext = inputFileContext;
+  }
+
+  @Override
+  public void syntaxError(Recognizer<?, ?> recognizer, Object offendingSymbol, int line, int charPositionInLine, String msg, RecognitionException e) {
```
petertrr, line 41: Is it safe to ignore the exception here or can it contain additional information not present in `msg`?

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
```diff
@@ -41,7 +43,7 @@ void shouldParseSimpleKeyValue() {
     parseProperties(code);
 
     assertThat(visitor.visited()).containsExactly(
-      "visitPropertiesFile foo=bar<EOF>",
+      "visitPropertiesFile foo=bar<EOF><EOF>",
```
petertrr, line 46: Why the double EOF?

petertrr (approved):
LGTM

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
mstachniuk: I changed the following rule:

```
propertiesFile
    : row* EOF
    ;
```

I added EOF here.

Thanks to that the whole content of the input will be consumed and I was able to raise error on `=abc` input. Before that such input produced empty `PropertiesFile`.

My guess is that the double `<EOF>` is not a problem at all.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/ErrorListener.java
mstachniuk: The default `ConsoleErrorListener` just prints `msg` and location into stderr. Here, the code throws the exception. I can add debug log, maybe it will be useful.

petertrr: I mean the parameter `RecognitionException e`, maybe it's more descriptive than `msg`. From the documentation it seems the opposite though, so feel free to do it as you see fit

PR #1287, opened 2024-05-02
SONARIAC-1446 Implement "SpringConfigParser"

jonas-wielage-sonarsource:


iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/SpringConfigParserTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Test
+  void shouldParseSimpleProperties() {
+    File tree = (File) parser.parse("foo=bar", inputFilePropertiesContext);
+    assertThat(tree.profiles().get(0).properties()).hasSize(1);
+  }
+
+  @Test
+  void shouldThrowParseExceptionOnEmptyYamlFile() {
+    assertThatThrownBy(() -> parser.parse("", inputFileYamlContext))
```
mstachniuk, line 56: I'm slightly surprised that the empty file throws `ParseException` - this will raise issue S2260.
For Kubernetes, Cloudformation and ARM (another YAML parser) there is a pre-validation in `customFilePredicate()` that the file contains some identifier (= is not empty).
For Spring config we don't have such pre-validation. 
We should not fail on an empty file. I can imagine that someone committed an empty file or created a file (in SonarLint context some day in the future) and in that cases we should not raise S2260.

mstachniuk (changes requested):
Please fix this so that we don't fail on an empty file.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/SpringConfigParserTest.java
jonas-wielage-sonarsource: Actually, that's implemented in the `IacSensor` itself. If you look at line 226 there's this snippet:
```
      if (EMPTY_FILE_CONTENT_PATTERN.matcher(content).matches()) {
        return;
      }
```
Basically we don't parse empty files at all.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/SpringConfigParser.java
```diff
@@ -19,18 +19,38 @@
...
+      }
+    }
+    throw ParseException.createParseException("Unsupported file extension", inputFileContext, null);
+  }
+
+  private static boolean isYamlFile(InputFileContext inputFileContext) {
+    return YamlLanguage.KEY.equals(inputFileContext.inputFile.language());
+  }
+
+  private static boolean isPropertiesFile(InputFileContext inputFileContext) {
```
petertrr, line 53: There is already a method like this in `SpringConfigSensor`. Maybe it can be kept in a single place?

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/SpringConfigParserTest.java
mstachniuk: OK, I missed that.

mstachniuk (approved):
LGTM

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/SpringConfigParserTest.java
jonas-wielage-sonarsource: I will write a small comment to clarify this behavior.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/SpringConfigParser.java
jonas-wielage-sonarsource: Ok, I will move the logic to the Sensor

PR #1289, opened 2024-05-03
Update protobuf-go to 1.34, check if generated Go code is fully committed

petertrr:


.cirrus/modules/build.star
```diff
@@ -40,11 +40,12 @@ def build_script():
         "source cirrus-env BUILD",
         "source .cirrus/use-gradle-wrapper.sh",
         "regular_gradle_build_deploy_analyze ${BUILD_ARGUMENTS}",
+        "echo 'Checking if any files are uncommitted in the Go code (this may happen to the generated code)'",
+        "git diff --exit-code --name-only -- sonar-helm-for-iac/",
```
petertrr, line 44: We don't regenerate Go protobuf definitions automatically. Probably, it can be tuned in `build.gradle.kts` not to forget about it later.

mstachniuk (approved):
LGTM

.cirrus/modules/build.star
petertrr: Added a dependency between tasks. Shouldn't affect performance much, because these tasks are cacheable

PR #1290, opened 2024-05-03
SONARIAC-1448 S2260: Java parsing failure

mstachniuk:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/CRLexerATNSimulator.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final int NEXT_CHAR_TO_CONSUME = 2;
+
+  public CRLexerATNSimulator(Lexer recog, ATN atn, DFA[] dfas, PredictionContextCache sharedContextCache) {
+    super(recog, atn, dfas, sharedContextCache);
+  }
+
+  @Override
+  public void consume(CharStream input) {
+    int curChar = input.LA(CURRENT_CHAR_TO_CONSUME);
+    if (curChar == '\n') {
```
petertrr, line 49: Why is this branch needed, while, AFAIU, for `\r\n` you are not doing a manual line break? Can only `\r` w/o `\n` be handled manually?

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/SpringConfigPropertiesParser.java
```diff
@@ -32,7 +32,7 @@ public class SpringConfigPropertiesParser implements TreeParser<Tree> {
   @Override
   public SpringConfig parse(String source, @Nullable InputFileContext inputFileContext) {
     var inputCode = CharStreams.fromString(source);
-    var propertiesLexer = new PropertiesLexer(inputCode);
+    var propertiesLexer = new PropertiesLexer(inputCode, true);
```
petertrr, line 35: Is the second parameter needed only to select the correct overload?

petertrr (approved):
LGTM, please see my questions

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/SpringConfigPropertiesParser.java
mstachniuk: Yes

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/properties/CRLexerATNSimulator.java
mstachniuk: Because we override the method and the logic for `\n` is needed here. Without we will break break locations when other line separators are used.

petertrr: Ah, okay, this branch will handle both `\n` and `\r\n` after checking if `\r` is not standalone. I see it now

PR #1309, opened 2024-05-13
SONARIAC-1397 S2092: Support detection of misconfigured "Secure" cookie flags in Spring configuration

jonas-wielage-sonarsource:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/checks/AbstractSensitiveKeyCheck.java
```diff
@@ -0,0 +1,57 @@
...
+public abstract class AbstractSensitiveKeyCheck implements IacCheck {
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(Tuple.class, this::checkTuple);
+  }
+
+  protected abstract Set<String> sensitiveKeys();
+
+  protected abstract void reportOnSensitiveValue(CheckContext ctx, Tuple tuple, String value);
```
petertrr, line 40: Maybe `checkValue` or something similar? Because at this point we still need to check if an issue should be raised.

petertrr (approved):
LGTM! Please see my suggestion

PR #1310, opened 2024-05-13
SONARIAC-1393 S6437: Support detection of Hardcoded Secrets for Spring configuration

petertrr:


iac-extensions/spring-config/src/test/resources/checks/HardcodedSecretsCheck/HardcodedSecretsCheck.properties
```diff
@@ -0,0 +1,232 @@
...
+management.signalfx.metrics.export.access-token=${PASSWORD}
+management.wavefront.api-token=${PASSWORD}
+spring.devtools.remote.secret=${PASSWORD}
+
+#---
+# Noncompliant@+1
+spring.datasource.password=${PASSWORD}-suffix
+#---
+# Noncompliant@+1
+spring.datasource.password=prefix-${PASSWORD}
```
jonas-wielage-sonarsource, line 220: Do we actually want to raise here? IMO we should clarify this cases with AppSec.

jonas-wielage-sonarsource (approved):
LGTM, please have a look at my comment.

iac-extensions/spring-config/src/test/resources/checks/HardcodedSecretsCheck/HardcodedSecretsCheck.properties
petertrr: Clarified; for such strings, we shouldn't raise. I'll update the PR.

jonas-wielage-sonarsource: Thanks for clarifying it! It will make it a bit more complicated though

PR #1315, opened 2024-05-15
SONARIAC-1461 Parsing error on `@` character

petertrr:


iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessor.java
```diff
@@ -53,4 +54,12 @@ private static String removeInlineCommentAfterDocumentBreak(String line) {
     }
     return line;
   }
+
+  /**
+   * Surround Maven variable substitution in quotes instead of '@' to avoid parsing issues.
+   * We don't resolve these variables, so we shouldn't lose information by doing this.
+   */
+  private static String transformMavenSubstitutions(String line) {
+    return MAVEN_VARIABLE_AT_TOKEN_START.matcher(line).replaceAll("$1'${name}'$3");
```
petertrr, line 63: I was thinking about either substituting `@...@` with `${,,,}`, but that would require tracking text range adjustments, or tracking the information about the substituted symbols and replacing them back in the converter. However, I don't see an immediate value from it, because it's not likely that maven variable name will clash with a sensitive property value, where we would raise an issue.

iac-common/src/main/java/org/sonar/iac/common/api/tree/impl/TextPointer.java
rudy-regazzoni-sonarsource, line 1: Nice improvement 👍 Records are truly a thing.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessor.java
```diff
@@ -27,23 +28,23 @@ public class SpringConfigYamlPreprocessor {
   private static final Pattern LINE_SEPARATORS = Pattern.compile("(\n|\r|\r\n|\u2028|\u2029)");
```
rudy-regazzoni-sonarsource, line 28: I'm expecting this regex to not work as expected.
This would match `\r` before it has a chance to match `\r\n`, transforming `data1\r\ndata2` into `['data1', '\ndata2']`.
This is not related to this PR, so it's probably better to handle it in a separate ticket/PR.
```suggestion
  private static final Pattern LINE_SEPARATORS = Pattern.compile("(\r\n|\n|\r|\u2028|\u2029)");
```

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessorTest.java
```diff
@@ -68,4 +68,29 @@ void shouldRemoveInlineCommentsAfterDocumentBreak() {
...
+      bar: prefix-@maven.property@-suffix
+      baz: |
+        @maven.property@
+      """;
+    // language=yaml
+    var expected = """
+      foo: 'maven.property'
+      foo2: 'maven.property'-suffix
+      foo3: '@maven.property@'
+      bar: prefix-@maven.property@-suffix
```
rudy-regazzoni-sonarsource, line 88: What about a line with two variables?
```
      bar: prefix-@maven.property1@-@maven.property2@
```

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessor.java
rudy-regazzoni-sonarsource: Yes, it's probably better not to mess with offset. This is generally not so fun to handle, as shown with our docker preprocessor.
Seems good to me to keep it as it is for now, until we get additional values from implementing a more complete/complex solution.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessor.java
```diff
@@ -27,23 +28,23 @@ public class SpringConfigYamlPreprocessor {
...
 
-    if (Stream.of(lines).noneMatch(SpringConfigYamlPreprocessor::containsInlineCommentAfterDocumentBreak)) {
+    if (Stream.of(lines).noneMatch(containsInlineCommentAfterDocumentBreak.or(containsMavenVariableAtTokenStart))) {
       return source;
     }
 
     return Stream.of(lines)
       .map(SpringConfigYamlPreprocessor::removeInlineCommentAfterDocumentBreak)
+      .map(SpringConfigYamlPreprocessor::transformMavenSubstitutions)
       .collect(Collectors.joining("\n"));
```
rudy-regazzoni-sonarsource, line 45: We lose the original line separator.
This should not be an issue, although I still want to raise awareness about it (this was an issue for docker, where the preprocessor was changing the offset, changing a `\r\n` into `\n` would have higher impact).

rudy-regazzoni-sonarsource (approved):
LGTM! I added a few comments, maybe the multiple variable thing would require some modification if it's a valid concern. It can then be tackled in a different ticket/PR, I let you take the decision on it.

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessorTest.java
petertrr: Good catch, I'll add a test case. The idea is to only substitute the first variable in a line, because when `@` is in the middle of a value, snakeyaml doesn't have a problem with it.

iac-extensions/spring-config/src/main/java/org/sonar/iac/springconfig/parser/yaml/SpringConfigYamlPreprocessor.java
petertrr: Good catch! I added a quick test case for it as well to this PR.

PR #1323, opened 2024-05-16
Cleanup properties grammar, removing dead rule in propertiesFile

mstachniuk:
Remove dead parsing rule for propertiesFile as WHITESPACE in DEFAULT_MODE is in channel(HIDDEN)

### DIFF
--- a/iac-extensions/spring-config/src/main/antlr/PropertiesParser.g4
+++ b/iac-extensions/spring-config/src/main/antlr/PropertiesParser.g4
@@ -24,7 +24,7 @@ options {
 }
 
 propertiesFile
-    : WHITESPACE? row* EOF
+    : row* EOF
     ;
 
 row
--- a/iac-extensions/spring-config/src/main/antlr/PropertiesLexer.g4
+++ b/iac-extensions/spring-config/src/main/antlr/PropertiesLexer.g4
@@ -21,7 +21,7 @@ lexer grammar PropertiesLexer;
 
 // Adds support for CR as line terminator into the generated Java code
 @members {
-public PropertiesLexer(CharStream input, boolean crLexerCostructor) {
+public PropertiesLexer(CharStream input, boolean crLexerConstructor) {
   super(input);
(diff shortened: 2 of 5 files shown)


iac-extensions/spring-config/src/main/antlr/PropertiesParser.g4
```diff
@@ -24,7 +24,7 @@ options {
 }
 
 propertiesFile
-    : WHITESPACE? row* EOF
+    : row* EOF
```
petertrr, line 27: So, antlr can match a rule from the current mode even if it is not explicitly mentioned in the grammar? And everything still works, because spacing is sent into the hidden channel?

petertrr (approved):
LGTM, good catch!

iac-extensions/spring-config/src/main/antlr/PropertiesParser.g4
mstachniuk:
```
WHITESPACE      : [ \t\f\r\n\u2028\u2029]+ -> channel(HIDDEN);
```

This means that in `DEFAULT_MODE`, any detected whitespace will be sent to the hidden channel. Thus, the parser that consumes tokens (from the "default channel") will never see any leading whitespace, and the part `WHITESPACE?` in the `propertiesFile` rule will never be matched.

In other modes, we also produce whitespaces, e.g., `KEY_TERM  : [\r\n\u2028\u2029]+  -> type(WHITESPACE), popMode;` but those will not be in the hidden channel but in "default channel" (steam of tokens). Thanks to that we can use it later in `eol : WHITESPACE+ | EOF;` rule.

Trying to answer your questions:
1. I'm not sure how to answer, but the parser reads just a stream of tokens and doesn't know/care in what mode those tokens were produced.
2. Everything works because only this `WHITESPACE`ing from `DEFAULT_MODE` (effectively only leading spacing) is sent to the HIDDEN channel. This `WHITESPACE?` in `propertiesFile` rule was just dead code.

PR #1325, opened 2024-05-17
SONARIAC-1141 S6587 should not raise on apt-get when installing a local package

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PackageInstallationCacheCheck.java
```diff
@@ -51,12 +51,14 @@ public class PackageInstallationCacheCheck implements IacCheck {
   private static final Predicate<String> isFlag = s -> s.startsWith("-");
   private static final Predicate<String> flagWithNecessaryRmOptions = isFlag.and(containsROrF);
   private static final Predicate<String> flagWithoutContainingNecessaryFlags = isFlag.and(not(flagWithNecessaryRmOptions));
+  private static final Predicate<String> packageName = s -> !s.contains("/") && !s.endsWith(".deb");
```
petertrr, line 54: Since we also support `apk`, `.apk` can also be added here as stated in the [docs](https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper#Add_a_local_Package)

iac-extensions/docker/src/test/resources/checks/PackageInstallationCacheCheck/mixed.dockerfile
```diff
@@ -42,6 +42,17 @@ RUN apt-get install nginx && aptitude clean
 RUN aptitude install nginx && apt clean
 RUN aptitude install nginx && apt-get clean
 
+# Compliant - direct package installation
+RUN apk add nginx.deb
+RUN apt-get install wget.deb
+RUN apk add wget.deb
+RUN apk add ./nginx.other
+RUN apt-get install ./wget.other
```
petertrr, line 50: Will `apt-get install wget.other` result in an issue? Generally speaking, it's not a valid name for a Debian package ([link](https://askubuntu.com/questions/1262528/what-are-usable-characters-are-for-package-names)) but maybe it can be tested for completeness

petertrr (approved):

PR #1326, opened 2024-05-17
SONARIAC-1260 S6596 should not raise an issue on docker special image scratch

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpecificVersionTagCheck.java
```diff
@@ -54,6 +54,10 @@ private static void checkFromInstruction(CheckContext ctx, FromInstruction fromI
     }
     String fullImageName = resolvedImage.value();
 
+    if ("scratch".equals(fullImageName)) {
```
petertrr, line 57: I see that we already have the same check in `PrivilegedUserCheck.isScratchImage`. Maybe you could extract it into `CheckUtils`?

petertrr (approved):
LGTM!

PR #1330, opened 2024-05-21
SONARIAC-1467 S6380 should not raise on `storageAccounts` when `allowBlobPublicAccess` is not set

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
```diff
@@ -130,4 +132,13 @@ public static Double asNumericValueOrNull(ArmTree expr) {
     }
     return null;
   }
+
+  public static Predicate<Expression> isVersionGreaterOrEqualThan(String version) {
+    // Usually, ARM versions are in the format "YYYY-MM-DD", so comparing them lexicographically yields the correct result.
+    // They can also have a suffix like `-preview`; for these versions we assume they are not greater than the same version without the suffix.
```
petertrr, line 138: I didn't find a dedicated piece of docs stating this, but I didn't see any examples with other formats either

rudy-regazzoni-sonarsource: I would rename it to `isVersionOlderOrEqualThan`.
Also, we could imagine that instead of having a single string argument, we could have a `Date`/`Time` object.
This prevents requiring to check what format to use and facing incorrect/strange behavior otherwise.

rudy-regazzoni-sonarsource (approved):
LGTM! Please see my comment.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/utils/CheckUtils.java
petertrr: IMO, comparing versions as `Date`s wouldn't improve things much, because the format is fixed and simple, and there can be suffixes don't represent a date.

PR #1331, opened 2024-05-21
SONARIAC-1418 S6596 should not raise on references to previous build stages when previous stage is unresolvable

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/test/resources/checks/SpecificVersionTagCheck/specificVersionTagCheck.dockerfile
```diff
@@ -79,3 +79,7 @@ foobar:1.2.3
 FROM :
 FROM :bar
 FROM foo:
+
+
+FROM ${UNRESOLVED} as myBase
```
mstachniuk, line 84: Please add a comment why it is compliant

iac-extensions/docker/src/test/resources/checks/SpecificVersionTagCheck/specificVersionTagCheck.dockerfile
```diff
@@ -79,3 +79,7 @@ foobar:1.2.3
...
+FROM myBase
```
mstachniuk, line 85: Please also validate a case like `FROM base as base` - see a ticket description.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SpecificVersionTagCheck.java
```diff
@@ -48,6 +48,11 @@ private static void checkBody(CheckContext ctx, Body body) {
   }
 
   private static void checkFromInstruction(CheckContext ctx, FromInstruction fromInstruction, Set<String> encounteredAlias) {
+    var alias = fromInstruction.alias();
+    if (alias != null) {
+      encounteredAlias.add(alias.alias().value());
+    }
```
mstachniuk, line 54: Please double-check if the implementation is correct because the ticket description says: `But, the addition of the alias should not be done at the top of this method, as we don’t want to miss code like: `.
Please also add more tests cases to be on the safe side.

mstachniuk (approved):

PR #1333, opened 2024-05-21
SONARIAC-1468 S1192 should not raise on module path

petertrr:


iac-extensions/arm/src/test/resources/checks/StringLiteralDuplicatedCheck/StringLiteralDuplicatedCheck.bicep
```diff
@@ -54,3 +54,40 @@ resource storageAccount 'Microsoft.Storage/storageAccounts' = {
   }
 }
 
+// Compliant: module paths cannot be variables
+module moduleName_one '../dosomething.bicep' = {
```
petertrr, line 58: This is not exactly the example from the ticket (added `../`), because for that example no issue was being raised. The user might've simplified the example without checking if it still raises, because an exclusion for `dosomething.bicep` has been in this rule from the very beginning.

iac-extensions/arm/src/test/resources/checks/StringLiteralDuplicatedCheck/StringLiteralDuplicatedCheck.bicep
```diff
@@ -54,3 +54,40 @@ resource storageAccount 'Microsoft.Storage/storageAccounts' = {
...
+    argument: 'test4'
+  }
+}
+module moduleName_five '../dosomething.bicep' = {
+  name: 'deployName_five'
+  params: {
+    argument: 'test5'
+  }
+}
+module moduleName_five '../dosomething.bicep' = {
```
mstachniuk, line 88: Please add a separate example for `dosomething.bicep ` and `dir/something.bicep`

mstachniuk (approved):
LGTM and please see my small comment

PR #1334, opened 2024-05-21
SONARIAC-1473 Should not throw parsing error for empty key

mstachniuk:


iac-extensions/spring-config/src/main/antlr/PropertiesParser.g4
```diff
@@ -45,6 +46,11 @@ eol
     | EOF
     ;
 
+// The lines that starts from delimiter doesn't make much sense but Spring accept such entries without any error
+invalidLine
+    : DELIMITER key? eol
```
petertrr, line 51:
```suggestion
    : DELIMITER value = key? eol
```

iac-extensions/spring-config/src/main/antlr/PropertiesParser.g4
```diff
@@ -45,6 +46,11 @@ eol
     | EOF
     ;
 
+// The lines that starts from delimiter doesn't make much sense but Spring accept such entries without any error
```
petertrr, line 49: I suggest to also mention that this is a separate parser rule because we don't want to insert such lines into the AST in the visitor from the regular `visitLine` method (at least this is how I understand it)

iac-extensions/spring-config/src/test/java/org/sonar/iac/springconfig/parser/properties/PropertiesParserBaseVisitorTest.java
```diff
@@ -1026,14 +1026,73 @@ void shouldVerifyTextRangesWhenCR() {
...
+      "visitKey bar2",
+      "visitEol <EOF>");
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"=", ":"})
+  void shouldParseLineContainingDelimiterBetweenOtherKeys(String delimiter) {
+    var code = """
+      foo=bar
+      %sabc
```
petertrr, line 1074: Maybe also add a test with both key and value missing, so just with a delimiter on a line?

petertrr (approved):
LGTM! Please see my suggestions

PR #1338, opened 2024-05-21
SONARIAC-1030 S6330 Should consider correct default queue encryption (SSE-SQS)

petertrr:


petertrr:
Relevant RSPEC PR: https://github.com/SonarSource/rspec/pull/2626 (also touches other languages so probably should be split before we release)

iac-extensions/cloudformation/src/test/java/org/sonar/iac/cloudformation/checks/UnencryptedSqsQueueCheckTest.java
```diff
@@ -28,14 +28,14 @@ class UnencryptedSqsQueueCheckTest {
 
   @Test
   void test_yaml() {
-    CloudformationVerifier.verify("UnencryptedSqsQueueCheck/test.yaml", new UnencryptedSqsQueueCheck());
+    CloudformationVerifier.verify("UnencryptedSqsQueueCheck/UnencryptedSqsQueueCheck.yaml", new UnencryptedSqsQueueCheck());
   }
 
   @Test
   void test_json() {
```
rudy-regazzoni-sonarsource, line 35: Nitpick: take the opportunity to rename test methods with camelCase.

iac-extensions/cloudformation/src/test/resources/checks/UnencryptedSqsQueueCheck/UnencryptedSqsQueueCheck.yaml
```diff
@@ -0,0 +1,36 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Resources:
+  UnencryptedQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+     "DisplayName": "unencrypted_queue"
```
rudy-regazzoni-sonarsource, line 6:
```suggestion
     DisplayName: "unencrypted_queue"
```
Nitpick, unless it was done on purpose?

iac-extensions/terraform/src/test/resources/checks/UnencryptedSqsQueueCheck/UnencryptedSqsQueueCheck.tf
rudy-regazzoni-sonarsource, line 1: For documentation purpose, you could add a test where `sqs_managed_sse_enabled` is false but `kms_master_key_id` is set.
Currently I'm guessing it would raise an issue, from the ticket description I'm not sure if it should, maybe this is something you want to check with the ticket reporter?

rudy-regazzoni-sonarsource (approved):
LGTM! Took me some time to understand the ticket description, to me it seems overcomplicated just to say `raise if SqsManagedSseEnabled = false`.
I let you check the logic with the ticket reporter if you have a doubt on it.
Please also see my few comments.

iac-extensions/cloudformation/src/test/resources/checks/UnencryptedSqsQueueCheck/UnencryptedSqsQueueCheck.yaml
petertrr: Well, it was like that before, but YAML shouldn't pay attention to quotes on a key.

iac-extensions/terraform/src/test/resources/checks/UnencryptedSqsQueueCheck/UnencryptedSqsQueueCheck.tf
petertrr: Good catch! I found in the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sqs-queue.html) that

> Only one server-side encryption option is supported per queue (for example, SSE-KMS or SSE-SQS).

I'll update the logic accordingly

PR #1342, opened 2024-05-22
SONARIAC-1392 Should not throw parse exception strconv.Atoi when read _resources.tpl

mstachniuk:
First commit: Refactoring. Moving some functions between files as a starting point for rewrite.

### DIFF
--- a/sonar-helm-for-iac/src/converters/io-handling.go
+++ b/sonar-helm-for-iac/src/converters/io-handling.go
@@ -21,6 +21,7 @@ package converters
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"github.com/samber/mo"
 	"io"
@@ -45,6 +46,8 @@ type InputReader interface {
 	// END
 	ReadInput(scanner *bufio.Scanner) (string, Files, error)
 }
+
+var stdinReader InputReader = StdinReader{}
 
 type StdinReader struct{}
 
@@ -150,4 +153,28 @@ func CreateScanner(reader io.Reader) *bufio.Scanner {
 	buf := make([]byte, INPUT_BUFFER_SIZE)
 	scanner.Buffer(buf, INPUT_BUFFER_SIZE)
 	return scanner
+}
+
+func ReadAndValidateSources() (*TemplateSources, error) {
+	scanner := CreateScanner(os.Stdin)
+	templateName, sources, err := stdinReader.ReadInput(scanner)
+	if err != nil {
+		return nil, fmt.Errorf("error reading content: %w", err)
+	}
+	if err = validateInput(sources); err != nil {
+		return nil, fmt.Errorf("error validating content: %w", err)
+	}
+
+	return NewTemplateSourcesFromRawSources(templateName, sources), nil
+}
+
+func validateInput(sources Files) error {
+	if len(sources) == 0 {
+		return errors.New("no input received")
+	}
+	return nil
+}
+
+func NewTemplateSourcesFromRawSources(templateName string, rawSources Files) *TemplateSources {
+	return NewTemplateSources(templateName, rawSources)
 }
--- a/sonar-helm-for-iac/src/main_test.go
+++ b/sonar-helm-for-iac/src/main_test.go
@@ -19,7 +19,6 @@
 package main
 
 import (
-	"bufio"
 	"errors"
 	"github.com/SonarSource/sonar-iac/sonar-helm-for-iac/src/converters"
 	pbstructs "github.com/SonarSource/sonar-iac/sonar-helm-for-iac/src/org.sonar.iac.helm"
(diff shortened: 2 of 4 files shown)


sonar-helm-for-iac/src/converters/io-handling.go
```diff
@@ -46,8 +48,77 @@ type InputReader interface {
...
+
+func readByteAsInt(input *os.File) (int, error) {
+	numberBytes := make([]byte, 4)
+	_, err := input.Read(numberBytes)
+	if err != nil {
+		//TODO better error handling
+		fmt.Fprintf(os.Stderr, "ReadInput2 error 1")
+		return 0, err
+	}
+	number := (int(numberBytes[0]))*int(math.Pow(2, 24)) +
```
petertrr, line 112: It looks like Go has the `>>` operator, so this can be rewritten

sonar-helm-for-iac/src/main_test.go
```diff
@@ -76,29 +51,14 @@ func Test_exit_code_with_serialization_error(t *testing.T) {
 	stdin, _ := cmd.StdinPipe()
 	defer stdin.Close()
 	cmd.Start()
-	stdin.Write([]byte("foo.yaml\n1\napiVersion: v1\nvalues.yaml\n0\n\nEND\n"))
+	stdin.Write([]byte("\x00\x00\x00\x08foo.yaml\x00\x00\x00\x0EapiVersion: v1\x00\x00\x00\x01\x00\x00\x00\x0Bvalues.yaml\x00\x00\x00\x00"))
 	err := cmd.Wait()
```
petertrr, line 55: Is the trailing 0 something for this test only or not? I didn't see it neither on the Java side nor in input handling. Or did I miss something?

mstachniuk: The trailing 0 (4 bytes) is the content size of values.yaml

iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/HelmEvaluator.java
```diff
@@ -96,29 +96,30 @@ Process startProcess() throws IOException {
...
+    var contentBytes = content.getBytes(StandardCharsets.UTF_8);
+    out.write(intTo4Bytes(contentBytes.length));
+    out.write(contentBytes);
   }
 
-  private static void writeStringAsBytes(OutputStream out, String content) throws IOException {
-    out.write(content.getBytes(StandardCharsets.UTF_8));
+  private static byte[] intTo4Bytes(int number) {
+    var array = new byte[4];
+    array[0] = (byte) (Math.floor(number / (Math.pow(2, 24))));
```
jonas-wielage-sonarsource, line 118: Can't the `Math.pow(2, x)` be calcuated beforehand for everycall? Not sure the JVM does this optimization.

sonar-helm-for-iac/src/converters/io-handling.go
```diff
@@ -19,135 +19,134 @@
...
+func readBytes(input *os.File, contentLength int) ([]byte, int, error) {
+	contentBytes := make([]byte, contentLength)
+	numberOfBytesRead, err := input.Read(contentBytes)
+	if err != nil {
+		message := fmt.Sprintf("Error reading from stdin, expecting to read %d bytes, but got %d", contentLength, numberOfBytesRead)
+		return nil, numberOfBytesRead, errors.New(message)
+	}
+	sum := numberOfBytesRead
+	if numberOfBytesRead != contentLength {
+		bytes, numberOfBytesReadInRecursion, err := readBytes(input, contentLength-numberOfBytesRead)
```
petertrr, line 100: Are you sure 2 attempts are enough? Maybe, it would be more reliable to read in a loop while the desired number of bytes is read? This can maybe depend on the buffer size on the OS level when two processes are communicating.

sonar-helm-for-iac/src/converters/io-handling.go
```diff
@@ -19,135 +19,134 @@
...
+	return contentBytes, sum, nil
 }
 
-// By default bufio.Scanner scans input line by line using \n (LF) as separator, trims \r (CR)
-// and doesn't include new line separator in token.
-// This fuction split tokens using \r\n or \r or \n or \u2028 or \u2029 and include them in tokens.
-func ScanLinesIncludeNewLine(data []byte, atEOF bool) (advance int, token []byte, err error) {
-	if atEOF && len(data) == 0 {
-		return 0, nil, nil
+func readByteAsInt(input *os.File) (int, error) {
```
petertrr, line 111:
```suggestion
func readBytesAsInt(input *os.File) (int, error) {
```

sonar-helm-for-iac/src/converters/io-handling.go
mstachniuk: It is a recursion call so it will be done until it is needed.

PR #1344, opened 2024-05-22
SONARIAC-1322 Empty file suffixes are not substituted with defaults with SQ 10.4

rudy-regazzoni-sonarsource:


its/plugin/src/integrationTest/java/org/sonarsource/iac/PropertiesTest.java
```diff
@@ -142,13 +143,9 @@ private void executeBuildAndAssertMetric(
     String projectKey, String language, String subDir,
     String property, String propertyValue,
     @Nullable String profileName,
-    int expectedFiles, int expectedNcloc) {
+    Integer expectedFiles, Integer expectedNcloc) {
```
petertrr, line 146: Do I understand correctly that this is because they will be captured as `null`s for empty values? Could you add the `@Nullable` annotation in this case?

petertrr (approved):
LGTM!

its/plugin/src/integrationTest/java/org/sonarsource/iac/PropertiesTest.java
rudy-regazzoni-sonarsource: Yes you are correct, and that's a good suggestion! I will add it.

PR #1347, opened 2024-05-24
SONARIAC-1485 Docker parser should not create invalid offset on multiline bash script

petertrr:


its/ruling/src/integrationTest/resources/sources/docker/parsingError/SONARIAC-1485.dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM debian:stretch-slim
+
+ENV BDBVER db-4.8.30.NC
+RUN wget --quiet https://download.oracle.com/berkeley-db/$BDBVER.tar.gz; \
+        echo "Done"
```
jonas-wielage-sonarsource, line 5: Can this be also added to the Unit Test for the check? There we can verify the range the issue raised on in more detail.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/ArgumentResolutionSplitter.java
```diff
@@ -107,7 +107,16 @@ private static void parseRemainderOfTheCommand(
 
   private static ArgumentResolution buildSubArgument(ArgumentResolution resolvedArgument, String firstCommand, int offsetShift) {
     var argumentRange = resolvedArgument.argument().textRange();
-    SyntaxToken token = new SyntaxTokenImpl(firstCommand, range(argumentRange.start().line(), argumentRange.start().lineOffset() + offsetShift, firstCommand),
+
+    // This is a workaround for the cases when after variable resolution offset becomes too big. See SONARIAC-1485.
+    // At this point, we can't reliably backtrack which parts of the argument value were resolved from variables.
+    // To highlight only a part of a long command, we use the length of the first command, if available, falling back to highlighting
+    // everything.
+    int endOffset = Math.min(argumentRange.start().lineOffset() + firstCommand.length(), argumentRange.end().lineOffset()) + offsetShift;
```
petertrr, line 115: It's not possible to always use `argumentRange.end()`, because `resolvedArgument` can contain multiple commands, e.g. `firstCommand == "command1"`, `resolvedArgument.value() == "command1;command2"`. In these cases using the original range would lead to incorrect behavior (`command1` having the range of the entire input string), so I'm adding it as a fallback in case checking `firstCommand.length()` would lead to an error

rudy-regazzoni-sonarsource (approved):
LGTM! Nice workaround, this should prevent failure and should be enough for now, good job!

PR #1352, opened 2024-05-27
Change sca_scan condition to only run on main branch.

rudy-regazzoni-sonarsource:
Funnily, we can see in th `STARLARK.md` file in **Task definitions** chapter that the example code provided is the one about the definition of `sca_scan_task` and has the property `"only_if": is_main_branch()`.
Was it changed at some point for some reason?

### DIFF
--- a/.cirrus/modules/build.star
+++ b/.cirrus/modules/build.star
@@ -126,7 +126,7 @@ def whitesource_script():
 def sca_scan_task():
     return {
         "sca_scan_task": {
-            "only_if": is_branch_qa_eligible(),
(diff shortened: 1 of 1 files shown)


olivier-schmitt-sonarsource (approved):
LGTM!
Indeed the condition should be:

  only_if: $CIRRUS_USER_COLLABORATOR == 'true' && $CIRRUS_TAG == "" &&
    ($CIRRUS_BRANCH == $CIRRUS_DEFAULT_BRANCH || $CIRRUS_BRANCH =~ "branch-.*")

PR #1357, opened 2024-06-12
Fix authentication error on windows integration test 

hedinasr:
The windows base image use repox as a choco repository mirror. We need to authenticate to repox in order to be able to fetch packages.

### DIFF
--- a/.cirrus/modules/env.star
+++ b/.cirrus/modules/env.star
@@ -11,7 +11,6 @@ def go_env():
     Provides typical environment variables to work with Go.
     The following default values are provided:
     - GO_VERSION: 1.21.8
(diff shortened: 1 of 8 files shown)


.cirrus/modules/env.star
```diff
@@ -10,7 +10,8 @@ def go_env():
     """
     Provides typical environment variables to work with Go.
     The following default values are provided:
-    - GO_VERSION: 1.21.7
+    - GO_VERSION: 1.21.8
+    - GO_ZIP_CHECKSUM: 872ac1c6ba1e23927a5cd60ce2e7a9e64cc6e5a550334c0fbcc785b4347d5f0d
```
petertrr, line 14: I would advise against putting it here, because it's a checksum only of a Windows archive, and this dictionary is used for all platforms

petertrr (approved):
LGTM, please see my comment

PR #1363, opened 2024-06-17
SONARIAC-1507 Convert AST resources to "ProjectResource" classes

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/model/ProjectResource.java
```diff
@@ -0,0 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.kubernetes.model;
```
petertrr, line 20: Not sure if it should be placed somewhere deeper in the hierarchy

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/model/LimitRange.java
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.kubernetes.model;
+
+import java.util.List;
+
+public record LimitRange(
+  List<LimitRangeItem> limits) implements ProjectResource {
```
petertrr, line 25: An example of LimitRange with multiple limits can be found e.g. [here](https://github.com/kubernetes-sigs/cluster-capacity/blob/42fefeb1a4fb5fdd490d2b7e00b1b99bed04676f/examples/limits.yml#L4)

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/model/ProjectResource.java
mstachniuk: IMO it is ok to place it here. I'm not super happy with the `model` package name as it is too general world. In sonar-php the `symbol` package is used for a similar purpose, but we don't have symbols here.
Maybe `project` or `shared` or `xfile` or `crossfile` will be a better name here? (I don't have a strong opinion here).

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectResourceFactory.java
```diff
@@ -0,0 +1,108 @@
...
+      .map(ScalarTree.class::cast);
+    var automountServiceAccountToken = automountServiceAccountTokenTree
+      .map(TextUtils::trileanFromTextTree)
+      .orElse(Trilean.UNKNOWN);
+    var name = PropertyUtils.value(tree, "metadata", MappingTree.class)
+      .flatMap(metadata -> PropertyUtils.value(metadata, "name"))
+      .map(ScalarTree.class::cast)
+      .map(ScalarTree::value);
+
+    return name.map(s -> new ServiceAccount(s, automountServiceAccountToken, automountServiceAccountTokenTree.map(HasTextRange::textRange).orElse(null)))
```
mstachniuk, line 60: What if `ServiceAccount` doesn't have a defined `name`? I guess it is useless as another resource can't reference them. 🤔

mstachniuk (approved):

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectResourceFactory.java
petertrr: Yes, then the method would return `null`, and it should be filtered out by the caller.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/model/ProjectResource.java
petertrr: I'll stick with the `model` then

PR #1367, opened 2024-06-18
Add Github Action releasability-check

jonas-wielage-sonarsource:


.github/workflows/releasability-check.yml
```diff
@@ -0,0 +1,45 @@
...
+        env:
+          ARTIFACTORY_PRIVATE_USERNAME: vault-${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ROLE }}
+          ARTIFACTORY_PRIVATE_PASSWORD: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_PASSWORD }}
+          ARTIFACTORY_URL: ${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_URL }}
+          REPO: sonarsource-public-builds
+        run: |
+          echo LATEST_VERSION=$(curl -s -u ${ARTIFACTORY_PRIVATE_USERNAME}:${ARTIFACTORY_PRIVATE_PASSWORD} \
+          "${ARTIFACTORY_URL}/api/search/latestVersion?g=org.sonarsource.iac&a=sonar-iac-plugin&repos=${REPO}") >> "$GITHUB_OUTPUT"
+
+      - uses: SonarSource/gh-action_releasability@v1.0.3
```
petertrr, line 37: There is already a `1.0.4` version

.github/workflows/releasability-check.yml
```diff
@@ -0,0 +1,45 @@
...
+        id: releasability
+        with:
+          branch: ${{ github.ref_name }}
```
petertrr, line 40: ~I think the branch should either be configurable (e.g. for releases from `master` or from `branch-*`) or hardcoded to `master`~
As discussed, we can use `Run workflow from branch` instead of an input property, than this should work. Maybe, please add a comment here about the `LATEST_VERSION` and how to choose a branch

petertrr (approved):
LGTM and please see my comment

PR #1370, opened 2024-06-18
SONARIAC-1500 Create a "ProjectContextEnricherVisitor" and store "ProjectResource" per namespace and path

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContextEnricherVisitor.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class ProjectContextEnricherVisitor extends TreeVisitor<InputFileContext> {
+  private static final String DEFAULT_NAMESPACE = "default";
+
+  public ProjectContextEnricherVisitor(ProjectContext.Builder projectContextBuilder) {
+    register(FileTree.class, (ctx, fileTree) -> handleFileTree(ctx, fileTree, projectContextBuilder));
+  }
+
+  private static void handleFileTree(InputFileContext ctx, FileTree fileTree, ProjectContext.Builder projectContextBuilder) {
+    var path = ctx.inputFile.uri().toString();
```
jonas-wielage-sonarsource, line 37: I think we can also use `ctx.inputFile.toString()`, it will give us the relative path to the root.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectResourceFactory.java
```diff
@@ -41,6 +41,19 @@ private ProjectResourceFactory() {
...
+      .orElse("");
+    return switch (kind) {
+      case "ServiceAccount" -> createServiceAccount(tree);
+      case "LimitRange" -> createLimitRange(tree);
+      default -> null;
+    };
+  }
+
   @CheckForNull
   public static ProjectResource createServiceAccount(MappingTree tree) {
```
jonas-wielage-sonarsource, line 58: If we access the Factory only by the `createResource()` method, we can remove the `public` modifier and drop the first condition.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
```diff
@@ -19,14 +19,35 @@
...
 public final class ProjectContext {
+  private final Map<String, Map<String, Set<ProjectResource>>> projectResourcePerNamespacePerPath = new HashMap<>();
 
   private ProjectContext() {
   }
 
+  public Set<ProjectResource> getProjectResource(String namespace, String path, Class<? extends ProjectResource> clazz) {
+    if (projectResourcePerNamespacePerPath.containsKey(namespace)) {
+      var resourcesPerPath = projectResourcePerNamespacePerPath.get(namespace);
+      if (resourcesPerPath.containsKey(path)) {
```
jonas-wielage-sonarsource, line 41: I'm not sure about this part. I remember that we should add some logic to traverse the pathes, not only checking for equals.

jonas-wielage-sonarsource: After checking with Rudy, there's still some things to do here. Will write via slack.

jonas-wielage-sonarsource (changes requested):

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContextEnricherVisitor.java
petertrr: Looking at the implementation in SonarLint:
```java
  public String toString() {
    return "[uri=" + this.uri() + "]";
  }
```
I'm not sure if it will be reliable

jonas-wielage-sonarsource: I remembered that we're using it in sonarText. I created a ticket there to adjust it as well.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
```diff
@@ -19,24 +19,72 @@
...
 /**
  * Data class to provide information about the project. This allows to share cross-file knowledge to the individual checks.
  */
 public final class ProjectContext {
+  private final Map<String, Map<String, Set<ProjectResource>>> projectResourcePerNamespacePerPath = new HashMap<>();
+  private final FileSystem fileSystem;
 
-  private ProjectContext() {
+  private ProjectContext(FileSystem fileSystem) {
+    this.fileSystem = fileSystem;
```
petertrr, line 43: This is needed to acces HelmFileSystem later and to check if the file is inside a Helm project. Alternatively, could be done by providing an `InputFileContext` instead of `path` and retrieving this information from there

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContextEnricherVisitor.java
petertrr: Dealing with URIs everywhere makes the code a bit more complicated, however, I'm still under the impression that this should be the proper strategy to support all runtimes

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/visitors/ProjectContextTest.java
```diff
@@ -19,8 +19,110 @@
...
+    builder.addResource("default", toUri("path1/templates/resource.yaml").toString(), resource1);
+    var resource2 = mock(TestResource.class);
+    builder.addResource("default", toUri("path1/Chart.yaml").toString(), resource2);
+    var resource3 = mock(TestResource.class);
+    builder.addResource("default", toUri("path2/resource.yaml").toString(), resource3);
+    var ctx = builder.build();
+
+    try (var ignored = Mockito.mockStatic(HelmFileSystem.class)) {
+      when(HelmFileSystem.retrieveHelmProjectFolder(
+        argThat(path -> path.toString().contains("path1")),
```
jonas-wielage-sonarsource, line 107: Didn't know the `argThat` before, I will definitely use it in the future.

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/visitors/ProjectContextTest.java
```diff
@@ -19,8 +19,110 @@
...
+
+    var resource1 = mock(TestResource.class);
+    builder.addResource("default", toUri("path1/templates/resource.yaml").toString(), resource1);
+    var resource2 = mock(TestResource.class);
+    builder.addResource("default", toUri("path1/Chart.yaml").toString(), resource2);
+    var resource3 = mock(TestResource.class);
+    builder.addResource("default", toUri("path2/resource.yaml").toString(), resource3);
+    var ctx = builder.build();
+
+    try (var ignored = Mockito.mockStatic(HelmFileSystem.class)) {
```
jonas-wielage-sonarsource, line 105: In theory you could also build the input files and put them into the filesystem so you don't have to mock them.
But that would go more in the direction of an integration test, so I'm fine with statically mocking here.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
```diff
@@ -19,24 +19,72 @@
...
+  /**
+   * Get all resources of a given {@code clazz} in a given {@code namespace} and that are accessible to a file with the given {@code path}.
+   * This means that the resources can be in the same file, or in the same directory, or in the descendant directories, but not in the ancestor directories.
+   */
+  public Set<ProjectResource> getProjectResources(String namespace, String path, Class<? extends ProjectResource> clazz) {
+    if (projectResourcePerNamespacePerPath.containsKey(namespace)) {
+      var resourcesPerPath = projectResourcePerNamespacePerPath.get(namespace);
+
+      var filePath = Path.of(path);
+      var basePath = Optional.ofNullable(retrieveHelmProjectFolder(filePath, fileSystem))
```
jonas-wielage-sonarsource, line 55: I'm not a fan of calling `retrieveHelmProjectFolder` for every file.
In my opinion this information should be present in the Context.
If it's a `HelmInputFileContext` we should store the HelmProjectFolder once during the construction of the Context and retrieve it from there (there are multiple calls to retrieve the Folder during the whole analysis).
And if it's not a `HelmInputFileContext` we should directly go to `filePath.getParent()`

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
```diff
@@ -19,24 +19,72 @@
...
+        .or(() -> Optional.ofNullable(filePath.getParent()))
+        .map(Path::normalize)
+        .map(Path::toUri)
+        .map(URI::toString)
+        .orElse("");
+
+      return resourcesPerPath.entrySet().stream()
+        .filter(entry -> basePath.isEmpty() || entry.getKey().startsWith(basePath))
```
jonas-wielage-sonarsource, line 63: In which case is it `basePath.isEmpty()` true? Can you add a test case?

petertrr: Only if the file is located in the FS root. But then, basePath can be set to `/`. I'll try to add a test case

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
petertrr: Yes, this was another idea I had. Then I'll go this way and change the HelmInputFileContext

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/plugin/KubernetesAnalyzerTest.java
```diff
@@ -483,6 +485,18 @@ void shouldSkipProcessingWhenInputFileIsInvalid() {
     assertThat(logTester.logs()).doesNotContain("Helm content detected in file _helpers.tpl");
   }
 
+  @Test
+  void shouldSetHelmProjectDirectory() {
+    try (var ignored = mockStatic(HelmFileSystem.class)) {
+      when(HelmFileSystem.retrieveHelmProjectFolder(any(), any())).thenReturn(Path.of("/chart"));
+
+      var ifc = analyzer.createInputFileContext(sensorContext, inputFile);
```
jonas-wielage-sonarsource, line 493:
```suggestion
      var inputFileContext = analyzer.createInputFileContext(sensorContext, inputFile);
```

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/KubernetesAnalyzer.java
```diff
@@ -55,7 +57,12 @@ public KubernetesAnalyzer(String repositoryKey, TreeParser<? extends Tree> parse
 
   @Override
   protected InputFileContext createInputFileContext(SensorContext sensorContext, InputFile inputFile) {
-    return new HelmInputFileContext(sensorContext, inputFile);
+    var ifc = new HelmInputFileContext(sensorContext, inputFile);
```
jonas-wielage-sonarsource, line 60:
```suggestion
    var inputFileContext = new HelmInputFileContext(sensorContext, inputFile);
```

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/HelmInputFileContext.java
```diff
@@ -81,11 +82,18 @@ public String getSourceWithComments() {
...
     return linesShifting;
   }
+
+  public Path getHelmProjectDirectory() {
+    return helmProjectDirectory;
+  }
+
+  public void setHelmProjectDirectory(Path helmProjectDirectory) {
+    this.helmProjectDirectory = helmProjectDirectory;
+  }
```
jonas-wielage-sonarsource, line 98: I'm unsure if we should set this with a setter, or:

- as a constructor argument
- calculated in the constructor as we have all the information needed

I previously had a strong opinion to calculating it in the constructor. But seeing how other fields of this class are handled I'm unsure now.
WDYT?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
```diff
@@ -19,26 +19,71 @@
...
+  private final Map<String, Map<String, Set<ProjectResource>>> projectResourcePerNamespacePerPath = new HashMap<>();
 
   private ProjectContext() {
   }
 
+  /**
+   * Get all resources of a given {@code clazz} in a given {@code namespace} and that are accessible to a file with the given {@code path}.
+   * This means that the resources can be in the same file, or in the same directory, or in the descendant directories, but not in the ancestor directories.
+   */
+  public Set<ProjectResource> getProjectResources(String namespace, InputFileContext ctx, Class<? extends ProjectResource> clazz) {
```
jonas-wielage-sonarsource, line 46: Please try to use no abbreviations.
For `InputFileContext` I see `ifc` and `ctx` in this PR.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/ProjectContext.java
```diff
@@ -19,26 +19,71 @@
...
 public final class ProjectContext {
+  private final Map<String, Map<String, Set<ProjectResource>>> projectResourcePerNamespacePerPath = new HashMap<>();
 
   private ProjectContext() {
   }
 
+  /**
+   * Get all resources of a given {@code clazz} in a given {@code namespace} and that are accessible to a file with the given {@code path}.
+   * This means that the resources can be in the same file, or in the same directory, or in the descendant directories, but not in the ancestor directories.
+   */
```
jonas-wielage-sonarsource, line 45:
```suggestion
   * If the file is part of a helm project, all files inside the project are accessible. The location of the Chart.yaml serves as the root directory of the project.
   */
```

jonas-wielage-sonarsource (approved):
LGTM!
Please have a look at my suggestions.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/HelmInputFileContext.java
petertrr: Yes, I agree that it makes sense to move this to the constructor. As for other fields, they are set closer to Helm evaluation. For additional files it can also makes sense to set them in the constructor, I'll create a ticket for that.

PR #1377, opened 2024-06-20
SONARIAC-1204 S6864: Should not raise with LimitRange in the same namespace setting Memory Limit

petertrr:
Related RSPEC: https://github.com/SonarSource/rspec/pull/4002

### DIFF
--- a/iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
+++ b/iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
@@ -26,5 +26,7 @@ public interface KubernetesCheckContext extends CheckContext {
 
   void setShouldReportSecondaryInValues(boolean shouldReport);
 
+  HelmInputFileContext currentCtx();
+
   ProjectContext projectContext();
 }
--- a/iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
+++ b/iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
@@ -26,7 +26,7 @@ public interface KubernetesCheckContext extends CheckContext {
 
   void setShouldReportSecondaryInValues(boolean shouldReport);
 
-  HelmInputFileContext currentCtx();
+  HelmInputFileContext inputFileContext();
 
   ProjectContext projectContext();
 }
--- a/iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/limit_range.yaml
+++ b/iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/limit_range.yaml
@@ -1,10 +1,10 @@
 apiVersion: v1
 kind: LimitRange
 metadata:
-  name: mem-limit-range
-  namespace: default-mem-example
+  name: limit-range
+  namespace: with-global-limit
 spec:
-  limits:
+    limits:
     - type: Container
       default:
         memory: 512Mi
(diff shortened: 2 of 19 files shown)


iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/memory_limit_deployment_with_global_limit.yaml
```diff
@@ -0,0 +1,50 @@
...
+  selector:
+    matchLabels:
+      app: web
+  template:
+    metadata:
+      labels:
+        app: web
+    spec:
+      containers:
+        - name: noncompliant_without_memory_limit
```
mstachniuk, line 20: Actually the names here and in following lines should be adjusted as they are now compliant

iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/limit_range.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
mstachniuk, line 1: Please also add a test to Helm Chart (it can be in separate PR)

iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/memory_limit_pod_with_global_limit.yaml
```diff
@@ -0,0 +1,46 @@
+# This file duplicates `memory_limit_pod.yaml` but adds a namespace to the pod.
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example
+  namespace: with-global-limit
+spec:
+  containers:
+    - name: noncompliant_without_memory_limit
```
mstachniuk, line 9: Actually the names here and in following lines should be adjusted as they are now compliant

iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/memory_limit_deployment_with_global_limit.yaml
```diff
@@ -0,0 +1,50 @@
+# This file duplicates `memory_limit_deployment.yaml` but adds a namespace to the deployment.
+apiVersion: apps/v1
+kind: Deployment
```
mstachniuk, line 3: Other kinds should be also tested (it can be in separate PR)

iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/limit_range_cpu.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
mstachniuk, line 1: Actually, you can put all `LimitRange`s in one file and use different namespaces. 
Other cases to verify:
* `type: Pod` 
* `type: PersistentVolumeClaim` 
* `type: with-global-limit.example` - Full qualified name - it will be FN for now
* `min` instead of `default` is defined

mstachniuk (approved):
LGTM

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
```diff
@@ -26,5 +26,7 @@ public interface KubernetesCheckContext extends CheckContext {
 
   void setShouldReportSecondaryInValues(boolean shouldReport);
 
+  HelmInputFileContext currentCtx();
```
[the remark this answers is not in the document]
petertrr: I used the same name as in `ContextAdapter` in `iac-common`. However, I agree, `inputFileContext` will be better

iac-extensions/kubernetes/src/test/resources/checks/MemoryLimitCheck/limit_range_cpu.yaml
petertrr: Yeah, I'll add more tests later. There's a bit of a problem with `verifyContent` and x-file Kubernetes/Helm, which makes it not so straightforward to write many tests with different namespaces

PR #1378, opened 2024-06-20
SONARIAC-1498 Allow sensors to defines and provides the Analyzer they want to use

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM

PR #1383, opened 2024-06-24
SONARIAC-1278 S6869: Should not raise with LimitRange in the same namespace setting CPU Limit

petertrr:


jonas-wielage-sonarsource (approved):
LGTM

PR #1384, opened 2024-06-24
Add more tests for S6864

petertrr:
Continuation of https://github.com/SonarSource/sonar-iac/pull/1377#discussion_r1651050129


iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/KubernetesVerifier.java
```diff
@@ -76,7 +77,7 @@
 public class KubernetesVerifier {
 
   private static final Logger LOG = LoggerFactory.getLogger(KubernetesVerifier.class);
-  public static final Path BASE_DIR = Paths.get("src", "test", "resources", "checks");
+  public static final Path BASE_DIR = Paths.get("build", "resources", "test", "checks");
```
petertrr, line 80: This way we can create/modify files in a k8s/helm project, e.g. substitute placeholders with different namespaces, without messing the actual sources/resources

mstachniuk: Good idea. I thought about something similar in my PR but using temp dir.
Please verify if it has an influence on the Gradle cache (because some files will be created in the `build` dir).

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/KubernetesVerifier.java
```diff
@@ -206,6 +217,15 @@ private static String retrieveContent(InputFile inputFile) {
     }
   }
 
+  private static String createFileInBaseDir(String basePath, String content) {
+    try {
+      Files.writeString(BASE_DIR.resolve(basePath).resolve("test.yaml"), content, StandardCharsets.UTF_8);
```
mstachniuk, line 222: Maybe a more unique filename should be used here? I can imagine someone may create a `test.yaml` for some check, and then it will be overwritten. What about: `temp-file-for-verify-k8s-content.yaml` or something similar?

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/MemoryLimitCheckTest.java
```diff
@@ -38,18 +39,20 @@ void testPodKind() {
...
-    KubernetesVerifier.verify("MemoryLimitCheck/memory_limit_pod.yaml", check, "MemoryLimitCheck/limit_range_cpu.yaml");
+  @ParameterizedTest
+  @ValueSource(strings = {
+    "with-type-pvc",
+    "with-resource-cpu",
+    "with-fq-type-name",
+    "with-other-limit-member",
+    "",
+  })
+  void testPodKindWithNotMatchingGlobalLimit2(String namespace) {
```
mstachniuk, line 53:
```suggestion
  void testPodKindWithNotMatchingGlobalLimit(String namespace) {
```

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/MemoryLimitCheckTest.java
```diff
@@ -38,18 +39,20 @@ void testPodKind() {
...
-
-  @Test
-  @DisplayName("Should raise issues when a limit exists in the same namespace (default) but for different resource")
-  void shouldRaiseForPodWithGlobalLimitForCpu() {
-    KubernetesVerifier.verify("MemoryLimitCheck/memory_limit_pod.yaml", check, "MemoryLimitCheck/limit_range_cpu.yaml");
+  @ParameterizedTest
+  @ValueSource(strings = {
+    "with-type-pvc",
+    "with-resource-cpu",
+    "with-fq-type-name",
```
mstachniuk, line 49: What does `fq` mean here? Maybe `with-type-full-qualified-name` or `with-type-fqn` or any other more descriptive name will be better?

mstachniuk (approved):

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/KubernetesVerifier.java
petertrr: I didn't go with the temp dir approach because it would mean additional copying, and we don't modify most part of the resources.
For the Gradle cache, it probably is affected, but we should rerun tests if inputs change anyway. Ideally, these files should be cleaned up after test execution, but I'm not sure what would be the best approach, as KubernetesVerifier is not part of the Junit's lifecycle

petertrr: I added something with a callback; it should be added to each test class that uses this functionality, but at least it's a solution

PR #1390, opened 2024-06-28
SONARIAC-1299 S6892: Should not raise with LimitRange in the same namespace setting CPU Requests

petertrr:
RSPEC PR: https://github.com/SonarSource/rspec/pull/4017


jonas-wielage-sonarsource (approved):
LGTM

PR #1400, opened 2024-07-03
Don't fail when length of a Go AST node is too large

petertrr:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/helm/tree/impl/LocationImpl.java
```diff
@@ -134,6 +134,12 @@ public TextRange toTextRange(String sourceCode) {
   }
 
   private static TextPointer toTextPointer(int position, String sourceCode) {
+    if (position > sourceCode.length()) {
+      // This can happen sometimes for weird special characters. The Go side returns a node with too great value of the `length` field.
+      // Until this is fixed on the Go side, we just ignore it.
+      position = sourceCode.length();
```
jonas-wielage-sonarsource, line 140: Do we want to add TODO comment mentioning the follow up ticket here?

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelperTest.java
```diff
@@ -92,4 +94,22 @@ void shouldFindValuesForTwoHelmExpressionsInRange() throws IOException {
...
+    var code = """
+      apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: example
+      spec:
+        data:
+         \s""" + specialSymbol;
+
+    assertThatCode(() -> new GoAstCreator(tempDir).goAstFromSource(code, "", "name: test chart"))
```
jonas-wielage-sonarsource, line 112: Did it throw when you didn't have your fix in place?
If yes good, if no I would like an example for that :D

jonas-wielage-sonarsource (approved):
LGTM, please see my comments.

iac-extensions/kubernetes/src/test/java/org/sonar/iac/helm/tree/utils/GoTemplateAstHelperTest.java
petertrr: Yes, it did :)

PR #1406, opened 2024-07-05
SONARIAC-1310 S1874: Deprecated code should not be used

jonas-wielage-sonarsource:


iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/KubernetesVerifier.java
```diff
@@ -297,6 +297,7 @@ public static class KubernetesTestContext extends Verifier.TestContext implement
     private final InputFileContext inputFileContext;
     private final ProjectContext projectContext;
     private boolean shouldReportSecondaryInValues = true;
+    private boolean disableLocationShifting;
```
petertrr, line 300: I'd make the opposite default: `enableLocationShifting = true`, because we use this for the most rules, and it can be more readable like this

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
```diff
@@ -27,6 +27,8 @@ public interface KubernetesCheckContext extends CheckContext {
 
   void setShouldReportSecondaryInValues(boolean shouldReport);
 
+  void disableLocationShifting();
```
petertrr, line 30: I think it's a good idea to have it here, similarly to reporting in values.yaml, so that the rules can decide whether to enable shifting. It could've been more straightforward, if we were able to check rule class without aditional configuration, but context adapter doesn't have it

petertrr (approved):
LGTM!

iac-extensions/kubernetes/src/test/java/org/sonar/iac/kubernetes/checks/KubernetesVerifier.java
jonas-wielage-sonarsource: Was my first idea, wasn't sure though. I will change it

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/visitors/KubernetesCheckContext.java
jonas-wielage-sonarsource: Yeah, unfortunately it's not possible otherwise right now.

PR #1427, opened 2024-07-22
SONARIAC-1538 S7018: Arguments in multi-line RUN instructions should be sorted

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SortedArgumentListCheck.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final String INSTALL_COMMAND = "install";
+  private static final String ADD_COMMAND = "add";
+  private static final int MINIMAL_LENGTH_TO_SORT = 3;
+
+  private static final CommandDetector DEBIAN_PACKAGE_MANAGER_DETECTOR = CommandDetector.builder()
+    .with(Set.of("apt", "apt-get", "aptitude"))
+    .withAnyFlag()
+    .with(INSTALL_COMMAND)
+    .withAnyFlag()
+    .withOptionalRepeatingExcept("EOF")
```
rudy-regazzoni-sonarsource, line 44: I have bad news for you :sweat_smile: `EOF` is the token we typically use in our examples of the here-doc form, but in fact the user can specify [any token they want](https://docs.docker.com/reference/dockerfile/#here-documents):
```
COPY <<-EOT /script.sh
  echo "hello ${FOO}"
EOT
```
I guess instead, we should match until we have any multi-command operator (`&&,` `||,` `|,`...).
It may be something already handled by `CommandDetector` API.

iac-extensions/docker/src/test/resources/checks/SortedArgumentListCheck/SortedArgumentListCheck.dockerfile
rudy-regazzoni-sonarsource, line 1: I believe it is missing one/few examples with multiple commands, with operators like `&&` or `||`.

rudy-regazzoni-sonarsource (changes requested):
Nice job! The code logic itself LGTM, but there is the issue about `EOF` that needs to be fixed.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/SortedArgumentListCheck.java
petertrr: Yes, this matcher was only for the case of Heredoc. I think I'll disable the rule for heredocs completely, because with all commands glued together sorting can lead to unexpected results.

rudy-regazzoni-sonarsource (approved):
LGTM!

PR #1429, opened 2024-07-23
SONARIAC-1539 S7020: Too long RUN instruction should be split

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/LongRunInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    runInstruction.options().forEach(flag -> incrementLine(wordsPerLine, flag));
+    runInstruction.arguments().forEach(argument -> incrementLine(wordsPerLine, argument));
+    return wordsPerLine;
+  }
+
+  private static void incrementLine(Map<Integer, Integer> wordsPerLine, Tree tree) {
+    if (tree.textRange().start().line() != tree.textRange().end().line()) {
+      incrementLine(wordsPerLine, tree.textRange().start().line());
+    }
+    incrementLine(wordsPerLine, tree.textRange().end().line());
```
petertrr, line 123: Isn't this essentially the same as always using `start()`?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/LongRunInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+   */
+  private RunInstructionData computeRunInstructionDataPerLines(RunInstruction runInstruction) {
+    Map<Integer, Integer> wordsPerLine = countWordsPerLine(runInstruction);
+    var runInstructionData = new RunInstructionData(new HashMap<>(), new HashMap<>(), wordsPerLine);
+
+    TreeVisitor<TreeContext> visitor = new TreeVisitor<>();
+    visitor.register(SyntaxToken.class, (TreeContext ctx, SyntaxToken token) -> {
+      runInstructionData.firstOffsetPerLine.computeIfAbsent(token.textRange().start().line(), line -> token.textRange().start().lineOffset());
+      // Special case of token that finish on a different line: consider the first offset of the end line to start at 0.
+      if (token.textRange().start().line() != token.textRange().end().line()) {
```
petertrr, line 99: We only have access to start and end line, so if there is a syntax token on 3 lines with a too long line in between we won't detect it? Maybe something to represent in tests

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/LongRunInstructionCheck.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private static Map<Integer, Integer> countWordsPerLine(RunInstruction runInstruction) {
+    Map<Integer, Integer> wordsPerLine = new HashMap<>();
+    incrementLine(wordsPerLine, runInstruction.keyword());
+    runInstruction.options().forEach(flag -> incrementLine(wordsPerLine, flag));
+    runInstruction.arguments().forEach(argument -> incrementLine(wordsPerLine, argument));
+    return wordsPerLine;
+  }
+
+  private static void incrementLine(Map<Integer, Integer> wordsPerLine, Tree tree) {
```
petertrr, line 119: I'd call it something like `incrementWordCountOnLine`

petertrr (changes requested):
Good progress, however I'm not sure about the word count parameter.. Please also see my other comments

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/LongRunInstructionCheck.java
rudy-regazzoni-sonarsource: Not sure what you mean here.

The purpose is to increase the counter of lines on which the tree is present.
But, we should not increase the counter twice if the start line and end line are the same.
So here I'm increasing the end line every time, and the start line only if it's different.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/LongRunInstructionCheck.java
rudy-regazzoni-sonarsource: Yes you are correct.
I will add a test to this.

petertrr (approved):
LGTM!

PR #1430, opened 2024-07-23
SONARIAC-1540 S7021: WORKDIR instruction should only be used with absolute path

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/resources/org/sonar/l10n/docker/rules/docker/S7021.html
```diff
@@ -0,0 +1,27 @@
+<p><code>WORKDIR</code> instruction should be used with absolute path for clarity and reliability.</p>
```
jonas-wielage-sonarsource, line 1: IMO `instructions` and adding an `an` before absolute path would sound a bit better

iac-extensions/docker/src/test/resources/checks/WorkdirAbsolutPathCheck/Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM scratch
+
+WORKDIR /absolute/path/unix
+WORKDIR "/absolute/path/unix"
+
```
jonas-wielage-sonarsource, line 5: `WORKDIR ""` would be non compliant, is this expected?

jonas-wielage-sonarsource (approved):
LGTM, please have a look at my suggestions before merging

iac-extensions/docker/src/test/resources/checks/WorkdirAbsolutPathCheck/Dockerfile
rudy-regazzoni-sonarsource: Indeed, even if it's a bit odd, doesn't reflect well if something like that appear on SonarLint. I will handle this use case.

iac-extensions/docker/src/main/resources/org/sonar/l10n/docker/rules/docker/S7021.html
rudy-regazzoni-sonarsource: Good catch! Here is the [PR](https://github.com/SonarSource/rspec/pull/4091) on RSPEC.

PR #1432, opened 2024-07-23
SONARIAC-1542 Docker parser should not crash on empty interpolation or other formats

rudy-regazzoni-sonarsource:


petertrr (approved):
LGTM

PR #1433, opened 2024-07-23
SONARIAC-1543 Docker parser should not crash when heredoc is connected to another program

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -342,7 +343,8 @@ public ShellForm SHELL_FORM_GENERIC() {
     return b.<ShellForm>nonterminal(DockerLexicalGrammar.SHELL_FORM_GENERIC).is(
       f.shellForm(
         b.oneOrMore(
-          f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), ARGUMENT_GENERIC()))));
+          f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), ARGUMENT_GENERIC())),
+        b.optional(f.hereDocument(b.token(DockerLexicalGrammar.HEREDOC_EXPRESSION)))));
   }
 
   public HereDocument HEREDOC_FORM() {
```
rudy-regazzoni-sonarsource, line 350: Maybe rename the occurrences of `HereDocForm` to `HereDoc`, as we are not considering it anymore as a form like `ShellForm` or `ExecForm` but as content that is part of `ShellForm`.

rudy-regazzoni-sonarsource (commented):
Nice progress! I believe the approach is correct, having the `HereDoc` content as part of the `ShellForm` is valid.
Good luck with the heredoc regex :D Our possibilities are limited with it, sometimes it is not always possible to achieve what we want.

rudy-regazzoni-sonarsource (approved):
LGTM! Super nice work! I was afraid that introducing this would change too much things in the parser, but it seems you have handled it the right way.

PR #1434, opened 2024-07-23
SONARIAC-1553 S7029: Prefer COPY over ADD for copying local resources

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
```diff
@@ -0,0 +1,52 @@
...
+  private static final Set<String> ARCHIVE_EXTENSIONS = Set.of("tar", "gz", "tgz", "xz", "txz", "bz2", "tbz2", "tbz", "lz", "tlz", "lzma", "tlzma",
+    "lzo", "tlzo", "7z", "zip");
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(AddInstruction.class, PreferCopyOverAddCheck::checkAddInstruction);
+  }
+
+  private static void checkAddInstruction(CheckContext ctx, AddInstruction tree) {
+    tree.srcs().stream().filter(a -> !isRemoteResourceOrArchive(a)).forEach(a -> ctx.reportIssue(a, MESSAGE));
```
jonas-wielage-sonarsource, line 43: `a` is not a good name

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private static boolean isRemoteResourceOrArchive(Argument argument) {
```
jonas-wielage-sonarsource, line 46: This should only raise when the ArgumentResoltuion is resolved.
Example: `ADD ${UNRESOLVED} ${UNRESOLVED}` should not be non compliant

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
```diff
@@ -0,0 +1,52 @@
...
+import org.sonar.iac.common.api.checks.IacCheck;
+import org.sonar.iac.common.api.checks.InitContext;
+import org.sonar.iac.docker.symbols.ArgumentResolution;
+import org.sonar.iac.docker.tree.api.AddInstruction;
+import org.sonar.iac.docker.tree.api.Argument;
+
+@Rule(key = "S7029")
+public class PreferCopyOverAddCheck implements IacCheck {
+  private static final String MESSAGE = "Replace this ADD instruction with a COPY instruction.";
+  private static final Set<String> ARCHIVE_EXTENSIONS = Set.of("tar", "gz", "tgz", "xz", "txz", "bz2", "tbz2", "tbz", "lz", "tlz", "lzma", "tlzma",
```
jonas-wielage-sonarsource, line 34: I would remove those entries, whose suffix is also contained in this list (`tgz` when `gz` is present) and add a comment which cases are hereby also handled.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private static void checkAddInstruction(CheckContext ctx, AddInstruction tree) {
+    tree.srcs().stream().filter(a -> !isRemoteResourceOrArchive(a)).forEach(a -> ctx.reportIssue(a, MESSAGE));
+  }
+
+  private static boolean isRemoteResourceOrArchive(Argument argument) {
+    var argumentResolution = ArgumentResolution.of(argument);
+
+    var src = argumentResolution.value();
+    return src.startsWith("http") || ARCHIVE_EXTENSIONS.stream().anyMatch(src::endsWith);
```
jonas-wielage-sonarsource, line 50: We should detect `git` as well, as an add instruction could look like:
`ADD git@git.example.com:foo/bar.git /bar`
(See https://docs.docker.com/reference/dockerfile/#adding-private-git-repositories)

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
jonas-wielage-sonarsource: Also please add line breaks, before filter / foreach, makes it a bit easier to read (and comment here on the right location

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
```diff
@@ -0,0 +1,52 @@
...
+  private static final String MESSAGE = "Replace this ADD instruction with a COPY instruction.";
+  private static final Set<String> ARCHIVE_EXTENSIONS = Set.of("tar", "gz", "tgz", "xz", "txz", "bz2", "tbz2", "tbz", "lz", "tlz", "lzma", "tlzma",
+    "lzo", "tlzo", "7z", "zip");
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register(AddInstruction.class, PreferCopyOverAddCheck::checkAddInstruction);
+  }
+
+  private static void checkAddInstruction(CheckContext ctx, AddInstruction tree) {
```
jonas-wielage-sonarsource, line 42: According to the ticket, the issue should be raised on the Add instruction instead of the source.
This way `ADD ./app ./appp /app` would also not raise two times.

jonas-wielage-sonarsource (changes requested):

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/PreferCopyOverAddCheck.java
petertrr: Hm, good catch. I think I'd rather add `.` to the extensions not to match names like `star` as `tar`.

PR #1437, opened 2024-07-24
SONARIAC-1547 Docker parser should not crash when characters are positioned after EXEC form

rudy-regazzoni-sonarsource:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/ExecFormImpl.java
```diff
@@ -45,6 +48,9 @@ public List<Tree> children() {
     result.add(leftBracket);
     result.addAll(argumentsWithSeparators.elementsAndSeparators());
     result.add(rightBracket);
+    if (garbage != null) {
+      result.add(garbage);
```
petertrr, line 52: I wonder if we should add it here. Would this mean that we will highlight this part of the instruction even if it's ignored by both Docker and us?

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
```diff
@@ -96,6 +96,8 @@ public class DockerLexicalConstant {
 
   public static final String IMAGE_ALIAS = "[-a-zA-Z0-9_\\.]+";
 
+  public static final String GARBAGE = "[^\n\r]+";
```
petertrr, line 99: That's a funny name :D But I have nothing better to suggest (`IGNORED`? `IGNORED_TRAILING`?)

petertrr (approved):
LGTM! Please see my suggestion before merging

iac-extensions/docker/src/main/java/org/sonar/iac/docker/tree/impl/ExecFormImpl.java
rudy-regazzoni-sonarsource: I didn't add it at the beginning, but then it failed other parts of the parser.
The way our parser works requires us to provide in children all elements that leads to/is a `SyntaxToken`.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerLexicalConstant.java
rudy-regazzoni-sonarsource: Haha same I couldn't come with something better :sweat_smile: I will try to replace it with something more meaningful.

PR #1443, opened 2024-07-25
SONARIAC-1563 Use ADD for wget

mstachniuk:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/RetrieveRemoteResourcesCheck.java
```diff
@@ -0,0 +1,95 @@
...
+      if (!containsWgetAuthenticationFlags(args)) {
+        var textRangeList = args.stream().map(a -> a.argument().textRange())
+          .toList();
+        var textRange = TextRanges.merge(textRangeList);
+        ctx.reportIssue(textRange, MESSAGE.formatted("wget"));
+      }
+    });
+  }
+
+  private static boolean containsWgetAuthenticationFlags(List<ArgumentResolution> args) {
```
jonas-wielage-sonarsource, line 92: Did you consider using `withAnyFlagExcept()` on the `CommandDetector`? Not saying you should do it, because I'm not sure how easy it is to integrate it in this case here.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/StringPredicate.java
```diff
@@ -30,6 +30,17 @@ public static Predicate<String> startsWithIgnoreQuotes(String value) {
     return str -> stripQuotes(str).startsWith(value);
   }
 
+  public static Predicate<String> startsWithIgnoreQuotes(String... values) {
```
jonas-wielage-sonarsource, line 33: Wouldn't it be possible to reuse the startsWithIgnoreQuotes(String value) and chain it with `Or`?

jonas-wielage-sonarsource (approved):
LGTM! Nice implementation and thorough testing

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/StringPredicate.java
mstachniuk: Yes, it is possible. However, most of the methods in the Command Detector API have the `List` or varargs variants.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/RetrieveRemoteResourcesCheck.java
mstachniuk: I tried but it is not possible. The flags may be in any order and exclude flags may be in different forms, e.g.:
`--proxy-password=password` or `--proxy-password password`.
So the logic is split into 2 parts: 
* detect required flags and URL 
* exclude auth flags

PR #1450, opened 2024-07-26
SONARIAC-1545 Docker parser should support special double-quotes

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
```diff
@@ -346,7 +346,8 @@ public ShellForm SHELL_FORM_GENERIC() {
         b.oneOrMore(
           b.firstOf(
             f.hereDocument(b.token(DockerLexicalGrammar.HEREDOC_EXPRESSION)),
-            f.singleArgumentList(f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), ARGUMENT_GENERIC()))))));
+            f.singleArgumentList(f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), ARGUMENT_GENERIC())),
+            f.singleArgumentList(f.ignoreFirst(b.token(DockerLexicalGrammar.WHITESPACE), f.newArgument(b.oneOrMore(ANY_CHAR_STRING_LITERAL()))))))));
```
rudy-regazzoni-sonarsource, line 350: I was afraid that `ANY_CHAR_STRING_LITERAL` would match too many characters, preventing the logic in `ARGUMENT_GENERIC` from triggering.
But after some tests it seems to be working pretty well, so good job on this :smile:

rudy-regazzoni-sonarsource (approved):

iac-extensions/docker/src/main/java/org/sonar/iac/docker/parser/grammar/DockerGrammar.java
petertrr: Well, it has the lowest precedence, so if `ARGUMENT_GENERIC` matches, this shouldn't steal the match :)

PR #1452, opened 2024-07-26
SONARIAC-1565 S7021 should not raise on special locations such as ~ (unix) or %location% (windows)

jonas-wielage-sonarsource:


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #1472, opened 2024-08-14
SONARIAC-1581 Issue is reported on incorrect line with Go variable declaration

petertrr:


iac-extensions/kubernetes/src/test/resources/checks/VariableNameConventionCheck/helm/templates/template.yaml
```diff
@@ -1,11 +1,11 @@
-{{- $my_Variable := .Values.myVariableValue }} {{- /* Noncompliant {{Rename this variable "$my_Variable" to match the regular expression '^\$[a-z][a-zA-Z0-9]*$'.}} */}}
```
petertrr, line 1: Apparently, parsing Noncompliant comments doesn't work that well for Go templates and is now confusing rather than useful. We will need to follow up on this

jonas-wielage-sonarsource (approved):
LGTM!
Can you add an ITS finding for 117 as well?

PR #1399, opened 2024-07-02
SONARIAC-1521 Support calling checks visitor on the Go template AST

petertrr:

PR #1474, opened 2024-08-15
Remove the interface "ChecksGoTemplate"

jonas-wielage-sonarsource:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/tree/impl/HelmFileTreeImpl.java
```diff
@@ -47,4 +48,11 @@ public static HelmFileTreeImpl fromFileTree(FileTree fileTree, @Nullable GoTempl
   public GoTemplateTree getGoTemplateAst() {
     return goTemplateAst;
   }
+
+  @Override
+  public List<Tree> children() {
```
jonas-wielage-sonarsource, line 53: This allows us to remove the complexity of the `additionalTrees` in the `CrossFileAnalyzer` introduced in this PR: https://github.com/SonarSource/sonar-iac/pull/1399

petertrr: My previous concern was that this would mix two concepts, and that the Go AST is not a child of YAML file AST. However, I don't see how it could cause problems, and the implementation is indeed much cleaner now.

petertrr (approved):
LGTM!

PR #1481, opened 2024-08-22
SONARIAC-706 External importers should accept wildcards in properties

mstachniuk:


iac-extensions/cloudformation/src/main/java/org/sonar/iac/cloudformation/plugin/CloudformationSettings.java
```diff
@@ -70,7 +70,7 @@ public static List<PropertyDefinition> getExternalReportProperties() {
     return List.of(PropertyDefinition.builder(CFN_LINT_REPORTS_KEY)
       .index(33)
       .name("Cfn-Lint Report Files")
-      .description("Paths (absolute or relative) to the files with Cfn-Lint issues.")
+      .description("Paths (absolute or relative) to the files with Cfn-Lint issues. You can use Ant patterns to define paths.")
```
jonas-wielage-sonarsource, line 73: I would suggest to call it ` WildcardPatterns` as it's not completely ant but more ant style and users can find  WildcardPattern in our documentation.

iac-common/src/main/java/org/sonar/iac/common/reports/ExternalReportWildcardProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    return result;
+  }
+
+  private static List<File> getIOFiles(File baseDir, String reportPath) {
+    var pattern = WildcardPattern.create(reportPath, File.separatorChar + "");
+    try {
+      var baseDirPath = baseDir.toPath();
+      try (var stream = Files.find(baseDirPath,
+        Integer.MAX_VALUE,
+        (filePath, fileAttr) -> fileAttr.isRegularFile() && pattern.match(baseDirPath.relativize(filePath).toFile().getPath()))) {
```
jonas-wielage-sonarsource, line 77: Looking at the `RelativePathPattern.match()` method in `org.sonar.api.batch.fs.internal.PathPattern` where they also use the `WildcardPattern`, they do a lot of sanitization / normalization of the path before. Does it make sense here as well?

iac-common/src/main/java/org/sonar/iac/common/reports/ExternalReportWildcardProvider.java
```diff
@@ -0,0 +1,86 @@
...
+        return stream.map(Path::toFile)
+          .toList();
+      }
+    } catch (IOException e) {
+      LOG.debug("Exception, when searching files to import report.", e);
```
jonas-wielage-sonarsource, line 82: Stumbled a bit on this message and put it into Grammarly:

> This appears to be a sentence fragment. Consider rewriting it as a complete sentence.

It didn't make any good suggestions to change it though, so here's my suggestion:

```suggestion
      LOG.debug("Exception when searching for report files to import.", e);
```

iac-common/src/main/java/org/sonar/iac/common/reports/ExternalReportWildcardProvider.java
```diff
@@ -0,0 +1,86 @@
...
+    // util class
+  }
+
+  public static List<File> getReportFiles(SensorContext context, String externalReportsProperty) {
+    var minumumVersion = Version.create(MINIMUM_MAJOR_SUPPORTED_VERSION, MINIMUM_MINOR_SUPPORTED_VERSION);
+    boolean externalIssuesSupported = context.runtime().getApiVersion().isGreaterThanOrEqual(minumumVersion);
+    if (!externalIssuesSupported) {
+      LOG.error("Import of external issues requires SonarQube 7.2 or greater.");
+      return Collections.emptyList();
+    }
```
jonas-wielage-sonarsource, line 54: This should be ordered after 
```
   String[] reportPaths = context.config().getStringArray(externalReportsProperty);

    if (reportPaths.length == 0) {
      return Collections.emptyList();
    }
```
Otherwise, users under 7.2 will get the error message everytime even though they didn't specify any reports to import and don't want to use the feature.

iac-common/src/test/java/org/sonar/iac/common/reports/ExternalReportWildcardProviderTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Test
+  void shouldReturnEmptyListWhenEmptyProperty() {
+    var context = SensorContextTester.create(new File("."));
+
+    var reportFiles = ExternalReportWildcardProvider.getReportFiles(context, EXTERNAL_REPORTS_PROPERTY);
+
+    assertThat(reportFiles).isEmpty();
+    assertThat(logTester.logs(Level.ERROR)).doesNotContain("Import of external issues requires SonarQube 7.2 or greater.");
```
jonas-wielage-sonarsource, line 71: I wouldn't expect the error message when I don't want to import anything.

iac-common/src/test/java/org/sonar/iac/common/reports/ExternalReportWildcardProviderTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  static List<Arguments> shouldReturnFilesForPattern() {
+    return List.of(
+      Arguments.of("big*.txt", List.of("big_file_identifier_after_buffer.txt", "big_file_identifier_in_buffer.txt")),
```
jonas-wielage-sonarsource, line 76: I would add a test case without a pattern, to show it doesn't grab anything that it shouldn't

jonas-wielage-sonarsource (changes requested):

iac-common/src/main/java/org/sonar/iac/common/reports/ExternalReportWildcardProvider.java
mstachniuk: Great catch!

iac-common/src/test/java/org/sonar/iac/common/reports/ExternalReportWildcardProviderTest.java
mstachniuk: Me too and this assertion prove that `doesNotContain()`

rudy-regazzoni-sonarsource (approved):
LGTM!

PR #1489, opened 2024-08-28
SONARIAC-1493 S4423: Support detection of TLS Protocol Downgrades for Micronaut configuration

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM

PR #1491, opened 2024-08-28
SONARIAC-1494 S4830: Support detection of insecure-trust-all-certificates in Micronaut configuration

rudy-regazzoni-sonarsource:


mstachniuk (approved):
LGTM

PR #1492, opened 2024-08-29
SONARIAC-1506 S2092: Support detection of Secure flag in Micronaut configuration

rudy-regazzoni-sonarsource:


GabinL21 (approved):
No integration tests? Otherwise LGTM!

PR #1498, opened 2024-09-03
SONARIAC-402 Missing properties in issue/hotspot message should be surrounded with double quotes

petertrr:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheckTest.java
```diff
@@ -52,13 +52,13 @@ void shouldFindIssuesInDataFactoryJson(String type) {
   @Test
   void shouldFindIssuesInApiManagementServiceJson() {
     ArmVerifier.verify("AnonymousAccessToResourceCheck/Microsoft.ApiManagement_service.json", check,
-      issue(6, 14, 6, 47, "Omitting sign_in authorizes anonymous access. Make sure it is safe here."),
+      issue(6, 14, 6, 47, "Omitting \"signin\" authorizes anonymous access. Make sure it is safe here."),
```
petertrr, line 55: The correct name of the property is `signin` and not `sign_in`: https://learn.microsoft.com/en-us/azure/templates/microsoft.apimanagement/service/portalsettings-signin?pivots=deployment-language-bicep#options-for-name-property

PR #1499, opened 2024-09-03
SONARIAC-1614 Improve precision of S1874 to reduce the FP rate

petertrr:


sonar-helm-for-iac/go.sum
petertrr, line 1: This is the result of executing `go mod tidy`

iac-extensions/kubernetes/src/test/resources/checks/DeprecatedCodeCheck/helm/helm3/templates/_helpers.tpl
```diff
@@ -0,0 +1,11 @@
+{{/*
+Return the appropriate apiVersion for deployment.
+*/}}
+{{- define "deployment.apiVersion" -}}
```
mstachniuk, line 4: The name `deployment.apiVersion` is not the best as `apiVersion` is used in many places and may confuse us. Maybe `deployment.gitVersion` will be better?

iac-extensions/kubernetes/src/test/resources/checks/DeprecatedCodeCheck/helm/helm3/templates/deprecated_code.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  # Noncompliant@+1
+  {{- if semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion }}
+  ingressClassName: {{ .Values.ingress.className |  default "nginx" }}
+  {{- end }}
+  # Noncompliant@+1
+  {{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion)) }}
+  ingressClassName: "test"
+  {{- end }}
+  # It's not possible to raise inside the usage of "Capabilities.KubeVersion.GitVersion" inside the .tpl file
```
mstachniuk, line 29: Do we have a ticket to address it? If not please create a new ticket.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmParser.java
```diff
@@ -95,7 +95,7 @@ private FileTree evaluateAndParseHelmFile(String source, HelmInputFileContext in
   }
 
   static boolean isInvalidHelmInputFile(HelmInputFileContext helmFileCtx) {
-    return isValuesFile(helmFileCtx) || isChartFile(helmFileCtx) || isTplFile(helmFileCtx);
+    return isValuesFile(helmFileCtx) || isTplFile(helmFileCtx);
```
mstachniuk, line 98: I guess that thanks to this change we will be able now to raise issues in Chart.yaml 🤔 
Please verify that, and if yes, then please add a test case for some rule, e.g. TODO check.

mstachniuk (approved):
LGTM

iac-extensions/kubernetes/src/test/resources/checks/DeprecatedCodeCheck/helm/helm3/templates/deprecated_code.yaml
petertrr: A similar ticket is https://sonarsource.atlassian.net/browse/SONARIAC-1309, however now that we can parse the Go AST, we can parse the `.tpl` files as well and execute checks on them. I created a ticket for that: https://sonarsource.atlassian.net/browse/SONARIAC-1616

iac-extensions/kubernetes/src/test/resources/checks/DeprecatedCodeCheck/helm/helm3/templates/_helpers.tpl
petertrr: This function is actually to get the value that will be in the `apiVersion: ...` field of a Deployment. It uses `KubeVersion` to determine a suitable `apiVersion`

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/plugin/HelmParser.java
petertrr: Yes, added a TODO comment in ITs

PR #1501, opened 2024-09-03
SONARIAC-976 S6249 should not raise when the Resource field of the bucket policy is a list

petertrr:
Related RSPEC: https://github.com/SonarSource/rspec/pull/2456


rudy-regazzoni-sonarsource (approved):
LGTM!

PR #1502, opened 2024-09-04
SONARIAC-1607 S7026 should not raise an issue on wget/curl when specific request elements are precised

GabinL21:


iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/curl.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM sratch
+FROM scratch
```
mstachniuk, line 1: Good cach ❤️

iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/curl.dockerfile
```diff
@@ -428,7 +398,29 @@ RUN curl -L https://example.com/resource -s --output output.txt -k -H "X-Auth-To
...
+RUN curl -o output.txt https://example.com/resource --form-string "param1=value1"
+RUN curl -o output.txt https://example.com/resource --header "Content-Type: application/json"
+RUN curl -o output.txt https://example.com/resource -H "Content-Type: application/json"
+RUN curl -o output.txt https://example.com/resource --json '{"param1": "value1"}'
+RUN curl -o output.txt https://example.com/resource --referer "https://example.com/referer"
+RUN curl -o output.txt https://example.com/resource -e "https://example.com/referer"
+RUN curl -o output.txt https://example.com/resource --request POST
+RUN curl -o output.txt https://example.com/resource -X POST
+RUN curl -o output.txt https://example.com/resource --user-agent "Mozilla/5.0"
+RUN curl -o output.txt https://example.com/resource -A "Mozilla/5.0"
```
mstachniuk, line 421: I would expect that those examples are Noncompliant, because "ADD doesn’t support HTTP request options" (line 401)

mstachniuk: Consider to add/modify some of those examples to have "request options" not always at the end but before or after `-o output.txt` argument. It is for better coverage.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/RetrieveRemoteResourcesCheck.java
```diff
@@ -44,17 +44,23 @@ public class RetrieveRemoteResourcesCheck implements IacCheck {
   private static final String CURL = "curl";
 
   private static final List<String> WGET_AUTH_FLAGS = List.of("--http-user", "--http-password", "--proxy-user", "--proxy-password", "--load-cookies");
+  private static final List<String> WGET_REQUEST_FLAGS = List.of("--header", "--method", "--body-data", "--referer", "--save-headers",
```
mstachniuk, line 47: IMO it can be in one list named `WGET_FORBIDDEN_FLAGS` or similar

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/RetrieveRemoteResourcesCheck.java
```diff
@@ -44,17 +44,23 @@ public class RetrieveRemoteResourcesCheck implements IacCheck {
...
   private static final Predicate<String> WGET_DOWNLOAD_FLAG_PREDICATE = startsWithIgnoreQuotes("-O", "--output-document");
-  private static final Predicate<String> URL_PREDICATE = startsWithIgnoreQuotes("http");
 
   private static final List<String> CURL_AUTH_FLAGS = List.of("--anyauth", "--basic", "--digest", "--ntlm", "--negotiate",
     "--proxy-anyauth", "--proxy-basic", "--proxy-digest", "--proxy-ntlm", "--proxy-negotiate", "--user", "-u", "--oauth2-bearer",
     "--proxy-user", "-U", "--tlsuser", "--proxy-tlspassword", "--tlspassword", "--proxy-tlspassword", "--proxy-tlsuser", "--tlsuser", "-b",
     "--cookie", "-c", "--cookie-jar");
+  private static final List<String> CURL_REQUEST_FLAGS = List.of("--data", "-d", "--data-raw", "--data-ascii", "--data-binary",
+    "--data-raw", "--data-urlencode", "--form", "-F", "--form-escape", "--form-string", "--header", "-H", "--json", "--referer", "-e",
+    "--request", "-X", "--user-agent", "-A");
```
mstachniuk, line 57: IMO it can be in one list named `CURL_FORBIDDEN_FLAGS` or similar

iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/wget.dockerfile
```diff
@@ -166,6 +166,44 @@ RUN wget --limit-rate=100k -O /path/to/resource --no-check-certificate https://e
...
+RUN wget -O /path/to/resource -U "Mozilla/5.0" https://example.com/resource
+RUN wget -O /path/to/resource https://example.com/resource -U "Mozilla/5.0"
+
+RUN wget --post-data="key=value" -O /path/to/resource https://example.com/resource
+RUN wget -O /path/to/resource --post-data="key=value" https://example.com/resource
+RUN wget -O /path/to/resource https://example.com/resource --post-data="key=value"
+
+RUN wget --post-file=file.txt -O /path/to/resource https://example.com/resource
+RUN wget -O /path/to/resource --post-file=file.txt https://example.com/resource
+RUN wget -O /path/to/resource https://example.com/resource --post-file=file.txt
```
mstachniuk, line 205: Those are also Noncompliant examples

mstachniuk (changes requested):
Please fix the logic of detecting Noncompliant examples!

General comment. Currently, the rule is implemented in a way of checking existence of forbidden flags. What if the logic will only check if the command contains only allowed flags? I guess it should simplify the logic a lot.

iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/curl.dockerfile
GabinL21: Unsure about what you mean. To me these examples are compliant because ADD doesn't support those flags (like it doesn't support auth), so we can't actually raise an issue on these RUN commands; they can't be replaced by ADD commands as the issue would suggest.

iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/RetrieveRemoteResourcesCheck.java
GabinL21: I was hesitating between the 2, I'll change it thanks!

GabinL21:
> General comment. Currently, the rule is implemented in a way of checking existence of forbidden flags. What if the logic will only check if the command contains only allowed flags? I guess it should simplify the logic a lot.

Regarding this, I was originally going with this "whitelist" logic, but both commands have many possible flags. I believe inversing the logic would either result in longer flag lists, or in more FNs due to missing "authorized" flags.

iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/curl.dockerfile
mstachniuk: Sorry my bad, I confused myself. The selected examples are Compliant.
However, please add a few examples with different argument order, e.g.:
```
RUN curl --data "param1=value1" -o output.txt https://example.com/resource 
RUN curl -o output.txt --data "param1=value1" https://example.com/resource
RUN curl -o output.txt --data "param1=value1" https://example.com/resource --data "param2=value2"

```

iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/curl.dockerfile
```diff
@@ -428,7 +398,29 @@ RUN curl -L https://example.com/resource -s --output output.txt -k -H "X-Auth-To
...
+
+# Compliant: save in /dev/null
```
mstachniuk, line 423: One non compliant example comes to my mind:
```
RUN curl -o output.txt https://example.com/resource -X GET
```

mstachniuk (approved):
LGTM, sorry for the confusion with the compliant examples.

You can still try the approach of "allowed" flags instead of "forbidden" flags.

iac-extensions/docker/src/test/resources/checks/RetrieveRemoteResourcesCheck/curl.dockerfile
GabinL21: Should be covered by the test just above at line 419 (unless I'm missing something special about `GET` 🤔):
`RUN curl -o output.txt https://example.com/resource -X POST`

PR #1503, opened 2024-09-04
SONARIAC-1006 S6382 should handle both old name client_cert and new name client_certificate in impacted resources

rudy-regazzoni-sonarsource:


iac-extensions/terraform/src/main/java/org/sonar/iac/terraform/checks/azure/CertificateBasedAuthenticationCheck.java
```diff
@@ -38,8 +39,10 @@ public class CertificateBasedAuthenticationCheck extends AbstractResourceCheck {
   private static final String MESSAGE_WHEN_DISABLED = "Make sure that disabling certificate-based authentication is safe here.";
   private static final String TEMPLATE_WHEN_MISSING = "Omitting %s disables certificate-based authentication. Make sure it is safe here.";
 
-  private static final String CLIENT_CERT_MODE = "client_cert_mode";
-  private static final String CLIENT_CERT_ENABLED = "client_cert_enabled";
+  private static final String DEFAULT_CLIENT_CERT_MODE = "client_certificate_mode";
+  private static final String DEFAULT_CLIENT_CERT_ENABLED = "client_certificate_enabled";
+  private static final Set<String> CLIENT_CERT_MODE = Set.of("client_cert_mode", DEFAULT_CLIENT_CERT_MODE);
+  private static final Set<String> CLIENT_CERT_ENABLED = Set.of("client_cert_enabled", DEFAULT_CLIENT_CERT_ENABLED);
```
GabinL21, line 45: I'm unsure about how straightforward this is. The naming is a bit confusing, I would say. There are no real differences in the naming between a single value and a Set, and there are no indications of why there are 2 values (besides the ticket description).

What do you think about something like:
```java
  private static final String DEFAULT_CLIENT_CERT_MODE_KEY = "client_certificate_mode";
  private static final String DEFAULT_CLIENT_CERT_ENABLED_KEY = "client_certificate_enabled";
  private static final String OLD_CLIENT_CERT_MODE_KEY = "client_cert_mode";
  private static final String OLD_CLIENT_CERT_ENABLED_KEY = "client_cert_enabled";
  private static final Set<String> CLIENT_CERT_MODE_KEYS = Set.of(DEFAULT_CLIENT_CERT_MODE_KEY, OLD_CLIENT_CERT_MODE_KEY);
  private static final Set<String> CLIENT_CERT_ENABLED_KEYS = Set.of(DEFAULT_CLIENT_CERT_ENABLED_KEY, OLD_CLIENT_CERT_ENABLED_KEY);
  ```

File: iac-common/src/main/java/org/sonar/iac/common/checks/PropertyUtils.java [hunk 2]
```diff
@@ -90,6 +90,10 @@ public static <T extends Tree> Optional<T> get(@Nullable Tree tree, String key,
     return get(tree, key).filter(clazz::isInstance).map(clazz::cast);
   }
 
+  public static <T extends Tree> Optional<T> get(@Nullable Tree tree, Set<String> key, Class<T> clazz) {
```
GabinL21, line 93: If we're nitpicking, it should probably be `keys` instead of `key` since it's a Set :)

PR #1505, opened 2024-09-04
SONARIAC 1122 S6500 should not raise an issue if the option is present anywhere in the command

GabinL21:


petertrr:
> This issue has already been fixed

Ah, apparently SONARIAC-1122 is a duplicate of https://sonarsource.atlassian.net/browse/SONARIAC-1106. Could you please link the two issues in JIRA?

PR #1508, opened 2024-09-05
SONARIAC-1611 S117 should not raise an issue on variable name $ (dollar)

mstachniuk:


iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/VariableNameConventionCheck.java
```diff
@@ -35,7 +35,7 @@
 @Rule(key = "S117")
 public class VariableNameConventionCheck implements IacCheck {
   private static final String MESSAGE = "Rename this variable \"%s\" to match the regular expression '%s'.";
-  private static final String DEFAULT_FORMAT = "^\\$[a-z][a-zA-Z0-9]*$";
+  private static final String DEFAULT_FORMAT = "^\\$(([a-z][a-zA-Z0-9]*$)|$)";
```
petertrr, line 38: Do we need two groups here? Maybe it can be written as `^\\$([a-z][a-zA-Z0-9]*)?$`?

petertrr (approved):
LGTM! Please see my question

PR #1509, opened 2024-09-05
SONARIAC-1083 S6380 should not raise an issue when a child resource defined outside of its parent resource makes it compliant

petertrr:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
```diff
@@ -64,7 +64,8 @@ protected void registerResourceConsumer() {
   }
 
   private static void checkWebSites(ContextualResource resource) {
-    ContextualResource authSettingsV2 = resource.childResourceBy("config", it -> isValue(it.name(), WEBSITES_CONFIG_AUTH_SETTINGS_V2_RESOURCE_NAME).isTrue());
+    ContextualResource authSettingsV2 = resource.childResourceBy("config", it -> isValue(it.name(), WEBSITES_CONFIG_AUTH_SETTINGS_V2_RESOURCE_NAME).isTrue() ||
+      isValue(it.name(), resource.name + "/" + WEBSITES_CONFIG_AUTH_SETTINGS_V2_RESOURCE_NAME).isTrue());
```
petertrr, line 68: I didn't find a good way to hide this details from the predicate. With `type`, it's easier, because type is passed as a string and we can transform it during processing. However, for the same predicate to be applicable to an external child resource, we would need to create a synthetic copy of the node with trimmed name prefix.

mstachniuk: Could you extract those conditions to separate method, as it is a long condition?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
```diff
@@ -82,7 +83,8 @@ private static void checkWebSitesAuthSettings(ContextualResource contextualResou
   }
 
   private static void checkApiManagementService(ContextualResource resource) {
-    ContextualResource signIn = resource.childResourceBy("portalsettings", it -> isValue(it.name(), APIMGMT_PORTALSETTINGS_SIGNIN_RESOURCE_NAME).isTrue());
+    ContextualResource signIn = resource.childResourceBy("portalsettings", it -> isValue(it.name(), APIMGMT_PORTALSETTINGS_SIGNIN_RESOURCE_NAME).isTrue() ||
+      isValue(it.name(), resource.name + "/" + APIMGMT_PORTALSETTINGS_SIGNIN_RESOURCE_NAME).isTrue());
```
mstachniuk, line 88: I'm not sure how to interpret this sentence from the ticket:
```
Also the only child resource of type Microsoft.ApiManagement/service/apis defined, contains an authenticationSettings property.
```
but maybe additional verification of `Microsoft.ApiManagement/service/apis` here is needed? 🤔

mstachniuk (approved):
LGTM. Do you think other checks are somehow affected and should be extended? Please review them.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualResource.java
```diff
@@ -66,6 +71,27 @@ protected HasTextRange toHighlight() {
...
+    var topLevelResources = Optional.ofNullable(tree)
+      .map(ResourceDeclaration::parent)
+      .map(ContextualResource::getChildResources)
+      .orElse(Stream.empty());
+
+    return topLevelResources
+      .filter(this::isExternalChildOfThis)
+      .filter(it -> TextUtils.isValue(it.type(), this.type + "/" + type).isTrue())
+      .filter(predicate)
+      .findFirst()
```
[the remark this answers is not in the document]
petertrr: This call occurs already after all of the filtering: getting external children, checking type and applying the predicate is applied on the stream, so this will return only what was actually requested by caller.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
petertrr: We have a separate check for this resource (`AnonymousAccessToResourceCheck::checkApiManagementServiceApis`), and the issue should only be raised if it's misconfigured. We used to have this check inside this method, but then it was removed. I think in the current form it still represents the intended behavior.

petertrr: Actually, you are right; it seems I created a separate check for this resource to avoid duplication because at that time we didn't handle nested and separate children. Now it can be moved here again and handle both cases.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checkdsl/ContextualResource.java
```diff
@@ -66,6 +71,27 @@ protected HasTextRange toHighlight() {
...
 
   public ContextualResource childResourceBy(String type, Predicate<ResourceDeclaration> predicate) {
+    return Optional.ofNullable(nestedChildResourceBy(type, predicate))
+      .filter(ContextualResource::isPresent)
+      .orElseGet(() -> childResourceOutsideOfThisBy(type, predicate));
+  }
+
+  private ContextualResource childResourceOutsideOfThisBy(String type, Predicate<ResourceDeclaration> predicate) {
+    var topLevelResources = Optional.ofNullable(tree)
+      .map(ResourceDeclaration::parent)
```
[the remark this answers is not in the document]
petertrr: This might need a bit more attention, because now I'm not diving deep into the tree and only checking one level of depth. When jumping to the root node, a full-tree search should be performed to catch the case you described.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/AnonymousAccessToResourceCheck.java
petertrr: Looks like I'll accidentally solve https://sonarsource.atlassian.net/browse/SONARIAC-1044 by this as well

PR #1513, opened 2024-09-06
SONARIAC-1120 S6595 shouldn't raise when "install" command is part of ARG

petertrr:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/utils/CheckUtils.java
```diff
@@ -21,23 +21,32 @@
...
 import org.sonar.iac.docker.symbols.ArgumentResolution;
 import org.sonar.iac.docker.tree.api.CommandInstruction;
 import org.sonar.iac.docker.tree.api.DockerTree;
 import org.sonar.iac.docker.tree.api.Flag;
 import org.sonar.iac.docker.tree.api.HasArguments;
+import org.sonar.iac.docker.tree.api.ShellForm;
+import org.sonar.iac.docker.tree.api.Variable;
 
 public final class CheckUtils {
+  private static final ResolvedCommandPartParser RESOLVED_COMMAND_PART_PARSER = ResolvedCommandPartParser.create();
```
petertrr, line 39: It may seem an overkill, but I couldn't think of an easy way to parse the command without duplicating the logic for splitting shell form and creating artificial `ArgumentResolution`s to map text ranges

rudy-regazzoni-sonarsource (approved):
LGTM! That's an interesting approach, in some way this seems like an overkill, but also it allows to reach ou goal in a quite simple and clean way, so IMO it's good.

PR #1518, opened 2024-09-09
SONARIAC-1084 S6648 should not raise an issue for expression

rudy-regazzoni-sonarsource:


GabinL21 (approved):
No Bicep tests, only JSON? Otherwise LGTM!

PR #1520, opened 2024-09-10
SONARIAC-1589 Update Hadolint rule metadata with new Software Qualities

jonas-wielage-sonarsource:


iac-extensions/docker/src/main/resources/org/sonar/l10n/docker/rules/hadolint/rules.json
jonas-wielage-sonarsource, line 1: As I don't want to use a lot of time for this topic anymore, I just used mapping described in the ticket and didn't look into the rules again. Exception: One rule about root user.

PR #1521, opened 2024-09-10
Do not use WildcardPattern for report paths that are not wildcards

petertrr:


jonas-wielage-sonarsource (approved):
LGTM, please have a look at my suggestion

iac-common/src/main/java/org/sonar/iac/common/reports/ExternalReportWildcardProvider.java
```diff
@@ -60,13 +60,21 @@ public static List<File> getReportFiles(SensorContext context, String externalRe
...
     }
 
     return result;
   }
 
+  private static boolean isWildcard(String path) {
+    return path.contains("*") || path.contains("?");
+  }
+
   private static List<File> getIOFiles(File baseDir, String reportPath) {
```
jonas-wielage-sonarsource, line 78: Do we want to use FileProvider here as well?

PR #1525, opened 2024-09-10
SONARIAC-1619 Reporting an issue on a resource in bicep should highlight the symbolic name instead of the name attribute

petertrr:


rudy-regazzoni-sonarsource (approved):
LGTM! Nice improvement!

PR #1526, opened 2024-09-10
SONARIAC-1610 S6587 should not raise if a cache mount is used

petertrr:


jonas-wielage-sonarsource (approved):
LGTM, please check the snippets I send you before merging.

PR #1530, opened 2024-09-11
SONARIAC-1102 S6388 should check if encryptionSettings is absent or false

GabinL21:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/UnencryptedCloudServicesCheck.java
```diff
@@ -143,6 +139,24 @@ private static boolean isDiskEncryptionSetIdSet(ContextualObject disk) {
       .anyMatch(id -> id.isPresent() && id.is(isNotEmpty()));
   }
 
+  private static void checkEncryptionSettings(ContextualObject osDisk) {
+    ContextualProperty encryptionSettings = osDisk.property("encryptionSettings");
+    if (encryptionSettings.isAbsent()) {
+      osDisk.report(String.format(FORMAT_OMITTING, "encryptionSettings\", \"managedDisk.diskEncryptionSet.id\" or \"managedDisk.securityProfile.diskEncryptionSet.id"));
+      return;
+    } else if (encryptionSettings.is(isFalse())) {
```
petertrr, line 147: To my understanding, this condition is incorrect. According to the [docs](https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachines?pivots=deployment-language-arm-template#osdisk-1), this property should receive an object, and I believe old logic checking a boolean value here can be removed completely. Could you please confirm this with AppSec?

GabinL21: It was indeed a misunderstanding regarding the outcome of the ticket. Fixed it in e45d18b!

petertrr (approved):
LGTM!

PR #1531, opened 2024-09-12
SONARIAC-1487 Implement syntax highlighting for keys in YAML files

petertrr:


iac-extensions/jvm-framework-config/src/main/java/org/sonar/iac/jvmframeworkconfig/parser/yaml/JvmFrameworkConfigYamlConverter.java
```diff
@@ -68,7 +69,11 @@ private Profile convertToProfile(Node node) {
     commentsPerProfile = new ArrayList<>();
     List<Tuple> tuples = convert(node).map(TupleBuilder::build).toList();
 
-    return new ProfileImpl(tuples, commentsPerProfile, profileName(tuples), true);
+    var profile = new ProfileImpl(tuples, commentsPerProfile, profileName(tuples), true);
+    // This step is needed for syntax highlighting later. Keeping track of YAML nodes in the leaves would be more complicated,
+    // because in the flattened structure leaves would have a lot of common ancestors, and they would need to be filtered later.
+    profile.setOriginalYamlTree(yamlConverter.convert(node));
```
petertrr, line 75: An alternative I was considering: store all key nodes in all leaf nodes (i.e., in `Tuple`s), and inside the highlighting visitor get all these nodes for a profile, filter out duplicates, and highlight. This will require more code, and it's not obvious whether it will be cheaper than an additional conversion here

jonas-wielage-sonarsource: Do you have a rough estimate how long this conversion `yamlConverter.convert(node)` is taking for files? 
Having a measurement would help. We don't need it ultra detailed, just a rough estimate would be nice.

iac-extensions/jvm-framework-config/src/test/java/org/sonar/iac/jvmframeworkconfig/plugin/visitors/JvmFrameworkConfigHighlightingVisitorTest.java
```diff
@@ -71,12 +71,23 @@ void shouldHighlightYaml() {
         foo-arr:
           - bar1
           - bar2
+        key1:
+          key2:
+            key3: value
```
jonas-wielage-sonarsource, line 76: Can you add another child under key1 or key 2? 
Something like:
```
key1:
  key2:
    key3: value
  key4: value2
```

iac-extensions/jvm-framework-config/src/main/java/org/sonar/iac/jvmframeworkconfig/parser/yaml/JvmFrameworkConfigYamlConverter.java
petertrr: From running ITs locally:
```
INFO: 1265/1265 source files have been analyzed
INFO: Duration Statistics, Parse 372 ms, JvmFrameworkConfigMetricsVisitor 239 ms, JvmFrameworkConfigHighlightingVisitor 219 ms, ChecksVisitor 103 ms, **yaml-conversion 30 ms**, S6437 21 ms, S4423 12 ms, S1135 6 ms, S3330 3 ms, S2092 2 ms, S4830 2 ms, S5693 2 ms, S4507 1 ms
```
**`yaml-conversion 30 ms`**
So I'd say quite fast

iac-extensions/jvm-framework-config/src/main/java/org/sonar/iac/jvmframeworkconfig/plugin/visitors/JvmFrameworkConfigHighlightingVisitor.java
```diff
@@ -39,6 +44,15 @@ protected void languageSpecificHighlighting() {
         highlight(tree.key(), KEYWORD);
       }
     });
+    register(Profile.class, (InputFileContext ctx, Profile tree) -> {
```
jonas-wielage-sonarsource, line 47: Did you consider adding the `originalYamlTree` as a child and simply adding the `YamlHighlightingVisitor` to the Visitors?
 That way we can remove the `retrieveAllChildrenMethod`

iac-extensions/jvm-framework-config/src/main/java/org/sonar/iac/jvmframeworkconfig/parser/yaml/JvmFrameworkConfigYamlConverter.java
jonas-wielage-sonarsource: That sounds fast enough.

jonas-wielage-sonarsource: I believe you choose the right approach here.
The only thing I was wondering if it makes sense to put it under File and not Profile.
Was there a specific reason you choose this approach?

petertrr: Not really; I don't have a strong opinion, and probably putting it under File would reduce the number of entities that are aware of the original tree

iac-extensions/jvm-framework-config/src/main/java/org/sonar/iac/jvmframeworkconfig/plugin/visitors/JvmFrameworkConfigHighlightingVisitor.java
petertrr: Good suggestion, I didn't think about that. However, there is one problem: in this approach, comments will be present in the tree twice, and highlighting will break.

jonas-wielage-sonarsource: Mh, I didn't think about that. Either way there are present twice. Maybe we should clean them from one tree regardless?

petertrr: Cleaning them out of the YAML tree seems hacky; especially since every YamlTree node implements HasComments. Maybe, a solution with highlighting comments only from a properties file in this visitor will work. Essentially, it will turn it from JvmConfigHighlightingVisitor to PropertiesHighlightingVisitor, but that doesn't seem like a problem

petertrr: No, looks like it won't work without too many changes. Each syntax highlighting visitor tries to save highlighting in `after()` method, even if highlighting is empty. And saving highlighting twice for the same file is not supported by the plugin API

jonas-wielage-sonarsource: As discussed, let's take the original approach of your PR. I believe it adds the least complexity of any solution.

jonas-wielage-sonarsource (approved):
LGTM

PR #1534, opened 2024-09-12
SONARIAC-1605 S6865: Change the detection logic to a more realistic one

rudy-regazzoni-sonarsource:


iac-common/src/testFixtures/java/org/sonar/iac/common/testing/Verifier.java
rudy-regazzoni-sonarsource, line 1: Modifications on this files are about to fix a bug on error reporting.
Providing a emp file without content prevent the display of diff when running test with `Verifier`.

its/plugin/src/integrationTest/java/org/sonarsource/iac/SecondaryLocationTest.java
rudy-regazzoni-sonarsource, line 1: Rule `S6865` is used for integration tests of SecondaryLocation desactivation for helm, so had to update the IT.

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/AutomountServiceAccountTokenCheck.java
```diff
@@ -20,91 +20,100 @@
...
-    var specAsAttributeObject = blockObject.attribute("spec");
-    if (specAsAttributeObject.tree != null && isContainersPresentInSpecBlock(specAsBlockObject)) {
-      var tokenAttribute = specAsBlockObject.attribute(KEY);
-      if (!tokenAttribute.isAbsent()) {
-        tokenAttribute.reportIfValue(TreePredicates.isSet().negate(), message);
-        tokenAttribute.reportIfValue(isTrue(), message);
-        return;
-      }
+  private static boolean isContainersAbsent(BlockObject spec) {
+    return spec.attribute("containers").tree == null;
```
petertrr, line 67: We have a method `isAbsent` on `AttributeObject`, will it work here?

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/AutomountServiceAccountTokenCheck.java
```diff
@@ -20,91 +20,100 @@
...
-  private static void reportIssueWithLinkedAccount(ServiceAccount linkedAccount, BlockObject blockObject, String message) {
-    var blockSpec = blockObject.block("spec");
-    var serviceAccountNameAttr = blockSpec.attribute("serviceAccountName");
-    List<SecondaryLocation> secondaryLocations = new ArrayList<>();
-    secondaryLocations.add(new SecondaryLocation(serviceAccountNameAttr.tree.value(), "Through this service account"));
-    if (linkedAccount.automountServiceAccountToken().isTrue()) {
-      secondaryLocations.add(new SecondaryLocation(linkedAccount.valueLocation(), "Change this setting", linkedAccount.filePath()));
-    }
-    blockObject.attribute("spec").reportOnKey(message, secondaryLocations);
+  private boolean hasAnyBondedRole(BlockObject document, String namespace, @Nullable String accountName) {
```
petertrr, line 99:
```suggestion
  private boolean hasAnyBoundRole(BlockObject document, String namespace, @Nullable String accountName) {
```
I believe it would sound better :)

iac-extensions/kubernetes/src/main/java/org/sonar/iac/kubernetes/checks/AutomountServiceAccountTokenCheck.java
rudy-regazzoni-sonarsource: Ah indeed, good catch! I kept this method from the previous implementation, I guess there was not `isAbsent(...)` at that time.

PR #1540, opened 2024-09-17
SONARIAC-1609 S7020 exception logic should be replaced with a more precise solution

GabinL21:


iac-extensions/docker/src/main/java/org/sonar/iac/docker/checks/LongRunInstructionCheck.java
```diff
@@ -118,11 +130,47 @@ private static void incrementWordCountOnLine(Map<Integer, Integer> wordsPerLine,
...
+    };
+  }
+
+  private static boolean isStringUrl(String value) {
+    try {
+      new URL(value).toURI();
+      return true;
+    } catch (MalformedURLException | URISyntaxException e) {
+      return false;
+    }
```
GabinL21, line 166:
```suggestion
    Pattern regex = Pattern.compile("\\b(?:(https?|ftp|file)://|www\\.)?[-A-Z0-9+&#/%?=~_|$!:,.;]*[A-Z0-9+&@#/%=~_|$]\\.[-A-Z0-9+&@#/%?=~_|$!:,.;]*[A-Z0-9+&@#/%=~_|$]", Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
    Matcher regexMatcher = regex.matcher(value);
    return regexMatcher.matches();
```

Can use regex for validation and handle URLs without protocol, unsure if it's better or not.

petertrr: I think it's safe to assume that we will see full-qualified URLs in Dockerfiles more often than not. For example, in test resources in sonar-iac we have no examples of `www.` being used without a protocol in Dockerfiles.

petertrr (approved):
LGTM!

PR #1543, opened 2024-09-18
SONARIAC-1588 Bicep parser should support String starting with `#` and containing variable interpolation

petertrr:


GabinL21 (approved):
LGTM! 🚀

PR #1546, opened 2024-09-18
SONARIAC-1629 Bicep parser should support spread operator

GabinL21:


iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/PropertyImpl.java
GabinL21, line 1: What do you think about splitting `PropertyImpl` in `KeyValueProperty`and `SpreadProperty`?

PR #1548, opened 2024-09-19
SONARIAC-1475 Should not throw ClassCastException when decorator contains a dot

petertrr:


iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/ArmTreeUtilsTest.java
```diff
@@ -0,0 +1,41 @@
...
+  @CsvSource(textBlock = """
+    foo(), false
+    bar, true
+    """)
+  void shouldReturnFunctionName(String code, boolean isNull) {
+    var expression = (Expression) BicepParser.create(BicepLexicalGrammar.EXPRESSION).parse(code);
+    var functionName = ArmTreeUtils.functionNameOrNull(expression);
+
+    Assertions.assertThat(functionName == null).isEqualTo(isNull);
+  }
```
GabinL21, line 40: I would have personally split this test in two. This one only checks whether the result is null (not the actual function name returned), and I feel like a parametrized test is a bit overkill for two cases 🤔

iac-extensions/arm/src/main/java/org/sonar/iac/arm/checks/elementsorder/ElementsOrderResourceCheckBicep.java
```diff
@@ -87,16 +96,52 @@ private static void checkResource(CheckContext checkContext, ResourceDeclaration
...
+        .map(it -> ((Variable) it).identifier())
+        .filter(Identifier.class::isInstance)
+        .map(it -> ((Identifier) it).value())
+        .map(it -> it + memberExpression.separatingToken().value())
+        .orElse(null);
+
+      var identifier = Optional.ofNullable(memberExpression.expression())
+        .filter(FunctionCall.class::isInstance)
+        .map(it -> ((FunctionCall) it).name())
+        .orElse(null);
```
GabinL21, line 132: The new `functionNameOrNull` utility method could be used there I believe

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/ArmTreeUtils.java
```diff
@@ -132,4 +133,14 @@ public static <T extends ArmTree> Stream<T> findAllNodes(ArmTree tree, Class<T>
       .filter(type::isInstance)
       .map(type::cast));
   }
+
+  /**
+   * If the provided expression is a function call, return the function name, otherwise return null.
+   */
+  public static Identifier functionNameOrNull(Expression expression) {
```
GabinL21, line 140: I feel like the method name doesn't convey the fact that it applies only to function calls (i.e. not to function declarations). Since it's a public utility method that could be used elsewhere, maybe rename it to something like `calledFunctionNameOrNull`?

GabinL21 (approved):
LGTM! Please check my small comments before merging 😄

PR #1549, opened 2024-09-19
SONARIAC-1633 Skip long CI tasks if changes in a PR affect only rule metadata

petertrr:


.cirrus/modules/build.star
```diff
@@ -97,7 +98,7 @@ def build_test_env():
 def build_test_analyze_task():
     return {
         "build_test_analyze_task": {
-            "only_if": is_branch_qa_eligible(),
+            "only_if": "({}) && !({})".format(is_branch_qa_eligible(), is_rule_metadata_update_pr()),
```
petertrr, line 101: One problem with that is that we won't get Sonar analysis results, and they are also required. Combining everything in `build` task will make it ~1 minute longer than nowm and running `build_test_analyze` task for rule metadata PRs will make them ~5 minutes long (both `build` and `build_test_analyze` take about 2 minutes).

PR #1550, opened 2024-09-19
SONARIAC-1631 Bicep parser should parse array types

GabinL21:


GabinL21:
Also tried to make instead a new `TYPE_REFERENCE` that contains `AMBIENT_TYPE_REFERENCE` and `IDENTIFIER` to be closer to [the grammar](https://github.com/Azure/bicep/blob/main/docs/grammar.md?plain=1#L192). However, it is hard to implement due to the parse ordering and [this comment](https://github.com/SonarSource/sonar-iac/blob/b970a2d4eaff6542d278bcaccf7d1b9b88c68de2/iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java#L434), so ended up going for a new `ARRAY_TYPE`.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -517,6 +519,11 @@ public AmbientTypeReference AMBIENT_TYPE_REFERENCE() {
       f.ambientTypeReference(b.token(BicepLexicalGrammar.AMBIENT_TYPE_REFERENCE_VALUE)));
   }
 
+  public ArrayTypeReference ARRAY_TYPE_REFERENCE() {
+    return b.<ArrayTypeReference>nonterminal(BicepLexicalGrammar.ARRAY_TYPE_REFERENCE).is(
+      f.arrayTypeReference(b.token(BicepLexicalGrammar.ARRAY_TYPE_REFERENCE_VALUE)));
```
petertrr, line 524: By capturing the type expression as a single syntax token we lose some information (or will have to parse it later if we want the part without brackets).
I would suggest to structure it as in the grammar: extract `AMBIENT_TYPE_REFERENCE` and `IDENTIFIER` into `TYPE_REFERENCE`, add `TypeReference` interface with nullable fields for brackets and a token inside them.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -431,6 +432,7 @@ public TypeExpressionAble PRIMARY_TYPE_EXPRESSION() {
     return b.<TypeExpressionAble>nonterminal(BicepLexicalGrammar.PRIMARY_TYPE_EXPRESSION).is(
       b.firstOf(
         AMBIENT_TYPE_REFERENCE(),
+        ARRAY_TYPE_REFERENCE(),
```
petertrr, line 435: An idea to try to keep the capturing precedence here: move `literal_value` tothe top, then have `type_reference` with two alternatives (for now): ambient type reference and identifier.

petertrr (changes requested):
As discussed, let's try the direction with a new `TYPE_REFERENCE` rule

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ArrayTypeReferenceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+  public ArrayTypeReferenceImpl(TypeExpressionAble type, SyntaxToken lBracket, @Nullable NumericLiteral length, SyntaxToken rBracket) {
+    this.type = type;
+    this.lBracket = lBracket;
+    this.length = length;
+    this.rBracket = rBracket;
+  }
+
+  @Override
+  public String value() {
+    String value = ((TextTree) type).value() + lBracket.value();
```
GabinL21, line 50: Not a huge fan of this cast, I would add a new `TypeReference` interface that extends `TypeExpressionAble` and `TextTree` to avoid this. Is it too much?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/BicepGrammar.java
```diff
@@ -427,13 +429,37 @@ public SingularTypeExpression SINGULAR_TYPE_EXPRESSION() {
...
+    return b.<TypeExpressionAble>nonterminal(BicepLexicalGrammar.TYPE_REFERENCE).is(
+      f.typeReference(
+        b.firstOf(
+          AMBIENT_TYPE_REFERENCE(),
+          IDENTIFIER()),
+        b.zeroOrMore(TYPE_REFERENCE_SUFFIX())));
+  }
+
+  // Not an infinite recursion, SSLR can handle it
+  @SuppressWarnings("javabugs:S2190")
```
rudy-regazzoni-sonarsource, line 442: Is this part still required?

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/TreeFactory.java
```diff
@@ -476,6 +479,18 @@ public AmbientTypeReference ambientTypeReference(SyntaxToken token) {
     return new AmbientTypeReferenceImpl(token);
   }
 
+  public TypeExpressionAble typeReference(TypeExpressionAble type, Optional<List<TypeReferenceSuffix>> optionalSuffixes) {
+    List<TypeReferenceSuffix> suffixes = optionalSuffixes.or(List.of());
```
rudy-regazzoni-sonarsource, line 483: Nitpick, I would have used `Collections.emptyList()`

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/TreeFactory.java
```diff
@@ -476,6 +479,18 @@ public AmbientTypeReference ambientTypeReference(SyntaxToken token) {
...
+    for (TypeReferenceSuffix suffix : suffixes) {
+      type = suffix.setType(type);
```
rudy-regazzoni-sonarsource, line 485: I would change the method name to not look like a setter.
`setType` name made me initially think it was a simple setter, I didn't expect it to instantiate an object nor to return something.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/bicep/TypeReferenceSuffix.java
```diff
@@ -0,0 +1,28 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.iac.arm.tree.api.bicep;
+
+/**
+ * Interface for type suffix that build into {@link TypeExpressionAble}.
+ */
+public interface TypeReferenceSuffix {
+
+  TypeExpressionAble setType(TypeExpressionAble type);
```
rudy-regazzoni-sonarsource, line 27: Do we want to make this method available in the public API?
Maybe it's not better, but I may have seen sometimes that we keep such a method in the implementation only and cast the interface to the implementation in the `TreeFactory` to access this method.
Otherwise, as we discussed, there is still the solution of having a builder pattern, but it adds some overheads.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ArrayTypeReferenceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+    if (length != null) {
+      value += length.value();
+    }
+    return value + rBracket.value();
+  }
```
rudy-regazzoni-sonarsource, line 55: We usually don't have methods that compute values in that way or that access subelements.
I don't see the added value here, is this method already used in a check?
Usually, we would have a direct getter to `type` and `length,` currently, you have no possibility to access them properly, so you cannot report an issue on length, or you cannot check that `length` is null for example.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ArrayTypeSuffixImpl.java
```diff
@@ -0,0 +1,45 @@
...
+  public ArrayTypeSuffixImpl(SyntaxToken lBracket, @Nullable NumericLiteral length, SyntaxToken rBracket) {
+    this.lBracket = lBracket;
+    this.length = length;
+    this.rBracket = rBracket;
+  }
+
+  @Override
+  public TypeExpressionAble setType(TypeExpressionAble type) {
+    return new ArrayTypeReferenceImpl(type, lBracket, length, rBracket);
+  }
```
rudy-regazzoni-sonarsource, line 44: As mentioned earlier, I would change the name of this method to not be setter-like.

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ArrayTypeReferenceImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+package org.sonar.iac.arm.tree.impl.bicep;
+
+import org.junit.jupiter.api.Test;
+import org.sonar.iac.arm.parser.bicep.BicepLexicalGrammar;
+import org.sonar.iac.arm.tree.api.bicep.ArrayTypeReference;
+
+import static org.sonar.iac.arm.ArmAssertions.assertThat;
+
+class ArrayTypeReferenceImplTest extends BicepTreeModelTest {
+
```
rudy-regazzoni-sonarsource, line 29: I would expect more tests to assert/document what is the resulting tree of parsing this element.
`string[5]`
`string[]`
`string[5][15][0]`

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/PrimaryTypeExpressionTest.java
```diff
@@ -265,12 +291,40 @@ void shouldParseSimpleObjectType() {
...
+  }
+
+  @Test
+  void shouldParseMultiDimensionalArrayType() {
+    String code = "string[3][5][7]";
+    ArrayTypeReference tree = parse(code, BicepLexicalGrammar.PRIMARY_TYPE_EXPRESSION);
+    assertThat(tree.value()).isEqualTo("string[3][5][7]");
+    assertThat(tree.is(ArmTree.Kind.ARRAY_TYPE_REFERENCE)).isTrue();
+    assertThat(recursiveTransformationOfTreeChildrenToStrings(tree)).containsExactly("string", "[", "3", "]", "[", "5", "]", "[", "7", "]");
+  }
```
rudy-regazzoni-sonarsource, line 329: I would have moved this test in the dedicated test class `ArrayTypeReferenceTest`.

rudy-regazzoni-sonarsource (approved):
LGTM! Nice job, I added a few comments, please have a look.
The Jira ticket would benefit some more explanation, without the title it's not so obvious that the parsing error come from the array.
Also, with this we now support array suffixes. What about other suffixes? Do we have tickets already about them? I would add TODO in the code to reference them.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/parser/bicep/TreeFactory.java
GabinL21: Totally agree, I'm trying to come up with a better name but didn't settle on one I like 😆

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/bicep/TypeReferenceSuffix.java
GabinL21: I guess having an implicit expected method on implementations and then casting is definitely not something I want to do, the OOP experts would kill me for it

I guess suffixes are kind of builders already? They just aren't named builders explicitly (personally prefer it this way), but they are not part of the AST (since they don't inherit `ArmTree`). Maybe it's just not in the right package then? But unsure where I should put them 🤔

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/impl/bicep/ArrayTypeReferenceImpl.java
GabinL21: I thought `ArrayTypeReference` had to extend `TextTree` like `AmbientTypeReference`, removing it will simplify the code, love it 😄

And you're right, totally forgot to generate the getters, got bad habits from using Lombok too much 😔

GabinL21:
Created [SONARIAC-1673](https://sonarsource.atlassian.net/browse/SONARIAC-1673) and [SONARIAC-1674](https://sonarsource.atlassian.net/browse/SONARIAC-1674) for the 2 remaining type suffix.

iac-extensions/arm/src/main/java/org/sonar/iac/arm/tree/api/bicep/ArrayTypeReference.java
```diff
@@ -0,0 +1,33 @@
...
+import org.sonar.iac.arm.tree.api.NumericLiteral;
+
+/**
+ * Marker interface for array types (e.g. {@code string[]})
+ */
+public interface ArrayTypeReference extends TypeExpressionAble {
+  TypeExpressionAble getType();
+
+  @CheckForNull
+  NumericLiteral getLength();
```
petertrr, line 32: Why is it called `length`? For me it seemed that this is an index

iac-extensions/arm/src/test/java/org/sonar/iac/arm/tree/impl/bicep/ArrayTypeReferenceImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+    assertThat(tree.is(ArmTree.Kind.ARRAY_TYPE_REFERENCE)).isTrue();
+    assertThat(recursiveTransformationOfTreeChildrenToStrings(tree)).containsExactly("int", "[", "3", "]", "[", "5", "]", "[", "7", "]");
+  }
+
+  @Test
+  void shouldParseMultiDimensionalArrayTypeWithMixedLength() {
+    String code = "object[][15][0]";
+    ArrayTypeReference tree = parse(code, BicepLexicalGrammar.TYPE_REFERENCE);
+    assertThat(tree.is(ArmTree.Kind.ARRAY_TYPE_REFERENCE)).isTrue();
+    assertThat(recursiveTransformationOfTreeChildrenToStrings(tree)).containsExactly("object", "[", "]", "[", "15", "]", "[", "0", "]");
```
petertrr, line 71: I would add a more detailed assertion here showing which types are base and which are suffixes (i.e. the first type is `object[][1] + [0]`, the second is `object[] + [1]`...)

petertrr (approved):
LGTM! Please see my suggestions before merging.

